PR #2, opened 2018-07-28
Add example query for rate calculation

svenklemm:
This adds an example query for rate calculation to the docs.

### DIFF
--- a/using-timescaledb/reading-data.md
+++ b/using-timescaledb/reading-data.md
@@ -96,6 +96,29 @@ SELECT time, AVG(temperature) OVER(ORDER BY time
   WHERE location = 'garage' and time > NOW() - interval '1 day'
   ORDER BY time DESC;
 ```
+
+### Rate [](rate)
+
+To calculate the rate of monotonically increasing counters like bytes sent
+of a host or container you need to account for counter resets caused for
+example by host reboots or container restarts.  The following query calculates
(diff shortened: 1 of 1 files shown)


using-timescaledb/reading-data.md
```diff
@@ -96,6 +96,29 @@ SELECT time, AVG(temperature) OVER(ORDER BY time
   WHERE location = 'garage' and time > NOW() - interval '1 day'
   ORDER BY time DESC;
 ```
+
+### Rate [](rate)
+
+To calculate the rate for monotonically increasing counters like bytes sent
+of a host or container you need to account for counter resets caused for
```
solugebefola: "of" >> "from"?
comma after "caused"

using-timescaledb/reading-data.md
```diff
@@ -96,6 +96,29 @@ SELECT time, AVG(temperature) OVER(ORDER BY time
...
+example by host reboots or container restarts.  The following query calculates
+bytes per second sent and takes counter resets into account.
```
solugebefola: "and takes" >> "while taking"?

RobAtticus:
Nit: Can we use the same style for commit messages here as we do main repo? In this case, capitalize 'add'

svenklemm:
I adjusted the wording

PR #5, opened 2018-08-03
Add example for increase calculation

svenklemm:
This adds an example query for increase calculation and adjusts the wording for rate calculation to reference it.

### DIFF
--- a/using-timescaledb/reading-data.md
+++ b/using-timescaledb/reading-data.md
@@ -97,12 +97,36 @@ SELECT time, AVG(temperature) OVER(ORDER BY time
   ORDER BY time DESC;
 ```
 
+### Increase[](increase)
+
+To calculate the increase from monotonically increasing counters like bytes sent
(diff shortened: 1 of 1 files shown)


using-timescaledb/reading-data.md
```diff
@@ -97,12 +97,36 @@ SELECT time, AVG(temperature) OVER(ORDER BY time
   ORDER BY time DESC;
 ```
 
-### Rate [](rate)
+### Increase[](increase)
 
-To calculate the rate of monotonically increasing counters like bytes sent
+To calculate the increase of monotonically increasing counters like bytes sent
 of a host or container you need to account for counter resets caused for
```
solugebefola: "of" >> "from"

solugebefola: comma after "caused"

svenklemm:
ok adjusted the wording

PR #8, opened 2018-08-07
Add documentation for adaptive chunking

erimatnor:
This adds documentation for new API functions that are added with the
new adaptive chunking functionality. It also adds a section under
"Using TimescaleDB" that gives an introduction and overview of
adaptive chunking.

### DIFF
--- a/api.md
+++ b/api.md
@@ -18,6 +18,7 @@
 > - [indexes_relation_size](#indexes_relation_size)
 > - [indexes_relation_size_pretty](#indexes_relation_size_pretty)
 > - [last](#last)
+> - [set_adaptive_chunking](#set_adaptive_chunking)
 > - [set_chunk_time_interval](#set_chunk_time_interval)
 > - [set_number_partitions](#set_number_partitions)
 > - [show_tablespaces](#show_tablespaces)
@@ -174,13 +175,15 @@ still work on the resulting hypertable.
 |---|---|
 | `partitioning_column` | Name of an additional column to partition by. If provided, the `number_partitions` argument must also be provided. |
 | `number_partitions` | Number of hash partitions to use for `partitioning_column`. Must be > 0. |
-| `chunk_time_interval` | Interval in event time that each chunk covers. Must be > 0. Default is 1 month. |
+| `chunk_time_interval` | Interval in event time that each chunk covers. Must be > 0. Default is 1 month unless [adaptive chunking][adaptive-chunking] is enabled, in which case it starts at 1 day. |
 | `create_default_indexes` | Boolean whether to create default indexes on time/partitioning columns. Default is TRUE. |
 | `if_not_exists` | Boolean whether to print warning if table already converted to hypertable or raise exception. Default is FALSE. |
 | `partitioning_func` | The function to use for calculating a value's partition.|
 | `associated_schema_name` | Name of the schema for internal hypertable tables. Default is "_timescaledb_internal". |
 | `associated_table_prefix` | Prefix for internal hypertable chunk names. Default is "_hyper". |
 | `migrate_data` | Set to `true` to migrate any existing `main_table` data to chunks in the new hypertable. A non-empty table will generate an error without this option. Note that, for large tables, the migration might take a long time. Defaults to false. |
+| `chunk_target_size` | The target size of a chunk (including indexes) in `kB`, `MB`, `GB`, or `TB`. Setting this to `estimate` or a non-zero chunk size, e.g., `2GB` will enable [adaptive chunking][adaptive-chunking]. The `estimate` setting will estimate a target chunk size based on system information. Adaptive chunking is disabled by default. |
+| `chunk_sizing_func` | Allows setting a custom chunk sizing function for [adaptive chunking][adaptive-chunking]. The built-in chunk sizing function will be used by default. Note that `chunk_target_size` needs to be set to use this function.  |
 
 >:WARNING: The use of the `migrate_data` argument to convert a non-empty table can
 lock the table for a significant amount of time, depending on how much data is
@@ -214,6 +217,22 @@ millisecond, nanosecond, etc.).  So if your time column is the number
 of milliseconds since the UNIX epoch, and you wish to each chunk to
 cover 1 day, you should specify `chunk_time_interval => 86400000`.
 
+The units of `chunk_target_size` follow the format of a ["Numeric with
+Unit"][memory-units] memory settings parameter in PostgreSQL. Note
+that this size includes indexes when using the default adaptive
+chunking algorithm. Be careful with specifying a plain number
+(`100000000`) rather than one with units (`100MB`), as the base unit
+is the size of a disk block (typically 8k, in which case 100,000,000 *
+8k ~ 800GB).
+
+<!-- -->
+>:TIP: Setting a reasonable initial `chunk_time_interval` is important
+even with adaptive chunking enabled, because it allows the adaptive
+algorithm to more quickly reach the target chunk size. It is better to
+set a too small `chunk_time_interval` as opposed to a too large
+one. If no `chunk_time_interval` is set with adaptive chunking, the
+default initial interval is 1 day.
+
 In case of hash partitioning (i.e., `number_partitions` is greater
 than zero), it is possible to optionally specify a custom partitioning
 function. If no custom partitioning function is specified, the default
@@ -225,6 +244,15 @@ partitioning function should take a single `anyelement` type argument
 and return a positive `integer` hash value. Note that this hash value
 is _not_ a partition ID, but rather the inserted value's position in
 the dimension's key space, which is then divided across the partitions.
+
+[Adaptive chunking][adaptive-chunking] can be enabled by setting the
+`chunk_target_size` to a non-zero human-readable value, e.g.,
+`2GB`. The `chunk_target_size` should ideally not exceed the setting
+of `shared_buffers`. It is also possible to set `chunk_target_size` to
+`estimate`, in which case the system determines a suitable chunk
+target size based on the PostgreSQL `shared_buffers` setting.
+Optionally, `chunk_sizing_func` can also be set to use a custom
+algorithm for adapting the chunk size.
 
 <!-- -->
 >:TIP: The `add_dimension` function can be used following hypertable
@@ -270,6 +298,12 @@ if `conditions` is already a hypertable:
 SELECT create_hypertable('conditions', 'time', if_not_exists => TRUE);
 ```
 
+Convert table `conditions` to hypertable with adaptive chunking
+enabled and a chunk target size of `2GB`:
+```sql
+SELECT create_hypertable('conditions', 'time', chunk_target_size => '2GB');
+```
+
 #### Best Practices [](create_hypertable-best-practices)
 
 Users of TimescaleDB often have two common questions:
@@ -277,11 +311,21 @@ Users of TimescaleDB often have two common questions:
 1. How large should I configure my intervals for time partitioning?
 1. Should I use space partitioning, and how many space partitions should I use?
 
-**Time intervals**: The current release of TimescaleDB does not
-perform adaptive time intervals (although this is in the works).
-So, users must configure it when creating their hypertable by
-setting the `chunk_time_interval` (or use the default of 1 month).
-The interval used for new chunks can be changed by calling `set_chunk_time_interval`.
+**Time intervals**: The current release of TimescaleDB enables both
+the manual and automated adaption of its time intervals. With
+manually-set intervals, users should specify a `chunk_time_interval`
+when creating their hypertable (the default value is 1 week). The
+interval used for new chunks can be changed by calling
+[`set_chunk_time_interval()`](#set_chunk_time_interval). With
+automatically adapted intervals (which are not enabled by default),
+the user should specify a `chunk_target_size` and the chunk interval
+will be adapted for future chunks with the specified
+`chunk_time_interval` as a starting point, or 1 day if not
+specified. The settings for adaptive chunking can be changed by
+calling [`set_adaptive_chunking()`](#set_adaptive_chunking).
+
+The following instructions apply if users are choosing to configure
+time intervals manually.
 
 The key property of choosing the time interval is that the chunk (including indexes) belonging to the most recent interval (or chunks if using space
 partitions) fit into memory.  As such, we typically recommend setting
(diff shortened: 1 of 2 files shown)


using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+
+The secret behind TimescaleDB's excellent performance with time-series
+data is its automatic partitioning scheme that splits the data across
+a number of N-dimensional *chunks*. However, the performance will only
+be good if the system achieves the "right" *chunk size* that won't
+blow up the memory budget. Oversized chunks result in slow INSERT
+performance when updating indexes and also negatively affect query
+speed.
+
+Unfortunately, the chunk size cannot be set directly as the size is
```
RobAtticus: I think this sentence is a bit confusing because some might think you can set the size, since you can modify the chunk interval. I think we should clarify here (and first paragraph) you are talking about chunk size _in bytes_. It's not needed everywhere in the document, but at least early on its useful so the reader knows what we mean by size.

using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+set its own function to calculate the next `chunk_time_interval`,
+allowing customized algorithms.
+
+>:TIP: It is recommended that hypertables with adaptive chunking
+enabled has an index on the dimension being adapted, typically the
+"time" dimension. Such an index is created by default, unless
+otherwise specified. The index is used by the default adaptive
+chunking algorithm to get the `min` and `max` value from a chunk to
+calculate its fill factor. Without an index, it has to resort to a
+heap scan, which is expensive. The system will print `WARNING`s in the
```
RobAtticus: `to a heap scan (i.e., scanning all the tuples in the chunk)` for readers who may not be familiar with database internals

using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+log if has to do heap scans.
+
+### Usage
+
+To create a hypertable that uses adaptive chunking, simply call
+[`create_hypertable`][1] as follows:
```
RobAtticus: what does `[1]` do here?

using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+speed.
+
+Unfortunately, the chunk size cannot be set directly as the size is
+governed by the constraints on the chunk and the data
+rate. Dynamically altering constraints on an existing chunk to achieve
+a certain chunk size is an expensive and often impossible operation,
+especially in face of changing data rates. Thus, a change in data rate
+might either "overfill" or "underfill" a chunk.
+
+Thus, setting a good `chunk_time_interval` is crucial to achieving a
```
solugebefola: Take out "Thus" here.  Start with "Setting"

using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+good chunk size, since it determines the time constraint on
+chunks. Fortunately, the `chunk_time_interval` can also be updated on
```
solugebefola: remove "also"

using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+an existing hypertable, although this new setting only applies to
+chunks created after the change. A conflict resolution mechanism
+within TimescaleDB automatically detects any "collisions" between new
+and old chunks due to the change in partitioning, and cuts chunks to
+fit within existing constraints.
+
+*Adaptive chunking* leverages the ability to set new
+`chunk_time_interval`s, automating this process.  The system picks the
```
solugebefola: "automating this process" >> "through automation"?

api.md
```diff
@@ -174,13 +175,15 @@ still work on the resulting hypertable.
 |---|---|
 | `partitioning_column` | Name of an additional column to partition by. If provided, the `number_partitions` argument must also be provided. |
 | `number_partitions` | Number of hash partitions to use for `partitioning_column`. Must be > 0. |
-| `chunk_time_interval` | Interval in event time that each chunk covers. Must be > 0. Default is 1 month. |
+| `chunk_time_interval` | Interval in event time that each chunk covers. Must be > 0. Default is 1 month unless adaptive chunking enabled, in which case it is 1 day. |
```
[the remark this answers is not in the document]
solugebefola: Is this argument ignored if adaptive chunking is enabled?  If so, that should be noted.

using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+first *open* (time) dimension to adapt, which is the same dimension
+that `chunk_time_interval` works on.  With adaptive chunking, however,
+the user instead sets a `chunk_target_size`, e.g., `1GB` and the
+system will try to adapt the `chunk_time_interval` on new chunks to
+reach this target chunk size using an algorithm.
+
+>:TIP: Setting a reasonable initial `chunk_time_interval` is still
+important because it allows the adaptive algorithm to quicker reach
```
solugebefola: "quicker" >> "more quickly"

using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+
+*Adaptive chunking* leverages the ability to set new
+`chunk_time_interval`s, automating this process.  The system picks the
+first *open* (time) dimension to adapt, which is the same dimension
+that `chunk_time_interval` works on.  With adaptive chunking, however,
+the user instead sets a `chunk_target_size`, e.g., `1GB` and the
+system will try to adapt the `chunk_time_interval` on new chunks to
+reach this target chunk size using an algorithm.
+
+>:TIP: Setting a reasonable initial `chunk_time_interval` is still
```
solugebefola: Maybe this tip should be duplicated on the API page.

using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+worse than setting a too small one, because the system will be
+committed to that larger interval until the next chunk. If no
+`chunk_time_interval` is set with adaptive chunking, the default is 1
+day.
+
+The default algorithm to calculate the next chunk's interval takes
+into account the *fill factor* of previous chunks, i.e., if they were
+over-filled or under-filled, and changes the interval accordingly. The
+system picks the first *open* (time) dimension to adapt. A user can
+set its own function to calculate the next `chunk_time_interval`,
```
solugebefola: "its own" >> "a custom"

using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+allowing customized algorithms.
+
+>:TIP: It is recommended that hypertables with adaptive chunking
+enabled has an index on the dimension being adapted, typically the
```
solugebefola: "has" >> "have"

using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+"time" dimension. Such an index is created by default, unless
+otherwise specified. The index is used by the default adaptive
+chunking algorithm to get the `min` and `max` value from a chunk to
+calculate its fill factor. Without an index, it has to resort to a
```
solugebefola: "has to" >> "must"

using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+```sql
+SELECT create_hypertable('conditions', 'time', chunk_target_size => '2GB');
+```
+
+This turns the table `conditions` into a hypertable that will try to
+adapt the time dimension to reach the given target chunk size.
+
+It is also possible to let the system *estimate* a target chunk size
+in case it is unclear what a good chunk size would be for the given
+system. In that case do:
```
solugebefola: "do" >> "run"?

using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+
+```sql
+SELECT create_hypertable('conditions', 'time', chunk_target_size => 'estimate');
+```
+
+This will estimate the chunk target size based on system settings and
+parameters.  In particular, it looks at the PostgreSQL configuration
+parameter for `effective_cache_size` as a measure of available disk
+cache. However, it will never go higher than system memory (actually
```
solugebefola: Not sure what you are trying to convey here.  Are you saying that PostgreSQL __should__ go higher than available memory if `effective_cache_size` is configured properly?  Or are you saying that __even if__ it's configured improperly, the DB will still keep within those memory upper and lower bounds?

using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+some fraction of it) or lower than `shared_buffers` if
+`effective_cache_size` is not configured properly.
+
+### Using a custom adaptive chunking algorithm
```
solugebefola: Maybe this should be under an "Advanced usage" headline?

using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,123 @@
...
+
+To enable adaptive chunking with a custom function, do:
+
+```sql
+SELECT create_hypertable('conditions', 'time',
+    chunk_target_size => '2GB',
+    chunk_sizing_func => 'custom_calculate_chunk_interval');
+```
+
+or, change the function on an existing hypertable:
```
solugebefola: "or, __to__ change..."?

mfreed:
Should we explicitly add a :WARNING: that adaptive chunking still in "beta" status?

api.md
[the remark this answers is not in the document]
erimatnor: Not ignored.

api.md
```diff
@@ -226,6 +229,14 @@ and return a positive `integer` hash value. Note that this hash value
 is _not_ a partition ID, but rather the inserted value's position in
 the dimension's key space, which is then divided across the partitions.
 
+[Adaptive chunking][adaptive-chunking] can be enabled by setting the
+`chunk_target_size` to a human-readable size, such as `'2GB'` or
```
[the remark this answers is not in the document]
erimatnor: Yes, but you will get the following error if not:

```
select create_hypertable('cond', 'time', chunk_target_size => '2gb');
ERROR:  invalid data amount
HINT:  Valid units for this parameter are "kB", "MB", "GB", and "TB".
```

But maybe still worth mentioning upfront?

api.md
```diff
@@ -546,6 +563,60 @@ SELECT set_chunk_time_interval('conditions', 86400000);
...
+```
+
+Disable adaptive chunking on the `conditions` hypertable, staying with the
+current chunk time interval.
+
+```sql
+SELECT * FROM set_adaptive_chunking('conditions', 'off');
+               chunk_sizing_func                | chunk_target_size
+------------------------------------------------+-------------------
+ _timescaledb_internal.calculate_chunk_interval |                 0
```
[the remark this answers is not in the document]
erimatnor: Yeah, it is because it is using the built-in vs custom function. Should make names more distinct to clarify.l

using-timescaledb/adaptive-chunking.md
erimatnor: I am trying to say something about how the estimate is made so that people know how to maybe affect it. But maybe I should just skip this, because estimate is probably for those that don't really care about internals or configuring other things.

using-timescaledb/adaptive-chunking.md
erimatnor: You mean replace current title or add an extra (higher-level) headline?

api.md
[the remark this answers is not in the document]
solugebefola: Yah, it should be explicit in the docs I think.

api.md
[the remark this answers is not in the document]
solugebefola: could `chunk_target_size` be null here? (not a comment on the docs, just a question about functionality) . It seems odd that it has a value when adaptive chunking is off.

api.md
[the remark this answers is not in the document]
solugebefola: You could make a separate section for "Units" similar to the API entry for `create_hypertable` http://docs.timescale.com/v0.10/api#create_hypertable

api.md
```diff
@@ -226,6 +237,14 @@ and return a positive `integer` hash value. Note that this hash value
 is _not_ a partition ID, but rather the inserted value's position in
 the dimension's key space, which is then divided across the partitions.
 
+[Adaptive chunking][adaptive-chunking] can be enabled by setting the
+`chunk_target_size` to a human-readable size, such as `'2GB'` or
+`'100MB'`. It is also possible to set it to `'estimate'`, in which
+case the system determines a suitable chunk target size based on
+system information, such as available memory. Optionally,
```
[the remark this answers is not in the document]
erimatnor: This isn't really much of an algorithm. Currently it is mostly just reading the shared_buffers value, but I can be specific. Still, I figured being a bit vague gives us options to change this in the future without updating the docs :).

api.md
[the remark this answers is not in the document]
erimatnor: I guess it could, strictly. But in the current implementation 0 means off.

using-timescaledb/adaptive-chunking.md
```diff
@@ -0,0 +1,130 @@
...
+
+This turns the table `conditions` into a hypertable that will try to
+adapt the time dimension to reach the given target chunk size.
+
+It is also possible to let the system *estimate* a target chunk size
+in case it is unclear what a good chunk size would be for the given
+system. In that case run:
+
+```sql
+SELECT create_hypertable('conditions', 'time', chunk_target_size => 'estimate');
```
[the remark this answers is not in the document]
erimatnor: `SELECT chunk_target_size FROM _timescaledb_catalog.hypertable;` would work. But I agree that a getter might be useful. Still, we have no getters for other things, like `chunk_time_interval`. Maybe this can be a story about better/easier visibility into system settings in general?

api.md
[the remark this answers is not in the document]
erimatnor: Thought I'd expand on the unit thing here.

As indicated in the PostgreSQL docs pointed to above, the "Memory With Unit" parameters (which is used by `chunk_target_size`) has a base unit, e.g., Kb (1024) or a block (8k). Currently, our base unit is one block (8k), which is the same as for the `shared_buffers` setting in PostgreSQL. 

This has the implication that if you'd set `chunk_target_size` to a "machine"-numeric value, such as 1, then the size in bytes would be 8k bytes (8 * 1024).

The reason for 8k block size is that the internal parsing function only accepts an `int` so in order to not overflow for large sizes, the base unit is 8k.

Maybe we should mention the base unit in case people input numerical values (probably not common). They might expect such a value to be in bytes and not blocks?

I chose to avoid talking about this since it is not the preferred way to input the `chunk_target_size` and the configuration page for `shared_buffers` isn't explicit about this either.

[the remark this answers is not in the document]
mfreed: Ah.  So it sounds like this also accepts plain numerics "8388608"? (1024*1024*8)  But in this case, rather than 8MB, this could actually be 64GB?   I think that mistake will be made by somebody, so we should definitely document.

api.md
```diff
@@ -270,6 +295,12 @@ if `conditions` is already a hypertable:
...
 
+Convert table `conditions` to hypertable with adaptive chunking
+enabled and a chunk target size of `2GB`:
+```sql
+SELECT create_hypertable('conditions', 'time', chunk_target_size => '2GB');
+```
+
 #### Best Practices [](create_hypertable-best-practices)
 
 Users of TimescaleDB often have two common questions:
```
mfreed: Best practices start with a paragraph that needs to be changed:

"Time intervals: The current release of TimescaleDB does not perform adaptive time intervals (although this is in the works). So, users must configure it when creating their hypertable by setting the chunk_time_interval (or use the default of 1 month). The interval used for new chunks can be changed by calling set_chunk_time_interval."

Perhaps rewrite to:

"Time intervals: The current release of TimescaleDB enables both the manual and automated adaption of its time intervals.   With manually-set intervals, users should configure the `chunk_time_interval` when creating their hypertable (the default value is 1 week).   The interval used for new chunks can be changed by calling set_chunk_time_interval.  With automatically adapted intervals (which are *not* enabled by default), the chunk interval will start at 1 day (unless otherwise specified at hypertable creation) and the algorithm will automatically adapt the interval for future chunks.  

The following instructions apply if users are choosing the configure time intervals themselves."

api.md
```diff
@@ -214,6 +217,19 @@ millisecond, nanosecond, etc.).  So if your time column is the number
 of milliseconds since the UNIX epoch, and you wish to each chunk to
 cover 1 day, you should specify `chunk_time_interval => 86400000`.
 
+The units of `chunk_target_size` follow the format of a ["Numeric with
+Unit"][memory-units] memory settings parameter in PostgreSQL. Note
+that this size includes indexes when using the default adaptive
+chunking algorithm.
```
mfreed: "Be careful with specifying a plain number ("100000000") rather than one with units ("100MB"), as such plain numbers can correspond to an unexpected unit, e.g., the number of memory pages (in which case 100,000,000 * 8K pages = 800GB)."

erimatnor:
Is this ready for merge? @solugebefola still hasn't approved. Is there still something that is waiting to be addressed? (I think I addressed all outstanding issues/comments.)

solugebefola:
@erimatnor I didn't see the last change b/c the change request from Mike was in a different location from the change.  Approved

PR #16, opened 2018-08-20
Rewrite gap filling queries to not use CTE

svenklemm:
This modifies the 2 gap filling queries in the docs section to not use CTEs.

### DIFF
--- a/using-timescaledb/reading-data.md
+++ b/using-timescaledb/reading-data.md
@@ -277,9 +277,9 @@ This query will then output data in the following form:
  2017-09-22 |   9855
 ```
 Unlike with date_trunc, we can use a custom time interval with the time_bucket function,
-but in order for the generates times to align with the subselect we need use time_bucket on
+but in order for the generated times to align with the subselect we need to use time_bucket on
 the first argument to generate_series.
-For example, let's say you want 1080 data points in the last two weeks, and as many graphing
+For example, let's say you want 1080 data points in the last two weeks and, as many graphing
 libraries require time data points with null values to draw gaps in a graph, we need to
 generate the correct timestamp for each of the data points even if there is no data there.
 Note that we can do basic arithmetic operations on intervals easily in order to get the correct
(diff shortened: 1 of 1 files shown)


using-timescaledb/reading-data.md
```diff
@@ -282,31 +277,29 @@ This query will then output data in the following form:
  2017-09-22 |   9855
 ```
 Unlike with date_trunc, we can use a custom time interval with the time_bucket function,
-but in order to do gap filling properly with it we will also need to use time_bucket on our generated series.
+but in order for the generates times to align with the subselect we need use time_bucket on
```
solugebefola: "generates" >> "generated"

solugebefola: "...we need _to_ use..."

using-timescaledb/reading-data.md
```diff
@@ -282,31 +277,29 @@ This query will then output data in the following form:
...
+the first argument to generate_series.
 For example, let's say you want 1080 data points in the last two weeks, and as many graphing
```
solugebefola: Not part of this PR, but comma should go from in front of "and" to after, I believe.

using-timescaledb/reading-data.md
```diff
@@ -282,31 +277,29 @@ This query will then output data in the following form:
...
 generate the correct timestamp for each of the data points even if there is no data there.
 Note that we can do basic arithmetic operations on intervals easily in order to get the correct
 value to pass to time_bucket.
 ```sql
-WITH data AS (
+SELECT
+  period AS time,
+  avg_data
+FROM
+  generate_series(time_bucket(interval '2 weeks' / 1080, now()-'2 weeks'::interval), now(), '2 weeks'::interval / 1080) AS period
```
davidkohn88: Add a new line to distinguish between time_bucket call and generate series call.

davidkohn88: We might also want to use named arguments? (Just for clarity?)

PR #18, opened 2018-08-20
Add locf example query

svenklemm:
This adds an example query for locf, i feel like the description could be better but nothing came to mind.


using-timescaledb/reading-data.md
```diff
@@ -325,6 +325,37 @@ This query will output data of the form:
  2018-03-09 20:34:40+00 |  1054.10
  2018-03-09 20:53:20+00 |  1037.78
 ```
+
+### Last Observation Carried Forward [](locf)
```
davidkohn88: I think I might move this to after last point queries? just cause we basically use one in the query. So it might be a nice way to put the two things together. (and explicitly state that in the docs).

using-timescaledb/reading-data.md
```diff
@@ -325,6 +325,37 @@ This query will output data of the form:
...
+your results. In this situation you want to carry forward the last observed
+value.
+
+```sql
+SELECT
+  period AS time,
+  coalesce(
+    data_value,
+    (
+      SELECT data_value FROM my_hypertable m WHERE m.time < period ORDER BY time DESC LIMIT 1
```
davidkohn88: we need a meter_id here if this is going to be correct. 
Other thing that might be better here is to pop the gap filling query into a with clause (or a nested from) and then do two levels of a coalesce, first coalescing on the gap filling query itself, then going back to the hypertable only if needed. The best way might be with a recursive CTE that used the previous result as the second level of the coalesce so we'd only ever go to the hypertable once even with a horrible series where the first several values are null.

davidkohn88: ie ```sql
WITH gap_filled as (
SELECT * 
FROM
  generate_series(time_bucket(interval '2 weeks' / 1080, now()-'2 weeks'::interval), now(), '2 weeks'::interval / 1080) AS period
  LEFT JOIN (
    SELECT
      time_bucket('2 weeks'::interval / 1080, time::timestamptz) as btime,
      meter_id,
      avg(data_value) AS data_value
    FROM my_hypertable
    WHERE
      time > now() - '2 weeks'::interval
      AND meter_id IN (1, 2, 5, 8, 10)
    GROUP BY 1, 2
  ) t ON t.btime = period) 
SELECT
  period AS time,
  coalesce(
    data_value,
    (SELECT data_value FROM gap_filled gf1 WHERE gf1.period <  gf.period  and gf1.meter_id = gf.meter_id ORDER BY time DESC LIMIT 1), 
    (SELECT data_value FROM my_hypertable m WHERE m.time < gf.period  AND m.meter_id = gf.meter_id ORDER BY time DESC LIMIT 1)
  ) AS data_value 
FROM gap_filled gf
```

PR #35, opened 2018-10-06
Use apt-get instead of dpkg for installing

svenklemm:
dpkg will not install dependencies and requires an extra step
to install missing dependencies. apt-get will automatically install
needed packages.


svenklemm:
you need ./package.deb to make it work

RobAtticus:
LGTM, thanks

PR #37, opened 2018-10-06
Use ORDER BY time in SQL snippet cause timestamp column must be named time

svenklemm:
Since Grafana requires timestamp column to be aliased to time its probably better to use that in the example SQL snippet.

### DIFF
--- a/using-timescaledb/visualizing-data.md
+++ b/using-timescaledb/visualizing-data.md
@@ -16,8 +16,8 @@ Queries run through Grafana will continue to benefit from the performance improv
 built into TimescaleDB. In fact, this data source was built by TimescaleDB engineers,
 and it is designed to take advantage of the databases' time-series capabilities.
 
->:WARNING: Grafana expects data received to come in timestamp order. When querying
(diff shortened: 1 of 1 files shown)


using-timescaledb/visualizing-data.md
```diff
@@ -16,8 +16,8 @@ Queries run through Grafana will continue to benefit from the performance improv
 built into TimescaleDB. In fact, this data source was built by TimescaleDB engineers,
 and it is designed to take advantage of the databases' time-series capabilities.
 
->:WARNING: Grafana expects data received to come in timestamp order. When querying
-Grafana using SQL, you must include the `ORDER BY timestamp` statement so that
+>:WARNING: Grafana expects data received to be ordered by time. When querying
+Grafana using SQL, you must include the `ORDER BY time` statement so that
```
solugebefola: `ASC` or `DESC` not required?

svenklemm: it needs to be ASC which is the default so doesnt need to be included

PR #40, opened 2018-10-12
Document create_hypertable return values

svenklemm:


api.md
```diff
@@ -186,6 +186,17 @@ still work on the resulting hypertable.
 | `chunk_target_size` | The target size of a chunk (including indexes) in `kB`, `MB`, `GB`, or `TB`. Setting this to `estimate` or a non-zero chunk size, e.g., `2GB` will enable [adaptive chunking][adaptive-chunking]. The `estimate` setting will estimate a target chunk size based on system information. Adaptive chunking is disabled by default. |
 | `chunk_sizing_func` | Allows setting a custom chunk sizing function for [adaptive chunking][adaptive-chunking]. The built-in chunk sizing function will be used by default. Note that `chunk_target_size` needs to be set to use this function.  |
 
+#### Return value
+
+|Name|Description|
+|---|---|
+| `hypertable_id` | ID of the hypertable in _timescaledb_catalog.hypertable |
```
RobAtticus: maybe instead: `ID of the hypertable in TimescaleDB's internal catalog.` 

Not sure it's worth bringing up our internal schemas in the docs.

api.md
```diff
@@ -186,6 +186,17 @@ still work on the resulting hypertable.
...
+| `schema_name` | Schema name of the table converted to hypertable. |
+| `table_name` | Table name of the table converted to hypertable. |
+
+>:TIP: If you use `SELECT * FROM create_hypertable(...)` you will get the return value formatted
+as table with column headings.
```
dianasaur323: minor nit: I think it might read better as: you will get the return value formatted as a table with column headings.

api.md
```diff
@@ -186,6 +186,17 @@ still work on the resulting hypertable.
 | `chunk_target_size` | The target size of a chunk (including indexes) in `kB`, `MB`, `GB`, or `TB`. Setting this to `estimate` or a non-zero chunk size, e.g., `2GB` will enable [adaptive chunking][adaptive-chunking]. The `estimate` setting will estimate a target chunk size based on system information. Adaptive chunking is disabled by default. |
 | `chunk_sizing_func` | Allows setting a custom chunk sizing function for [adaptive chunking][adaptive-chunking]. The built-in chunk sizing function will be used by default. Note that `chunk_target_size` needs to be set to use this function.  |
 
+#### Return value
```
solugebefola: Switch to "Returns" to match other functions

api.md
```diff
@@ -186,6 +186,17 @@ still work on the resulting hypertable.
...
+
+|Name|Description|
```
solugebefola: Also here, to match other functions `Column|Description`

solugebefola: Unless it doesn't return as a table?

PR #44, opened 2018-10-17
Document created flag in return value for add_dimension and create_hypertable

svenklemm:


api.md
```diff
@@ -66,6 +66,7 @@ be run only on an empty hypertable.
 | `schema_name` | Schema name of the hypertable.|
 | `table_name` | Table name of the hypertable. |
 | `column_name` | Column name of the column to partition by. |
+| `created` | True when the dimension was added, false when `if_not_exists` was true and no dimension was added. |
```
RobAtticus: when->if (both times)
was->is

api.md
```diff
@@ -202,6 +203,7 @@ still work on the resulting hypertable.
 | `hypertable_id` | ID of the hypertable in TimescaleDB's internal catalog. |
 | `schema_name` | Schema name of the table converted to hypertable. |
 | `table_name` | Table name of the table converted to hypertable. |
+| `created` | True when the hypertable was created, false when `if_not_exists` was true and no hypertable was created. |
```
RobAtticus: same as above

PR #46, opened 2018-10-29
Add documentation for the time_bucket origin parameter

cevian:


api.md
```diff
@@ -857,6 +857,7 @@ or 1 hour.
 |Name|Description|
 |---|---|
 | `offset` | The time interval to offset all buckets by (interval) |
+| `origin` | A time for the edge of one bucket boundary (timestamp/timestamptz/date) |
```
RobAtticus: I'm not sure this is accurate (and it's not super clear). Since if I do a WHERE that restricts the start/end time, the `origin` is not going to be present as a bucket boundary.

I am having a hard time coming up with a better description though

PR #51, opened 2018-11-05
Fix instances of `Postgres` to `PostgreSQL`

RobAtticus:
Also added some line breaks for cleaner/consistent display in
editors.


getting-started/installation-homebrew.md
```diff
@@ -27,7 +27,7 @@ brew install timescaledb
 /usr/local/bin/timescaledb_move.sh
 ```
 
-#### Update Postgresql.conf
+#### Update postgresql.conf
```
solugebefola: Should we put the filename in backtick quotes?  Or maybe change it completely to "Update PostgreSQL config"?

PR #54, opened 2018-11-17
Fix warning and small clarifications with time_bucket

mfreed:


api.md
```diff
@@ -908,7 +908,8 @@ SELECT time_bucket('5 minutes', time, '-2.5 minutes') + '2.5 minutes'
   ORDER BY five_min DESC LIMIT 10;
 ```
 
-To shift the alignment of the buckets you can use the origin parameter.
+To shift the alignment of the buckets you can use the origin parameter
+(passed as a timestamp, timestamptz, or date type).
```
RobAtticus: Our more common style is to use all uppercase for types. There are a few exceptions in this page, so maybe you can change those too.

solugebefola: Actually, if those other changes apply to older versions than 1.0, it would be better to put in a separate PR.

api.md
```diff
@@ -943,6 +944,8 @@ to the server's timezone setting.
  with the ISO standard for Monday as the start of a week. This should only affect
```
svenklemm: This does not only affect multi-day calls but all calls where 1 day divided by period has a remainder.

mfreed: Will address in a separate PR on the underlying text.

PR #57, opened 2018-12-02
Add description of timescaledb_information.hypertable

svenklemm:
This probably needs some discusion how we want to integrate the view documentation

### DIFF
--- a/api.md
+++ b/api.md
@@ -22,6 +22,7 @@
 > - [set_adaptive_chunking](#set_adaptive_chunking)
 > - [set_chunk_time_interval](#set_chunk_time_interval)
 > - [set_number_partitions](#set_number_partitions)
+> - [timescaledb_information.hypertable](#timescaledb_information-hypertable)
 > - [show_tablespaces](#show_tablespaces)
 > - [time_bucket](#time_bucket)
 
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -25,6 +25,24 @@
 > - [show_tablespaces](#show_tablespaces)
 > - [time_bucket](#time_bucket)
 
+## TimescaleDB informational views [](timescaledb_information)
+
+## timescaledb_information.hypertable [](timescaledb_information.hypertable)
+
+Information about hypertables
```
dianasaur323: Can we match syntax of other helper functions? "Get information about hypertables."

api.md
```diff
@@ -25,6 +25,24 @@
...
+| `schema_name` | Schema name of the hypertable. |
+| `table_name` | Table name of the hypertable. |
+| `table_owner` | Owner of the hypertable. |
+| `num_dimensions` | Number of dimensions. |
+| `num_chunks` | Number of chunks. |
+| `table_bytes` |Disk space used by hypertable |
+| `index_bytes` |Disk space used by indexes|
+| `toast_bytes` |Disk space of toast tables|
+| `total_bytes` |Total disk space used by the specified table, including all indexes and TOAST data|
+
```
dianasaur323: Could we add sample use? 

Get all hypertables. SELECT * ....

Check if a table is a hypertable. SELECT * from ... where table_name = ...

mfreed: Agree with Diana, just to make clear that this is a "view" that one could query through a normal SQL.  Probably obvious to Sven, might not be to everybody.

api.md
```diff
@@ -25,6 +25,24 @@
 > - [show_tablespaces](#show_tablespaces)
 > - [time_bucket](#time_bucket)
 
+## TimescaleDB informational views [](timescaledb_information)
```
solugebefola: Is this meant to be a top level section (i.e. in the sidebar)?

api.md
```diff
@@ -25,6 +25,24 @@
...
+
+## timescaledb_information.hypertable [](timescaledb_information.hypertable)
```
solugebefola: I think a period in the anchor name may mess up the scrolling mechanism.

api.md
solugebefola: Yes, there should be a "Sample Usage" section.

api.md
solugebefola: Also this should be a header (####) to match the format of other functions.  Maybe "Available columns" as the title?

api.md
svenklemm: i wanted feedback on this so i didnt add any sections but diana wanted this to go under utility which is where i put it now

api.md
```diff
@@ -950,6 +951,48 @@ to the server's timezone setting.
...
+| `table_bytes` |Disk space used by hypertable |
+| `index_bytes` |Disk space used by indexes|
+| `toast_bytes` |Disk space of toast tables|
+| `total_bytes` |Total disk space used by the specified table, including all indexes and TOAST data|
+
+#### Sample Usage
+
+Get information about all hypertables.
+
+SELECT * FROM timescaledb_information.hypertable;
```
solugebefola: I think this line can come out.

api.md
```diff
@@ -950,6 +951,48 @@ to the server's timezone setting.
...
+
+```sql
+SELECT * FROM timescaledb_information.hypertable;
+ table_schema | table_name | table_owner | num_dimensions | num_chunks | table_size | index_size | toast_size | total_size 
+--------------+------------+-------------+----------------+------------+------------+------------+------------+------------
+ public       | metrics    | postgres    |              1 |          5 | 99 MB      | 96 MB      |            | 195 MB
```
solugebefola: Just to clarify what a user would get with multiple hypertables in the DB, can you put one or two more example rows here?

solugebefola:
Last thing, I assume this only applies to `1.0`?  If so, can you label the PR?

PR #69, opened 2018-12-26
Add documentation for time_bucket_gapfill

svenklemm:


api.md
```diff
@@ -919,6 +922,48 @@ The expected output:
 
 ---
 
+## interpolate() [](interpolate)
+
+The `interpolate` does linear interpolation for missing values.
```
solugebefola: "The 'interpolate' _function_ ..."?  Also, maybe the intro sentences can be combined as: "The interpolate function does linear interpolation for missing values.  It can only be used ..."

api.md
```diff
@@ -919,6 +922,48 @@ The expected output:
...
+
+|Name|Description|
+|---|---|
+| `value` | The value to carry forward (anyelement) |
+| `prev` | The lookup expression for values before the gapfill time range |
+| `next` | The lookup expression for values after the gapfill time range |
+
+#### Sample Usage [](interpolate-examples)
+
+Get the temperature every 5 minutes for each device over the 2 weeks interpolating for missing readings:
```
solugebefola: "over the _last_ two weeks..."?

api.md
```diff
@@ -919,6 +922,48 @@ The expected output:
...
+```sql
+SELECT
+  time_bucket_gapfill('5 minutes', time, now() - interval '2 week', now()) as time,
+  device_id,
+  interpolate(avg(temperature))
```
solugebefola: I don't quite understand how the argument syntax fits in with the info from the "required arguments" section.  is `avg(temperature)` the value?  The second example arguments also seem to follow a different syntax, where the value and the two lookup expressions seem to all be nested within the `avg` function?  Can you clarify how the `interpolate` function works with `avg`, perhaps in the intro or the required arguments section?  

From the examples, it seems that the required arguments are intended to be within an aggregation function?  It may require a non-standard "required arguments" section to fully describe the behavior.

api.md
```diff
@@ -950,6 +995,47 @@ SELECT device_id, time_bucket('5 minutes', time) as interval,
...
+| `prev` | The lookup expression for values outside of the gapfill time range |
+
+#### Sample Usage [](locf-examples)
+
+Get the temperature every 5 minutes for each device over the 2 weeks carrying forward the last value for missing readings:
+```sql
+SELECT
+  time_bucket_gapfill('5 minutes', time, now() - interval '2 week', now()) as time,
+  device_id,
+  locf(avg(temperature))
```
solugebefola: Same question as for `interpolate` with respect to the description of the required arguments and `prev` being an optional argument?

api.md
solugebefola: Oh, also, it seems that the `prev` and `next` functions are "optional arguments" rather than required?

api.md
```diff
@@ -1070,6 +1156,66 @@ to the server's timezone setting.
...
+#### Required Arguments [](time_bucket_gapfill-required-arguments)
+
+|Name|Description|
+|---|---|
+| `bucket_width` | A PostgreSQL time interval for how long each bucket is (interval) |
+| `time` | The timestamp to bucket (timestamp/timestamptz/date)|
+| `start` | The start of the gapfill period (timestamp/timestamptz/date)|
+| `end` | The end of the gapfill period (timestamp/timestamptz/date)|
+
+#### Required Arguments [](time_bucket_gapfill-integer-required-arguments)
```
solugebefola: I think the way that we dealt with this previously was to have a separate "units" section.  https://docs.timescale.com/v1.0/api#create_hypertable is an example (actually the only example currently)

api.md
```diff
@@ -1070,6 +1156,66 @@ to the server's timezone setting.
  multi-day calls to time_bucket. The old behavior can be reproduced by passing
  2000-01-01 as the origin parameter to time_bucket.
 
+## time_bucket_gapfill() [](time_bucket_gapfill)
+
```
solugebefola: Can you put in an intro that minimum references that `time_bucket_gapfill` works similarly to `time_bucket`, but with additional functionality?

api.md
svenklemm: only value is required, the lookup expressions are optional and only needed if you want to look beyond the gapfill time range.

api.md
svenklemm: fixed

api.md
svenklemm: fixed

api.md
svenklemm: Only value is required, the rest optional. I adjusted the docs accordingly.
The value can be any expression that results in a supported datatype. But since its an aggregation query you have to use an aggregate function on columns not referenced by the GROUP BY clause. Thats why you see the avg in the examples.

api.md
svenklemm: Hmm the supported datatypes are identical to 2 argument time_bucket and the 3rd and 4th argument is the same datatype as the 2nd.

api.md
svenklemm: done

api.md
solugebefola: The one thing I would say needs to be added to match with `time_bucket` is a `### For Integer Time Inputs` before the second set of required arguments, otherwise it's not clear why there are two sets of required arguments.

using-timescaledb/reading-data.md
```diff
@@ -336,26 +319,14 @@ value.
...
-    GROUP BY 1,2
-  ) t ON t.btime = period;
+  time_bucket_gapfill('5 min'::interval, time, now() - '2 weeks'::interval, now()) as time,
+  meter_id,
+  locf(avg(data_value)) AS data_value
+FROM my_hypertable
+WHERE
+  time > now() - '2 weeks'::interval
+  AND meter_id IN (1,2,3,4)
+GROUP BY 1,2
```
mfreed: Given the long example above with gapfilling, which ended just with filling in NULLs, I think it would be better if we stayed with the same example (trades), and also show the new output that this function would show.   

    date    | volume
------------+--------
 2017-09-30 | 11315
 2017-09-29 |  11315
 2017-09-28 |   8216
 2017-09-27 |   5591
 2017-09-26 |   9182
 2017-09-25 |  14359
 2017-09-24 |      9855
 2017-09-23 |   9855
 2017-09-24 |  9855
 2017-09-23 |  9855
 2017-09-22 |   9855


Two other small nits:
- Do you want to order by time desc or asc?
- I think the numerical syntax "GROUP BY 1,2" is less known by a lot of people.  Perhaps change to the column name?

api.md
```diff
@@ -950,6 +1000,53 @@ SELECT device_id, time_bucket('5 minutes', time) as interval,
...
+```
+
+Get the temperature every 5 minutes for each device over the 2 weeks carrying forward the last value for missing readings with out of bounds lookup
+```sql
+SELECT
+  time_bucket_gapfill('5 minutes', time, now() - interval '2 week', now()) as time,
+  device_id,
+  locf(
+    avg(temperature),
+    (SELECT temperature FROM metrics m2 WHERE m2.time < now( - interval '2 week') AND m.device_id = m2.device_id)
```
mfreed: m2.time < now() - interval '2 week'

api.md
```diff
@@ -919,6 +922,53 @@ The expected output:
...
+  time_bucket_gapfill('5 minutes', time, now() - interval '2 week', now()) as time,
+  device_id,
+  interpolate(avg(temperature))
+FROM metrics
+  WHERE time > now () - interval '2 week'
+GROUP BY 1,2
+ORDER BY 1 DESC;
+```
+
+Get the temperature every 5 minutes for each device over the 2 weeks interpolating for missing readings with lookup queries for values before and after the gapfill time range:
```
mfreed: I don't think the average reader will understand what the before/after queries actually do.  Think you need some text around this explaining the issue before giving the syntax, i.e., the gapfill query defines a time-range, but what if you don't actually have any values at the beginning of the time range, and need to actually search *outside* of the time-range for the right values to interpolate from. This query defines how you perform that search, etc.

I think many of these queries should also have sample output to better explain...

api.md
```diff
@@ -950,6 +1000,53 @@ SELECT device_id, time_bucket('5 minutes', time) as interval,
...
+  time_bucket_gapfill('5 minutes', time, now() - interval '2 week', now()) as time,
+  device_id,
+  locf(avg(temperature))
+FROM metrics
+  WHERE time > now () - interval '2 week'
+GROUP BY 1,2
+ORDER BY 1 DESC;
+```
+
+Get the temperature every 5 minutes for each device over the 2 weeks carrying forward the last value for missing readings with out of bounds lookup
```
mfreed: See the above comment on interpolate about explaining need for "out-of-bounds" lookup (nit: note hyphenation), as well as example output.

Side note:  It seems like we really want a tutorial here showing timebucket alone, timebucket gapfill, timebucket gapfill + locf.   Should this be organized better in the reading-data section (as a contiguous mini-section), then these API docs could point there?  cc: @solugebefola

api.md
```diff
@@ -919,6 +922,53 @@ The expected output:
 
 ---
 
+## interpolate() [](interpolate)
+
+The `interpolate` function does linear interpolation for missing values. It can only be used in an aggregation query with `time_bucket_gapfill`.
+The `interpolate` function call cannot be nested inside other function calls.
```
dianasaur323: I think we can add a link to the `time_bucket_gapfill` docs when we mention it here?

api.md
dianasaur323: I wonder if we want to include this discussion above in the introduction section as well. Think it reads better to present it there, and then show an example here

api.md
```diff
@@ -950,6 +1000,53 @@ SELECT device_id, time_bucket('5 minutes', time) as interval,
  alternative to an `ORDER BY time DESC LIMIT 1` clause to find the
  latest value (which will use indexes).
 
+## locf() [](locf)
+
+The `locf` function allows you to carry the last seen value in an aggregation group forward.
+It can only be used in an aggregation query with `time_bucket_gapfill`.
```
dianasaur323: Also add a link to `time_bucket_gapfill` here if possible?

api.md
```diff
@@ -950,6 +1000,53 @@ SELECT device_id, time_bucket('5 minutes', time) as interval,
...
+
+|Name|Description|
+|---|---|
+| `value` | The value to carry forward (anyelement) |
+
+#### Optional Arguments [](locf-optional-arguments)
+
+|Name|Description|
+|---|---|
+| `prev` | The lookup expression for values outside of the gapfill time range (anyelement) |
```
dianasaur323: you had (record) here instead of (anyelement) in interpolate. I'm not sure I completely understand what (record) means.

api.md
```diff
@@ -1070,6 +1167,64 @@ to the server's timezone setting.
...
+```sql
+SELECT
+  time_bucket_gapfill('5 minutes', time, now() - interval '1 week', now()) AS time,
+  avg(value)
+FROM metrics
+GROUP BY 1
+ORDER BY 1;
+```
+
+Simple gap fill query with 5 minute interval with LOCF:
```
dianasaur323: Could we do `LOCF (last object carried forward)` Some people don't understand LOCF for some reason.

api.md
```diff
@@ -1070,6 +1167,64 @@ to the server's timezone setting.
...
+|Name|Description|
+|---|---|
+| `bucket_width` | integer interval for how long each bucket is (int2/int4/int8) |
+| `time` | The timestamp to bucket (int2/int4/int8)|
+| `start` | The start of the gapfill period (int2/int4/int8)|
+| `end` | The end of the gapfill period (int2/int4/int8)|
+
+#### Sample Usage [](time_bucket_gapfill-examples)
+
+Simple gap fill query with 5 minute interval:
```
dianasaur323: Could we update this to match the text you have for last and LOCF? Get the metric value every 5 minutes for each device over the last week....

api.md
dianasaur323: Also, same comment re matching your sentence description for LOCF

api.md
svenklemm: record is a composite datatype, since interpolate needs a time:value pair to properly interpolate missing values, while locf takes the previous value verbatim. that's also why locf works with any datatype while interpolate only works with a subset

using-timescaledb/reading-data.md
svenklemm: The order is DESC which you can also see in the result. I think the numerical syntax is superior because otherwise you have to repeat the expression or use an alias which might give unexpected results if you have columns by the same name. The current version of the document also uses that syntax.

api.md
svenklemm: The meaning of those values is explained in the table with the arguments now.

api.md
```diff
@@ -950,6 +1029,79 @@ SELECT device_id, time_bucket('5 minutes', time) as interval,
...
+| `value` | The value to carry forward (anyelement) |
+
+#### Optional Arguments [](locf-optional-arguments)
+
+|Name|Description|
+|---|---|
+| `prev` | The lookup expression for values before gapfill start (anyelement) |
+
+The `prev` expression will only be evaluated when no previous value is returned by the outer query.
+
```
dianasaur323: Can we also add that little snippet here: This happens when the first bucket in the queried time range is empty or does not otherwise contain a valid value?

api.md
```diff
@@ -1070,6 +1222,102 @@ to the server's timezone setting.
  multi-day calls to time_bucket. The old behavior can be reproduced by passing
  2000-01-01 as the origin parameter to time_bucket.
 
+## time_bucket_gapfill() [](time_bucket_gapfill)
+
+The `time_bucket_gapfill` function works similar to `time_bucket` but also activates gap
+filling for the interval between `start` and `end`. It can only be used with an aggregation
+query.
+
```
dianasaur323: Small note about how gapfilling handles out-of-bounds?

api.md
```diff
@@ -919,6 +922,82 @@ The expected output:
...
+| `value` | The value to interpolate (int2/int4/int8/float4/float8) |
+
+#### Optional Arguments [](interpolate-optional-arguments)
+
+|Name|Description|
+|---|---|
+| `prev` | The lookup expression for values before the gapfill time range (record) |
+| `next` | The lookup expression for values after the gapfill time range (record) |
+
+The `prev` and `next` expression will only be evaluated when no suitable value is returned by the outer query.
```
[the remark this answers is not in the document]
mfreed: Agree with @dianasaur323 that just something more is needed to clarify what prev/next are used for.  Something perhaps like (please review for correctness):

Because the interpolation function relies on having values before and after each bucketed period to compute the interpolated value, it can have difficulty in calculating the interpolation for the first and last time bucket if those buckets do not otherwise contain valid values.  For example, the interpolation would require looking *before* this first time bucket period, yet the query's outer time predicate `WHERE time > ...` normally restricts the function to only evaluate values *within* this time range.  Thus, the `prev` and `next` expression tell the function how to look for values *outside* of the range specified by the time predicate; these expressions will only be evaluated when no suitable value is returned by the outer query.

api.md
mfreed: Agree.  Or perhaps just copy explanation or reference back to longer explanation given in interpolation?

api.md
```diff
@@ -919,6 +922,82 @@ The expected output:
...
+SELECT
+  time_bucket_gapfill('1 day', time, now() - interval '1 week', now()) as time,
+  device_id,
+  interpolate(avg(temperature) AS value,
+    (SELECT (time,temperature) FROM metrics m2 WHERE m2.time < now() - interval '1 week' AND m.device_id = m2.device_id),
+    (SELECT (time,temperature) FROM metrics m2 WHERE m2.time > now() AND m.device_id = m2.device_id)
+  )
+FROM metrics
+WHERE time > now () - interval '1 week'
+GROUP BY 1,2
```
mfreed: I don't think "GROUP BY 1, 2 ORDER BY 1" is less clear to users as column names.  Can we change this (and all below) to 

"GROUP BY time, device_d ORDER BY time" ?

(I noticed below that you responded:  "I think the numerical syntax is superior because otherwise you have to repeat the expression or use an alias which might give unexpected results if you have columns by the same name. The current version of the document also uses that syntax."  Maybe we can discuss with team if we have overall preference with docs.)

api.md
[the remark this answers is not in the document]
svenklemm: @dianasaur323 its almost true, there is no valid value constraint, explicit NULLS returned from outer query won't be silently ignored, if you dont want them you need to exclude them in WHERE clause

api.md
svenklemm: done

api.md
svenklemm: done

using-timescaledb/reading-data.md
svenklemm: changed

api.md
```diff
@@ -919,6 +922,92 @@ The expected output:
...
+enough data to calculate the interpolation for the first and last time bucket
+if those buckets do not otherwise contain valid values.
+For example, the interpolation would require looking before this first
+time bucket period, yet the query's outer time predicate WHERE time > ...
+normally restricts the function to only evaluate values within this time range.
+Thus, the `prev` and `next` expression tell the function how to look for
+values outside of the range specified by the time predicate.
+These expressions will only be evaluated when no suitable value is returned by the outer query.
+This happens when the first and / or last bucket in the queried time range is empty.
+The returned record for `prev` and `next` needs to be a time, value tuple.
```
dianasaur323: I am sooo sorry for the additional comment, could we move "The returned record for `prev` and `next` up to `interpolate` call to go right after the optional arguments table? I think it's nice to say what type those arguments are right up front.

dianasaur323: Although now that I read it again, I'm not longer convinced. What do you think?

mfreed: Not sure I follow @dianasaur323 's comment.  One grammar nit though, perhaps:

`These expressions will only be evaluated when no suitable value is returned by the outer query.
This happens when the first and / or last bucket in the queried time range is empty.`

should be changed to:

`These expressions will only be evaluated when no suitable value is returned by the outer query (i.e., the first and/or last bucket in the queried time range is empty).`

as otherwise a bit repetitive.

api.md
mfreed: Same nit:

The `prev` expression will only be evaluated when no previous value is returned by the outer query (i.e., the first bucket in the queried time range is empty).

api.md
svenklemm: since mike prefers current version i left the order

PR #70, opened 2018-12-26
Remove reference to docker-run.sh

svenklemm:
docker-run.sh seems to be outdated with the recent docker changes


solugebefola:
Please label with the earliest version :)

RobAtticus:
@solugebefola Can probably update the file as far back as you're willing to go, since once we delete the file the link on all branches will be dead

PR #71, opened 2019-01-01
Improve notation of link in the in the Hello NYC tutorial

komainu8:
Dear developers.

In my opinion, we easily overlook a link of sample data (e nyc_data.tar.gz).
Because this link within backquote, it is treated as inline code and we don't look like a link.

So, would it be possible to separate a link as in this PR.

### DIFF
--- a/tutorials/tutorial-hello-nyc.md
+++ b/tutorials/tutorial-hello-nyc.md
@@ -34,7 +34,9 @@ CREATE DATABASE nyc_data;
 CREATE EXTENSION IF NOT EXISTS timescaledb CASCADE;
 ```
 
-Now, download the file [`nyc_data.tar.gz`][nyc_data].
+Next, download the file from the below link:
+
+[:DOWNLOAD_LINK: `nyc_data.tar.gz`][nyc_data]
 
 Then, follow these steps:
 


tutorials/tutorial-hello-nyc.md
```diff
@@ -34,7 +34,9 @@ CREATE DATABASE nyc_data;
 CREATE EXTENSION IF NOT EXISTS timescaledb CASCADE;
 ```
 
-Now, download the file [`nyc_data.tar.gz`][nyc_data].
+Now, download the file from the below link.
+
+[`nyc_data.tar.gz`][nyc_data].
```
mfreed: To be consistent with the rest of the article and to correct grammatical issue, perhaps update to:
-----

Next, download the file from the below link:

[` nyc_data.tar.gz `][nyc_data]

mfreed: Like the change, but to be consistent with the rest of the article and to correct grammatical issue, perhaps update to:

Next, download the file from the below link:

[` nyc_data.tar.gz `][nyc_data]

solugebefola:
I would recommend using the :DOWNLOAD_LINK: symbol to add a download link icon to the link on the page.  See the README and https://github.com/timescale/docs.timescale.com-content/blob/master/tutorials/other-sample-datasets.md to see how it's entered in the markdown file, and https://docs.timescale.com/v1.1/tutorials/other-sample-datasets to see how it looks on the docs site.

komainu8:
@mfreed @solugebefola 
Thank you for your comments.
I've fixed them.

RobAtticus:
You could probably squash these commits into a single commit:
```
Improve link clarity in Hello NYC tutorial

The link to the dataset for the Hello NYC tutorial was a bit hidden
because it looked like code rather than an actual link. By moving
the link to a separate link and adding the download icon, it should
stand out as a link so users won't miss it.
```

komainu8:
@RobAtticus Thank you for your comment.
I have put these commits in one.

PR #72, opened 2019-01-03
Update a list of schemas for the latest version

komainu8:
Dear, developers.

We updated result of a `\dn` command in the Hello NYC tutorial.
Because the result of a `\dn` command in a tutorial is different from the result of it in the latest version.


solugebefola:
please add a label to indicate the earliest database version that this applies to :)

PR #74, opened 2019-01-10
Clarify the lack of max_parallel_workers setting in PG9.6

RobAtticus:
PostgreSQL 9.6 does not have the config limit of max_parallel_workers
which limits the number of parallel processes that can be used for
parallel queries. As such, it does not need to be set.

Also fix the line breaks to be nicer to read.

### DIFF
--- a/getting-started/configuring.md
+++ b/getting-started/configuring.md
@@ -19,25 +19,30 @@ PgTune may also be helpful.
 
 ## Worker settings [](workers)
 
-PostgreSQL utilizes worker pools to provide the required workers needed to support
-both live queries and background jobs. If you do not configure these settings,
-you may observe performance degradation on both queries and background jobs.
+PostgreSQL utilizes worker pools to provide the required workers needed to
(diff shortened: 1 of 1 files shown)


getting-started/configuring.md
```diff
@@ -19,25 +19,30 @@ PgTune may also be helpful.
...
+TimescaleDB background workers are configured using the
+`timescaledb.max_background_workers` setting. You should configure this
+setting to the sum of your total number of databases and the
+total number of concurrent background workers you want running at any given
+point in time. You need a background worker allocated to each database to run
+a lightweight scheduler that schedules jobs. On top of that, any additional
+workers you allocate here will run background jobs when needed.
+
+For larger queries, PostgreSQL automatically uses parallel workers if they are
+available. To configure this on PG10+, use the `max_parallel_workers`
```
svenklemm: Should we use PostgreSQL 9.6 and PostgreSQL 10 instead of PG9.6 and PG10 here?

PR #76, opened 2019-01-10
change order of supported pg versions

davidkohn88:


davidkohn88:
oh, we just merged something that replaced most of those with :pg_version_num: or something. shoudl we just do that there?

svenklemm:
ah nvm then, looks like we only list latest on docker page now which is fine too, dont feel so strongly about this then

RobAtticus:
Yeah with `pg-version` it should lead to cleaner pages. I don't think switching PG11 to the front is a worthwhile change

PR #77, opened 2019-01-10
Fix statements that partitions must be prefixes of UNIQUE

cevian:
Partitions have to be part of a unique index but it is not
necessary for them to be in the prefix. Fix language to reflect
that.

### DIFF
--- a/using-timescaledb/schema-management.md
+++ b/using-timescaledb/schema-management.md
@@ -115,13 +115,11 @@ matching your predicate, but in each one, your query also grabs a
 (potentially large) subset of the values, rather than just one
 distinct one.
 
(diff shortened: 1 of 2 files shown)


using-timescaledb/schema-management.md
```diff
@@ -116,12 +116,10 @@ matching your predicate, but in each one, your query also grabs a
 distinct one.
 
 >:TIP: [](unique_indexes) To a define an index as UNIQUE or PRIMARY KEY, the time column and, if it
```
solugebefola: Not part of the PR, but can you take out the extra "a" before "define" here?

solugebefola:
please label with the earliest database version that this PR applies to.

PR #79, opened 2019-01-16
Deprecate adaptive chunking.

cevian:
Adaptive chunking is a failed experiment. The adaptation we introduced
does not work well on backfill scenarios and even slightly out-of-order
data and lead to unpredictable performance in many scenarios.
Since this feature was in BETA we decided to deprecate it
altogether to ensure overall stability in the system.


api.md
```diff
@@ -654,8 +621,10 @@ SELECT set_chunk_time_interval('conditions', 86400000);
 
 ---
 
-## set_adaptive_chunking() [](set_adaptive_chunking)
-Changes the settings for [adaptive chunking (BETA)][adaptive-chunking]. The
+## DEPRECATED set_adaptive_chunking() [](set_adaptive_chunking)
+
+The adaptive chunking feature is now deprecated and should not be used.
```
mfreed: Perhaps this first line ("The adaptive chunking feature is now deprecated and should not be used.") should be a :WARNING: ?

api.md
```diff
@@ -665,47 +634,13 @@ are created; it does not modify the intervals of existing chunks.
 |Name|Description|
 |---|---|
 | `hypertable` | Identifier of hypertable to update the settings for.|
-| `chunk_target_size` | The target size of a chunk (including indexes) in `kB`, `MB`, `GB`, or `TB`. Setting this to `estimate` or a non-zero chunk size, e.g., `2GB` will enable [adaptive chunking (BETA)][adaptive-chunking]. The `estimate` setting will estimate a target chunk size based on system information. Adaptive chunking is disabled by default. |
+| `chunk_target_size` | The target size of a chunk (including indexes) in `kB`, `MB`, `GB`, or `TB`. Setting this to `estimate` or a non-zero chunk size, e.g., `2GB` will enable adaptive chunking. The `estimate` setting will estimate a target chunk size based on system information. Adaptive chunking is disabled by default. |
```
mfreed: No mention of deprecation?

api.md
```diff
@@ -665,47 +634,13 @@ are created; it does not modify the intervals of existing chunks.
...
-Set the target chunk size to `1GB` on the `conditions` hypertable, using
-a custom chunk sizing function (enabling adaptive chunking if
-previously disabled):
-
-```sql
-SELECT * FROM set_adaptive_chunking('conditions', '1GB', 'custom_calculate_chunk_interval');
-    chunk_sizing_func            | chunk_target_size
----------------------------------+-------------------
- custom_calculate_chunk_interval |        1073741824
+| `chunk_sizing_func` | Allows setting a custom chunk sizing function for adaptive chunking. The built-in chunk sizing function will be used by default. Note that `chunk_target_size` needs to be set to use this function. |
```
mfreed: No mention of deprecation?

cevian:
@solugebefola I don't think we want to remove it from the top list in this pr.

PR #89, opened 2019-01-28
Update CentOS instructions to use packagecloud

RobAtticus:
Packagecloud allows us to add first-party dependencies, like our
timescaledb-tools package, as well as integrate more naturally
with the package manager.

### DIFF
--- a/getting-started/installation-yum.md
+++ b/getting-started/installation-yum.md
@@ -7,7 +7,7 @@ This will install both TimescaleDB *and* PostgreSQL via `yum`
 
 #### Prerequisites
 
-- CentOS 7 (or Fedora/RHEL equivalent) or later
+- RHEL/CentOS 7 (or Fedora equivalent) or later
 
 #### Build & Install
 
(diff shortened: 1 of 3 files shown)


RobAtticus:
Could potentially backport to more but let's go with 1.0/1.1/1.2 for now

solugebefola:
Will the pg11 instruction apply to 1.0 and earlier?

getting-started/installation-yum.md
```diff
@@ -20,21 +20,34 @@ sure to remove non-`yum` installations before using this method.
 You'll need to [download the correct PGDG from PostgreSQL][pgdg] for
 your operating system and architecture and install it:
 ```bash
-# Download PGDG for PostgreSQL 9.6, e.g. for Fedora 24:
-sudo yum install -y https://download.postgresql.org/pub/repos/yum/9.6/fedora/fedora-24-x86_64/pgdg-fedora96-9.6-3.noarch.rpm
+# Download PGDG for PostgreSQL 11, e.g. for CentOS 7:
```
solugebefola: Doesn't this only apply to `v1.2` of the db and later?

solugebefola: Note: for `1.0` `1.1` branches, the pg version needs to be `10`

PR #91, opened 2019-01-29
Fix typo

solugebefola:


mfreed:
Are the only fixes the deletion of extraneous whitespace at the end of some lines?  Maybe just update git commit message?

PR #92, opened 2019-01-29
Fix typo

solugebefola:


RobAtticus:
If you change it to bash it does some weird highlighting on the output, I think `text` is preferred

solugebefola:
`text` breaks Jenkins.  It's not a known language for highlightjs (at least the version that's in the repo)

solugebefola:
can just leave without language

solugebefola:
Have to test, but this was the fix I had to use to fix the 1.2 branch of docs

RobAtticus:
Okay, change to blank then, bash ends up looking weird.

PR #95, opened 2019-01-30
Add tip for installing Apache binaries

RobAtticus:
Some users may want to continue using only Apache 2.0 bits, so we
should include tips on how to install those versions for the
channels that support it (Homebrew, Docker, Linux).


RobAtticus:
Ah forgot there were tips already

PR #98, opened 2019-01-31
Add timescaledb-tune to install pages

RobAtticus:
Now that our binary packages typically include timescaledb-tune,
we can remove the manual editting of postgresql.conf and replace it
with a call to timescaledb-tune. For new users, this is probably
easier and less error prone.

### DIFF
--- a/getting-started/installation-apt-debian.md
+++ b/getting-started/installation-apt-debian.md
@@ -37,22 +37,20 @@ sudo apt-get update
 sudo apt-get install timescaledb-postgresql-:pg_version:
 ```
 
-#### Update `postgresql.conf`
+#### Configure your database
 
->:TIP: The usual location of `postgres.conf`
-is `/etc/postgresql/:pg_version:/main/postgresql.conf`, but this may vary
-depending on your setup. If you are unsure where your `postgresql.conf` file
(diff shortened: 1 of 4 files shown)


getting-started/installation-apt-debian.md
```diff
@@ -37,22 +37,18 @@ sudo apt-get update
...
->:TIP: The usual location of `postgres.conf`
-is `/etc/postgresql/:pg_version:/main/postgresql.conf`, but this may vary
-depending on your setup. If you are unsure where your `postgresql.conf` file
-is located, you can query PostgreSQL with any database client (e.g., `psql`)
-using `SHOW config_file;`.
-
-You will need to edit your `postgresql.conf` file to include
-necessary libraries:
+There are a [variety of settings that can be configured][config] for your
+new database. The easiest way to get started is to run `timescaledb-tune`,
```
dianasaur323: Only comment is that in this paragraph, it's not clear that you MUST use timescale-tune to put in the shared_preload_library, so maybe a slight re-word. You will need to update your `postgresql.conf` file to include necessary libraries. The easiest way to get started is to run `timescale-tune`, which is installed by default when using `apt`

RobAtticus: Updated.

Issue #100, opened 2019-02-02
Format alter_job_schedule query


solugebefola:
`alter_job_schedule` sample usage query should be broken into several lines.

PR #101, opened 2019-02-02
Format queries to reduce width

solugebefola:


solugebefola:
Addresses https://github.com/timescale/docs.timescale.com-content/issues/100

PR #108, opened 2019-03-08
Change `Deleting a Hypertable` anchor to #drop

askmike:
Fixes the submenu which is pointing to `https://docs.timescale.com/v0.12/using-timescaledb/hypertables#drop` instead of #delete.


svenklemm:
@askmike thank you for the bugfix

PR #111, opened 2019-03-14
Improve clarity in drop_chunks examples

RobAtticus:
Some of these sentences read a bit ambiguously and caused confusion
for some users. Also, add a warning so users are careful about
interval type usage.

### DIFF
--- a/api.md
+++ b/api.md
@@ -535,6 +535,11 @@ The `older_than` and `newer_than` parameters can be specified in two ways:
     explicitly given as a TIMESTAMP / TIMESTAMPTZ / DATE or as a
     SMALLINT / INT / BIGINT. The choice of timestamp or integer must follow the type of the hypertable's time column.
 
+
+>:WARNING: When using just an interval type, the function assumes that
+you are are removing things _in the past_. If you want to remove data
+in the future (i.e., erroneous entries), use a timestamp.
+
 When both arguments are used, the function returns the intersection of the resulting two ranges. For example,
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -535,6 +535,12 @@ The `older_than` and `newer_than` parameters can be specified in two ways:
     explicitly given as a TIMESTAMP / TIMESTAMPTZ / DATE or as a
     SMALLINT / INT / BIGINT. The choice of timestamp or integer must follow the type of the hypertable's time column.
 
+
+>:WARNING: When using just an interval type, the assumption should be you
+are removing things _in the past_. If you want to remove data in the
+future (i.e., erroneous entries), use a timestamp or remember to use
+the interval `+ now()`.
+
```
svenklemm: I think this should be a bit stronger, maybe something like this:
The assumption for drop_chunks is that you want to delete from the past when using INTERVAL, if you want to delete in the future use absolute times.

api.md
```diff
@@ -535,6 +535,12 @@ The `older_than` and `newer_than` parameters can be specified in two ways:
     explicitly given as a TIMESTAMP / TIMESTAMPTZ / DATE or as a
     SMALLINT / INT / BIGINT. The choice of timestamp or integer must follow the type of the hypertable's time column.
 
+
+>:WARNING: When using just an interval type, the assumption should be you
```
solugebefola: "When using just an interval type, the function assumes that you are..."?

RobAtticus:
Updated both

PR #113, opened 2019-03-17
Remove reference to generate_series from gapfill section

svenklemm:


mfreed:
Sorry to be pedantic, but I think the first reference above still needs to be updated, i.e., we aren't outputting 0, and we aren't performing a JOIN in the query.
```
To instead output a value of "0" for each missing day, one can use the following query, which joins data from the prices table (akin to the query above) with a generated stream of dates.
```
How about something like the following (around current line 247):
```
To instead include time records for each missing day, one can use the following TimescaleDB function `time_bucket_gapfill`, which serves to generate a series of time buckets according to some interval (here, `1 day`) across a specified period.
```

PR #115, opened 2019-03-24
Document gapfill inferring start and finish from WHERE clause

svenklemm:


solugebefola:
What happens if one doesn't include a finish in the WHERE clause?

svenklemm:
start and finish still need to be passed either as argument or in WHERE clause, it will error if you dont

PR #125, opened 2019-04-20
Add decorators to functions, clean up horizontal breaks in API

mfreed:


using-timescaledb/reading-data.md
```diff
@@ -203,7 +203,7 @@ This query will output data in the following form:
  garage     | 10080 | {0,2679,957,2420,2150,1874,0}
 ```
 
-### Gap Filling [](gap-filling)
+### Gap Filling :timescale_function: :community_function: [](gap-filling)
```
solugebefola: Aren't all of the functions on the API page timescale functions?

using-timescaledb/reading-data.md
```diff
@@ -303,7 +303,7 @@ This query will output data of the form:
  2018-03-09 20:53:20+00 |  1037.78
 ```
 
-### Last Observation Carried Forward [](locf)
+### Last Observation Carried Forward :timescale_function: :community_function: [](locf)
```
solugebefola: see above

api.md
```diff
@@ -852,8 +853,6 @@ are meant to implement data retention or perform tasks that will improve query
 performance on older chunks. Each policy is assigned a scheduled job
 which will be run in the background to enforce it.
 
-
-
 ## add_drop_chunks_policy() :enterprise_function: [](add_drop_chunks_policy)
```
solugebefola: Is there a reason you didn't put a divider here?

api.md
```diff
@@ -1041,7 +1040,7 @@ WHERE hypertable = 'conditions';
 ```
 reschedules the reorder policy job for the `conditions` table so that it runs every two days.
 
-
+---
 ## Analytics [](analytics)
 
 ## first() [](first)
```
solugebefola: Did you not put a divider here because `first` comes immediately after the `Analytics` header?

using-timescaledb/reading-data.md
mfreed: This isn't the API page, this is the using timescale page.  In a bunch of the other examples, we flag when something is native postgres or timescale, so I'm just making consistent here.  Given that the other functions were things in OSS (e.g., `time_bucket`), I thought it was also worth flagging when in community/enterprise.

api.md
mfreed: Was the first function under a section head, the other two section headers don't have the first function with a divider either...  (Happy to do that instead)

PR #130, opened 2019-04-25
Add docs for our CREATE INDEX

JLockerman:
This commit adds docs for
CREATE INDEX ON <hypertable> WITH (timescaledb.transaction_per_chunk)

Currently this is being added to the api docs, until we finalize how
commands using `WITH` clauses will be laid out.

### DIFF
--- a/api.md
+++ b/api.md
@@ -11,6 +11,7 @@
 > - [chunk_relation_size](#chunk_relation_size)
 > - [chunk_relation_size_pretty](#chunk_relation_size_pretty)
 > - [create_hypertable](#create_hypertable)
+> - [create index (transaction per chunk)](#create_index)
 > - [create view (continuous aggregate)](#cagg_create_view)
 > - [detach_tablespace](#detach_tablespace)
 > - [detach_tablespaces](#detach_tablespaces)
@@ -415,6 +416,39 @@ field).  A very large number of such partitions leads both to poorer
 per-partition load balancing (the mapping of items to partitions using
 hashing), as well as much increased planning latency for some types of
 queries.
+
+---
+
+## CREATE INDEX (Transaction Per Chunk) [](create_index)
+
+```SQL
+CREATE INDEX ... WITH (timescaledb.transaction_per_chunk, ...);
+```
+
+This option extends [`CREATE INDEX`][postgres-createindex] with the
+ability to use a separate transaction for each chunk it creates an
+index on, instead of using a single transaction for the entire hypertable.
+This allows `INSERT`s, and other operations to to be performed concurrently
+during most of the duration of the `CREATE INDEX` command.
+While the index is being created on an individual chunk, it functions as
+if a regular `CREATE INDEX` were called on that chunk, however other chunks are
+completely un-blocked.
+
+>:TIP: This version of `CREATE INDEX` can be used as an alternative to `CREATE INDEX CONCURRENTLY`, which is not currently supported on hypertables.
+
+>:WARNING: If the operation fails partway through, indexes may not be created on all hypertable chunks. If this occurs, the index on the root table of the hypertable will be marked as invalid (this can be seen by running `\d+` on the hypertable). The index will still work, and will be created on new chunks, but if you wish to ensure _all_ chunks have a copy of the index, drop and recreate it.
+
+#### Sample Usage [](create_index-examples)
+
+Anonymous index
+```SQL
+CREATE INDEX ON conditions(time, device_id) WITH (timescaledb.transaction_per_chunk);
+```
+Other index methods
+```SQL
+CREATE INDEX ON conditions(time, location) USING brin
+  WITH (timescaledb.transaction_per_chunk);
+```
 
 ---
 
@@ -855,8 +889,8 @@ runs a reorder on the `_timescaledb_internal._hyper_1_10_chunk` chunk using the 
 
 ---
 ## Continuous Aggregates :community_function: [](continuous-aggregates)
-TimescaleDB allows users the ability to automatically recompute aggregates 
-at predefined intervals and materialize the results. This is suitable for 
+TimescaleDB allows users the ability to automatically recompute aggregates
+at predefined intervals and materialize the results. This is suitable for
 frequently used queries.
 
 *  [CREATE VIEW](#cagg_create_view)
@@ -864,16 +898,16 @@ frequently used queries.
 *  [REFRESH](#cagg_refresh_view)
 *  [DROP VIEW](#cagg_drop_view)
 
-## CREATE VIEW (Continuous Aggregate) [](cagg_create_view) 
+## CREATE VIEW (Continuous Aggregate) [](cagg_create_view)
 `CREATE VIEW` statement is used to create continuous aggregates.
 
 The syntax is:
 ``` sql
-    CREATE VIEW <view_name> [ ( column_name [, ...] ) ] 
+    CREATE VIEW <view_name> [ ( column_name [, ...] ) ]
     WITH ( timescaledb.continuous [, timescaledb.<option> = <value> ] )
     AS
     <select_query>
-    
+
     <select_query> is of the form :
     SELECT <grouping_exprs>, <aggregate_functions>
     FROM <hypertable>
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -412,6 +413,23 @@ queries.
...
+CREATE INDEX ON <hypertable> WITH (timescaledb.transaction_per_chunk);
+```
+
+This functions much like a regular `CREATE INDEX` except that it uses
+a separate transaction for each chunk it creates an index on, instead
+of using a single transaction for the entire hypertable. This allows
+`INSERT`s, and other operations to to be performed concurrently during
+most of the duration of the `CREATE INDEX`.
+
+>:TIP: This version of `CREATE INDEX` can be used as an alternative to `CREATE INDEX CONCURRENTLY`
```
dianasaur323: Can we add a sentence here on how the two are different?

api.md
```diff
@@ -412,6 +413,23 @@ queries.
...
+
+>:WARNING: If the operation fails partway through, indexes may not be created on some of the hypertables chunks. If this occurs, the index on the root table of the hypertable will be marked as invalid. The index will still work, and will be created on new chunks, but if you wish to ensure _all_ chunks have a copy of the index, drop and recreate it.
+---
```
dianasaur323: How can a user check if the index is marked as invalid?

api.md
JLockerman: @dianasaur323 hmmm, the most important one is that `CREATE INDEX CONCURRENTLY` doesn't work on hypertables at the moment.

JLockerman: If we do implement it, I think the difference would end up being that this version blocks INSERTs to the last chunk for a slightly longer time
(honestly, I do wonder if we should have just had `CREATE INDEX CONCURRENTLY` call this)

JLockerman: Had some more discussion with @davidkohn88 about this: since we don't implement `CREATE INDEX CONCURRENTLY` on hypertables at the moment, a users options are only: use regular `CREATE INDEX`, use this, or create indexes on chunks manually. As such, the main use for this tip to point users searching for how to run `CREATE INDEX CONCURRENTLY` on a hypertable at this instead. As such, I don't think a comparison with `CREATE INDEX CONCURRENTLY` is really relevant here, and is more suited for the blog post.

JLockerman: As an aside, should we add a hint to the error from `CREATE INDEX CONCURRENTLY` that says use `CREATE INDEX ... WITH (timescaledb. transaction_per_chunk)` instead?

JLockerman:
@dianasaur323 I pushed some changes to the doc, describing it more as an extension to `CREATE INDEX` instead of its own thing, would you re-review?

api.md
dianasaur323: The little note you put in there in the new version is perfect!

dianasaur323: Adding a note might be good for UX, but I don't feel particularly strongly about it. Think that's a nice clean-up UX feature.

api.md
```diff
@@ -412,6 +413,39 @@ queries.
...
+index on, instead of using a single transaction for the entire hypertable.
+This allows `INSERT`s, and other operations to to be performed concurrently
+during most of the duration of the `CREATE INDEX`.
+(While the the index is being created on an individual chunk, it functions as
+if a regular `CREATE INDEX` was called on that chunk, however other chunks are
+completely un-blocked)
+
+>:TIP: This version of `CREATE INDEX` can be used as an alternative to `CREATE INDEX CONCURRENTLY`, which is not currently supported on hypertables.
+
+>:WARNING: If the operation fails partway through, indexes may not be created on some of the hypertables chunks. If this occurs, the index on the root table of the hypertable will be marked as invalid (this can be seen by running `\d+` on the hypertable). The index will still work, and will be created on new chunks, but if you wish to ensure _all_ chunks have a copy of the index, drop and recreate it.
```
solugebefola: "indexes may not be created on all hypertable chunks"?

solugebefola: at minimum "hypertables" >> "hypertable's"

solugebefola:
Please label the PR with the earliest db version number to which the PR applies.

api.md
```diff
@@ -412,6 +413,39 @@ queries.
 
 ---
 
+## CREATE INDEX [](create_index)
+
+```SQL
+CREATE INDEX ... WITH (timescaledb.transaction_per_chunk, ...);
+```
+
+This option extends [`CREATE INDEX`](https://www.postgresql.org/docs/current/sql-createindex.html)
```
solugebefola: Please convert to footnote-style link

api.md
```diff
@@ -412,6 +413,39 @@ queries.
 
 ---
 
+## CREATE INDEX [](create_index)
+
+```SQL
+CREATE INDEX ... WITH (timescaledb.transaction_per_chunk, ...);
+```
+
+This option extends [`CREATE INDEX`](postgres-createindex) with the
```
solugebefola: the round brackets for the link need to be square brackets

api.md
```diff
@@ -412,6 +413,39 @@ queries.
...
+```SQL
+CREATE INDEX ... WITH (timescaledb.transaction_per_chunk, ...);
+```
+
+This option extends [`CREATE INDEX`][postgres-createindex] with the
+ability to use a separate transaction for each chunk it creates an
+index on, instead of using a single transaction for the entire hypertable.
+This allows `INSERT`s, and other operations to to be performed concurrently
+during most of the duration of the `CREATE INDEX`.
+(While the the index is being created on an individual chunk, it functions as
```
svenklemm: the the

api.md
```diff
@@ -412,6 +413,39 @@ queries.
 
 ---
 
+## CREATE INDEX [](create_index)
```
cevian: maybe the heading should be `CREATE INDEX ... WITH (timescaledb.transaction_per_chunk)`?

api.md
```diff
@@ -412,6 +413,39 @@ queries.
...
+
+```SQL
+CREATE INDEX ... WITH (timescaledb.transaction_per_chunk, ...);
+```
+
+This option extends [`CREATE INDEX`][postgres-createindex] with the
+ability to use a separate transaction for each chunk it creates an
+index on, instead of using a single transaction for the entire hypertable.
+This allows `INSERT`s, and other operations to to be performed concurrently
+during most of the duration of the `CREATE INDEX`.
```
solugebefola: "...most of the duration of the `CREATE INDEX` _operation_"?

api.md
solugebefola: I don't think parentheses are needed here

api.md
```diff
@@ -10,6 +10,7 @@
 > - [chunk_relation_size](#chunk_relation_size)
 > - [chunk_relation_size_pretty](#chunk_relation_size_pretty)
 > - [create_hypertable](#create_hypertable)
+> - [CREATE INDEX](#create_index)
```
solugebefola: I think we are using a different syntax for these links. `[create index (transaction per chunk)](#create_index)`. See @gayyappan 's API PR.

solugebefola:
Please check my previous comments as well

api.md
```diff
@@ -10,6 +10,7 @@
 > - [chunk_relation_size](#chunk_relation_size)
 > - [chunk_relation_size_pretty](#chunk_relation_size_pretty)
 > - [create_hypertable](#create_hypertable)
+> - [CREATE INDEX (Transaction Per Chunk)](#create_index)
```
solugebefola: `CREATE INDEX` needs to be lowercase here.

api.md
```diff
@@ -412,6 +413,39 @@ queries.
...
+CREATE INDEX ... WITH (timescaledb.transaction_per_chunk, ...);
+```
+
+This option extends [`CREATE INDEX`][postgres-createindex] with the
+ability to use a separate transaction for each chunk it creates an
+index on, instead of using a single transaction for the entire hypertable.
+This allows `INSERT`s, and other operations to to be performed concurrently
+during most of the duration of the `CREATE INDEX` command.
+While the index is being created on an individual chunk, it functions as
+if a regular `CREATE INDEX` was called on that chunk, however other chunks are
```
solugebefola: "was" => "were"

PR #132, opened 2019-05-01
Document telemetry_metadata

svenklemm:


using-timescaledb/telemetry.md
```diff
@@ -42,6 +45,9 @@ the `UUID`s contain the current timestamp. For full transparency, we expose a
 new API function, [`get_telemetry_report`][get_telemetry_report], that returns
 a text string of the exact JSON that is sent to our servers.
 
+Additionally any content of the table `_timescaledb_catalog.telemetry_metadata`
+and the value of `timescaledb.cloud` will be included in the telemetry report.
```
cevian: should we add timescaledb.cloud to the example data above?

solugebefola: what are the possible values of `timescaledb.cloud` in this example?

svenklemm: timescaledb_telemetry.cloud is text, but i think this should be documented in the list of GUCs (which we dont have yet)

PR #134, opened 2019-05-01
Add docs for drop_chunks with continuous aggs

JLockerman:
Drop chunks has a special flag to define it's behavior when a
continuous aggregate exists on the hypertable. This commit adds
documentation as to what this flag does.

### DIFF
--- a/api.md
+++ b/api.md
@@ -522,6 +522,7 @@ the same semantics as the `show_chunks` [function][show chunks].
 | `table_name` | Hypertable name from which to drop chunks. If not supplied, all hypertables are affected.
 | `schema_name` | Schema name of the hypertable from which to drop chunks. Defaults to `public`.
 | `cascade` | Boolean on whether to `CASCADE` the drop on chunks, therefore removing dependent objects on chunks to be removed. Defaults to `FALSE`.
+| `cascade_to_materializations` | Set to `TRUE` to delete chunk data in associated continuous aggregates. Defaults to `NULL`. `FALSE` is not yet supported.
 
 The `older_than` and `newer_than` parameters can be specified in two ways:
 
@@ -545,6 +546,8 @@ specifying `newer_than => 4 months` and `older_than => 3 months` will drop all f
 all full chunks between '2017-01-01' and '2017-02-01'. Specifying parameters that do not result in an overlapping
 intersection between two ranges will result in an error.
 
+>:TIP: By default, calling `drop_chunks` on a table that has a continuous aggregate will throw an error. This can be resolved by setting `cascade_to_materializations` to `TRUE`, which will cause the corresponding aggregated data to also be dropped.
+
 #### Sample Usage [](drop_chunks-examples)
 
 Drop all chunks older than 3 months ago:
@@ -555,7 +558,7 @@ SELECT drop_chunks(interval '3 months');
 The expected output:
 
 ```sql
- drop_chunks  
+ drop_chunks
 -------------
 
 (1 row)
@@ -594,6 +597,11 @@ SELECT drop_chunks(newer_than => interval '3 months');
 Drop all chunks older than 3 months ago and newer than 4 months ago:
 ```sql
 SELECT drop_chunks(older_than => interval '3 months', newer_than => interval '4 months', table_name => 'conditions')
+```
+
+Drop all chunks older than 3 months, and delete this data from any continuous aggregates based on it:
+```sql
+SELECT drop_chunks(interval '3 months', 'conditions', cascade_to_materializations => true);
 ```
 
 ---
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -522,6 +522,7 @@ the same semantics as the `show_chunks` [function][show chunks].
 | `table_name` | Hypertable name from which to drop chunks. If not supplied, all hypertables are affected.
 | `schema_name` | Schema name of the hypertable from which to drop chunks. Defaults to `public`.
 | `cascade` | Boolean on whether to `CASCADE` the drop on chunks, therefore removing dependent objects on chunks to be removed. Defaults to `FALSE`.
+| `cascade_to_materializations` | Boolean on whether the data in continuous aggregates which are backed by the chunk should be deleted. Defaults to `NULL`. Currently cannot be set to `FALSE`.
```
svenklemm: this sounds a bit complicated, maybe change to `Boolean to control whether deletes on chunks cascade to materializations`

JLockerman: hmm, I think that's a little to "repeats exactly what the argument name is" for me 🤔

svenklemm: also i think we try to avoid `currently` in docs, also why can it not be set to false if its a boolean, shouldnt the default be either true or false and NULL means it's whatever the default is

svenklemm: it just reads weird, i like the style used below more: `(BOOLEAN) Set to true to drop objects dependent upon chunks being dropped. Defaults to false.`
but i'm not native speaker so i'll let others chime in

cevian: Set to `TRUE` to delete chunk data in associate continuous aggregates. Defaults to `NULL`. `FALSE` is not yet supported.

api.md
```diff
@@ -872,6 +880,9 @@ one drop_chunks policy may exist per hypertable.
 |---|---|
 | `cascade` | (BOOLEAN) Set to true to drop objects dependent upon chunks being dropped. Defaults to false.|
 | `if_not_exists` | (BOOLEAN) Set to true to avoid throwing an error if the drop_chunks_policy already exists. A notice is issued instead. Defaults to false. |
+| `cascade_to_materializations` | Boolean on whether the data in continuous aggregates which are backed by the chunk should be deleted. Defaults to `NULL`. Currently cannot be set to `FALSE`. |
```
cevian: also change here as above

api.md
solugebefola: Can you elaborate on what this parameter does?

api.md
```diff
@@ -545,6 +546,8 @@ specifying `newer_than => 4 months` and `older_than => 3 months` will drop all f
 all full chunks between '2017-01-01' and '2017-02-01'. Specifying parameters that do not result in an overlapping
 intersection between two ranges will result in an error.
 
+>:NOTE: By default, calling `drop_chunks` on a table that has a continuous aggregates will throw an error. This can be resolved by setting `cascade_to_materializations` to `true`, which will cause the corresponding aggregated data to also be dropped.
```
solugebefola: "...has a continuous aggregate~s~..."  - remove "s"

solugebefola: By `>:NOTE:`, do you mean `>:TIP:`?

api.md
```diff
@@ -872,6 +880,9 @@ one drop_chunks policy may exist per hypertable.
 |---|---|
 | `cascade` | (BOOLEAN) Set to true to drop objects dependent upon chunks being dropped. Defaults to false.|
 | `if_not_exists` | (BOOLEAN) Set to true to avoid throwing an error if the drop_chunks_policy already exists. A notice is issued instead. Defaults to false. |
+| `cascade_to_materializations` | Set to `TRUE` to delete chunk data in associate continuous aggregates. Defaults to `NULL`. `FALSE` is not yet supported. |
```
solugebefola: Add `(BOOLEAN)` to the start of the description here

api.md
```diff
@@ -545,6 +546,8 @@ specifying `newer_than => 4 months` and `older_than => 3 months` will drop all f
 all full chunks between '2017-01-01' and '2017-02-01'. Specifying parameters that do not result in an overlapping
 intersection between two ranges will result in an error.
 
+>:NOTE: By default, calling `drop_chunks` on a table that has a continuous aggregates will throw an error. This can be resolved by setting `cascade_to_materializations` to `TRUE`, which will cause the corresponding aggregated data to also be dropped.
```
solugebefola: `>:NOTE:` => `>:TIP:`

solugebefola: "aggregates" => "aggregate"

api.md
```diff
@@ -522,6 +522,7 @@ the same semantics as the `show_chunks` [function][show chunks].
 | `table_name` | Hypertable name from which to drop chunks. If not supplied, all hypertables are affected.
 | `schema_name` | Schema name of the hypertable from which to drop chunks. Defaults to `public`.
 | `cascade` | Boolean on whether to `CASCADE` the drop on chunks, therefore removing dependent objects on chunks to be removed. Defaults to `FALSE`.
+| `cascade_to_materializations` | Set to `TRUE` to delete chunk data in associate continuous aggregates. Defaults to `NULL`. `FALSE` is not yet supported.
```
solugebefola: "associate" => "associated"

solugebefola:
small things to change

PR #140, opened 2019-05-14
Dropping a continuous agg requires CASCADE

feikesteenbergen:
As of https://github.com/timescale/timescaledb/commit/7a4191 the dropping
of a continuous agg requires the CASCADE parameter.

### DIFF
--- a/using-timescaledb/continuous-aggregates.md
+++ b/using-timescaledb/continuous-aggregates.md
@@ -161,10 +161,13 @@ Other alterations to the continuous aggregate view are currently disallowed. To
 alter a continuous aggregate view in other ways it must be dropped and
 re-created; this can entail some time to re-calculate aggregations.
 
-A continuous aggregate may be dropped by using the `DROP VIEW` command.
+A continuous aggregate may be dropped by using the `DROP VIEW` command, the
+`CASCADE` parameter is required.
 ```sql
-DROP VIEW device_summary;
+DROP VIEW device_summary CASCADE;
 ```
+>:TIP: `CASCADE` will only drop those objects that depend on the continuous aggregate,
+it does not drop the underlying hypertable or delete any data.
 
 ---
 


mfreed:
Can you clarify (both to me here in github, but possibly in docs changes as well) what the CASCADE means?  I typically would interpret it to mean to "CASCADE" to the underlying raw data, but that doesn't make sense?

(I.e., I know that, as of v1.3, you can't delete raw tables without correspondingly deleting their cagg view, e.g., you'd cascade "up" from the raw to the cagg.  But this seems to suggest the opposite, as your update was about the cagg view not the raw.)

feikesteenbergen:
The `CASCADE` here has a very PostgreSQL specific meaning, so I guess spelling it out makes sense here, would a tip suffice, something of the sort of:

```diff
+>:TIP: `CASCADE` will only drop those objects that depend on the continuous aggregate,
+it does not drop the underlying hypertable or delete any data.
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -161,10 +161,13 @@ Other alterations to the continuous aggregate view are currently disallowed. To
...
 
-A continuous aggregate may be dropped by using the `DROP VIEW` command.
+A continuous aggregate may be dropped by using the `DROP VIEW` command, the
+`CASCADE` parameter is required.
 ```sql
-DROP VIEW device_summary;
+DROP VIEW device_summary CASCADE;
 ```
+>:TIP: `CASCADE` will only drop those objects that depend on the continuous aggregate,
+it does not drop the underlying hypertable or delete any data.
```
mfreed: Just a nit:  It does delete the hypertable/data that comprises the view (here again is some language weirdness, because most VIEWs in PostgreSQL are not materialized, where our caggs are actually materialized views) :)

Maybe something like:

```
it does not drop the underlying hypertable on which the continuous aggregation is based (e.g., the raw data) or delete any data from this hypertable.```

mfreed: Missed that already merged.  Will submit as separate PR.

solugebefola:
please label with earliest db version these changes apply to

PR #141, opened 2019-05-14
Describe time_partitioning_func feature

feikesteenbergen:
The time_partition_func was introduced with the following commit:
https://github.com/timescale/timescaledb/pull/639/files

However the argument was not yet added to the api documentation.

I'm a bit unsure about the example, as it is quite a big example, but I think it may be needed to clarify how to use this functionality.

### DIFF
--- a/api.md
+++ b/api.md
@@ -217,6 +217,7 @@ still work on the resulting hypertable.
 | `associated_schema_name` | Name of the schema for internal hypertable tables. Default is "_timescaledb_internal". |
 | `associated_table_prefix` | Prefix for internal hypertable chunk names. Default is "_hyper". |
 | `migrate_data` | Set to `true` to migrate any existing `main_table` data to chunks in the new hypertable. A non-empty table will generate an error without this option. Note that, for large tables, the migration might take a long time. Defaults to false. |
+| `time_partitioning_func` | Function to convert incompatible primary time column values to compatible ones. The function must be `IMMUTABLE`. |
 | `chunk_target_size` | DEPRECATED - The target size of a chunk (including indexes) in `kB`, `MB`, `GB`, or `TB`. Setting this to `estimate` or a non-zero chunk size, e.g., `2GB` will enable adaptive chunking (a DEPRECATED feature). The `estimate` setting will estimate a target chunk size based on system information. Adaptive chunking is disabled by default. |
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -333,6 +334,20 @@ if `conditions` is already a hypertable:
 SELECT create_hypertable('conditions', 'time', if_not_exists => TRUE);
 ```
 
+Convert table `measurements`, containing a composite type `report` which contains a field named `created` to hypertable with just time partitioning.
```
solugebefola: Maybe don't use `created` as an example field?  It might confuse the issue with the actual `create_hypertable` command.

api.md
```diff
@@ -217,6 +217,7 @@ still work on the resulting hypertable.
 | `associated_schema_name` | Name of the schema for internal hypertable tables. Default is "_timescaledb_internal". |
 | `associated_table_prefix` | Prefix for internal hypertable chunk names. Default is "_hyper". |
 | `migrate_data` | Set to `true` to migrate any existing `main_table` data to chunks in the new hypertable. A non-empty table will generate an error without this option. Note that, for large tables, the migration might take a long time. Defaults to false. |
+| `time_partitioning_func` | The function to convert a `time_column_name` into a compatible value. The function must be `IMMUTABLE`. |
```
solugebefola: More explicit?
> Function to convert incompatible primary time column values to compatible ones.  The function must be `IMMUTABLE`.

mfreed:
I *think* @erimatnor wrote this code.  Adding him as reviewer.

api.md
erimatnor: I would say something like:

```
Time partition table `meausurements` on a composite column type `report` using a time partitioning function:
```

feikesteenbergen:
> LGTM. Made one suggestion for a tweak though.
> 
> Maybe we should also add a small blurb describing this functionality and why/when it is useful?

I guess that would be useful, I can think of a few use cases that we could address. Where should we document that? I don't want to bloat the api docs too much with elaborate examples, so would this blurb be part of *Using TimescaleDB*?

erimatnor:
@feikesteenbergen Under "Using TimescaleDB" there's a Tip that says:

```
The 'time' column used in the create_hypertable function supports timestamp, date, or integer types, so you can use a parameter that is not explicitly time-based, as long as it can increment. For example, a monotonically increasing id would work.```

I guess we could expand that to quickly mention support for other types using a time partitioning function

solugebefola:
please label your PR with the earliest db version the changes apply to

mfreed:
(Given accepts, please merge without my review ;)

PR #171, opened 2019-07-04
Reference documentation for add_data_node and delete_data_node

mkindahl:


svenklemm:
Looks like this PR is against wrong branch

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
 SELECT add_dimension('conditions', 'device_id', number_partitions => 2, if_not_exists => true);
 ```
 
+---
+## add_data_node() [](add_data_node)
```
mfreed: For this type of distributed hypertable / distributed TimescaleDB management, I think we need to create some canonical instructions about where they are executing, and possibly even capture this with the arguments.  For example, I *think* this is executed on the access node which is handling DDL for the distributed hypertable, but this should be clarified?

Also, I find it gets a little confusing between "database" and "bootstrap_database".  Both seem to refer to the name on the remote data node where the the distributed hypertable will be instantiated.

Similarly, to nit pick, are we _creating_ a "remote" hypertable, or is this about expanding an existing distributed hypertable?

Finally, I don't fully understand IP address / hostname management from this function, given that hostname appears to be _optional_ from these instructions.  What happens if the hostname isn't provided?  Minor q:  Is there support for submitting IP addresses vs. hostnames (or is it just passing in IPv4/IPv6 as a string to host argument?)

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+
+| Column          | Description                                   |
+|-----------------|-----------------------------------------------|
+| `node_name`     | Name of the data node                         |
+| `host`          | Host name for the remote data node            |
+| `port`          | Port for the remote data node                 |
+| `database`      | Name of database used on the remote data node |
+| `username`      | Local username                                |
+| `node_username` | Remote username                               |
+| `created`       |                                               |
```
mfreed: What is type of created?  It this a boolean or time?

api.md
mfreed: Now reading the same usage below, seems like this is meant to be used _below_ we actually create the distributed hypertable.  I think this could be better clarified in the preface to this API call?

Similarly, please clarify is this is _also_ used for elastically expanding an existing hypertable (i.e., you can in the middle of existing hypertable, add_data_node() then expand hypertable to that data node), or if that's a different API call.

erimatnor:
@mkindahl As Sven points out, this PR should be against the `clustering` branch. I'll update that.

api.md
```diff
@@ -171,7 +172,115 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2, if_not_e
 ```
 
 ---
-## attach_tablespace() [](attach_tablespace)
+### add_data_node() [](add_data_node)
+
+Create a new data node for use when creating distributed hypertables.
```
erimatnor: I would avoid the "create" terminology here since people might get the wrong idea of what this function does. I would say something like:

"Add a new data node to the current database for use with distributed hypertables. This includes bootstrapping the data node by creating and configuring a new database on it. Newly created distributed hypertables will be able to make use of this data node for distributing data across the cluster. Note, however, that already existing distributed hypertables may require a separate `attach_data_node()` call to schedule the data node for use with those tables as they do not automatically expand to include the new data node."

api.md
```diff
@@ -171,7 +172,115 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2, if_not_e
...
+
+To create a distributed hypertable it is necessary to define the data
+nodes where the data is going to be stored, which this utility is used
```
erimatnor: utility -> function

api.md
```diff
@@ -171,7 +172,115 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2, if_not_e
...
+for.
+
+The utility will attempt to bootstrap the data node by creating the
```
erimatnor: I would elide this part if you go with my suggestion above.

api.md
```diff
@@ -171,7 +172,115 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2, if_not_e
...
+database where the remote hypertables will later be created and also
+load the TimescaleDB extension on the data node.
+
+#### Errors
+
+* An error will be given if an attempt is made to run the function
+  inside a transaction.
+* An error will be given if the database already exists on the data
```
erimatnor: remote database

api.md
```diff
@@ -171,7 +172,115 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2, if_not_e
...
+  node and `if_not_exists` is not set.
+
+#### Required Arguments [](add_data_node-required-arguments)
+
+| Name        | Description       |
+| ----------- | -----------       |
+| `node_name` | Name of the node. |
```
erimatnor: Name of the data node.

api.md
```diff
@@ -171,7 +172,115 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2, if_not_e
...
+
+##### Parameter Description
+
+* `node_name` - Name to use for the node locally.
+
+#### Optional Arguments [](add_data_node-optional-arguments)
+
+| Name                 | Description                                           |
+|----------------------|-------------------------------------------------------|
+| `host`               | Host name for the remote data node.                   |
```
erimatnor: Should we put the default setting in this table instead of below? Or, is this the established pattern?

api.md
erimatnor: `created` is  a boolean that indicates whether the remote database was created or not (needed because of `if_not_exists`parameter).

api.md
```diff
@@ -171,7 +172,115 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2, if_not_e
...
+`dn1.example.com` and `dn2.example.com`.
+
+To do this, you can just type:
+
+```sql
+SELECT add_data_node('dn1', host => 'dn1.example.com');
+SELECT add_data_node('dn2', host => 'dn2.example.com');
+SELECT create_distributed_hypertable(
+    'conditions', 'time', 'location',
+	servers => '{"dn1","dn2"}'
```
erimatnor: You may elide the server list here and it will make use of all servers in the database. I think this should probably be the default documented behavior. We can go into more details about servers  (should probably be `data_nodes`) in the description of `create_distributed_hypertable()`.

getting-started/installation-timescale-cloud.md
```diff
@@ -0,0 +1,25 @@
+## Installation (Timescale Cloud) [](installation-timescale-cloud)
```
erimatnor: Why is this stuff in this PR?

api.md
mkindahl: Yup.

api.md
mkindahl: The tables become exceedingly wide and the default is occasionally derived from other parameters, hence require more elaborate descriptions, which is why I went with this approach.

api.md

getting-started/installation-timescale-cloud.md
mkindahl: No idea. Will check.

mkindahl: Couldn't see it in the diff I just pushed. Will check the newly pushed one.

api.md

api.md

api.md

api.md
mkindahl: This is executed on the access node and yes, there are no general instructions on how distributed hypertables work, which would help in making the role of this particular function clear.

I agree that all the users, passwords, and hosts is a little confusing. This seems to be an intermediate situation and will be resolved as we go forward. IMHO, we should still document it in it's current state though.

The function `attach_data_node()` is used to elastically scale an existing hypertable, as clarified by @erimatnor below. Followed his suggestion and updated the text.

mkindahl:
> @mkindahl As Sven points out, this PR should be against the `clustering` branch. I'll update that.

Fixed that locally as well.

api.md
```diff
@@ -170,6 +171,118 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+cluster.
+
+Note, however, that already existing distributed hypertables may
+require a separate `attach_data_node()` call to schedule the data node
+for use with those tables as they do not automatically expand to
+include the new data node.
+
+The function will attempt to bootstrap the data node by creating the
+database where the remote hypertables will later be created and also
+load the TimescaleDB extension on the data node.
```
dianasaur323: Ah, I like this explanation of bootstrapping here, so definitely think we can remove the bootstrapping reference above

api.md
```diff
@@ -170,6 +171,118 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
 SELECT add_dimension('conditions', 'device_id', number_partitions => 2, if_not_exists => true);
 ```
 
+---
+## add_data_node() [](add_data_node)
+
+Add a new data node to the current database for use with distributed
```
dianasaur323: Could we add a sentence here that says something along the lines of - this command must be run on an access node?

api.md
```diff
@@ -170,6 +171,118 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+hypertables.
+
+This includes bootstrapping the data node by creating and configuring
+a new database on it. Newly created distributed hypertables will be
```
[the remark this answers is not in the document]
mkindahl: Since bootstrapping is done, I think we need to describe it somewhere. The text here, however, might be unnecessarily terse.

api.md

api.md

api.md
```diff
@@ -170,6 +171,118 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+able to make use of this data node for distributing data across the
+cluster.
+
+Note, however, that already existing distributed hypertables may
```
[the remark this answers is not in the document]
mkindahl: Yeah, I have a tendency to add "however" too often. =)

Reformulated the sentence as well. PTAL.

api.md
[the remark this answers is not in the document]
mkindahl: I added a brief description about this below, so resolving this comment. Please tell me if you think I should change the text.

api.md
erimatnor: @mkindahl Note PR here: https://github.com/timescale/timescaledb-private/pull/193

It refactors and simplifies this function. Also, it generally hardens the server management functions with proper permissions checks. 

You need USAGE on servers to add servers and owner on a hypertable to be able to attach servers to it. Might also be worth mentioning.

dianasaur323:
oh @mkindahl do you mind flagging this function as "community" before you merge? 
image

mkindahl:
> oh @mkindahl do you mind flagging this function as "community" before you merge?
> image

Did that.

api.md
mkindahl: @erimatnor Yeah, but I think we can merge this PR in the current state and fix things as we go, adding a documentation task after the PR above is merged.

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
 SELECT add_dimension('conditions', 'device_id', number_partitions => 2, if_not_exists => true);
 ```
 
+---
+## add_data_node() :community_function: [](add_data_node)
+
+WHen executed on the access node, this will add a new data node to the
```
k-rus: Am I correct that adding the first data node makes the current node to become an access node? Can it be clarified what it means to call it on the node the first time?

k-rus: This might be good place to give idea to users what an access node is, or at a function description where it is created.

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+* `node_name` - Name to use for the node locally.
+
+#### Optional Arguments [](add_data_node-optional-arguments)
+
+| Name                 | Description                                           |
+|----------------------|-------------------------------------------------------|
+| `host`               | Host name for the remote data node.                   |
+| `port`               | Port to use on the remote data node.                  |
+| `database`           | Database name where the remote hypertable is created. |
+| `local_user`         | User name on the local node.                          |
```
k-rus: Should it be:
> User name on the local access node.
I.e., similar to the description of `remote_user`?

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+| `remote_user`        | User name on the remote data node.                    |
+| `password`           | Password on the remote data node.                     |
+| `if_not_exists`      | Do not fail if the hyper table already exists         |
```
k-rus: As it described above the functionality doesn't affect existing hyper tables. Is it about existing database? Or do I miss anything?

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+
+* `local_user` - Name of the local user. Default is to use the current
+  user.
+
+* `remote_user` - Name of the user on the remote data node. Default is
+  to use the same user as `local_user`.
+
+* `password` - Password to use on the remote data node. Defaults is to
+  not use a password.
+
```
k-rus: Is the description of `if_not_exists` missing.

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+* `bootstrap_database` - The database to use when bootstrapping the
+  server as described above. The default is to use `'postgres'`.
```
k-rus: Is it possible to provide hyperlink to the description above?

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+| `password`           | Password on the remote data node.                     |
+| `if_not_exists`      | Do not fail if the hyper table already exists         |
+| `bootstrap_database` | Database to bootstrap.                                |
+| `bootstrap_user`     | User to used for bootstrapping.                       |
+| `bootstrap_password` | Password for user use when bootstrapping.             |
+
+##### Parameter Description
+
+* `host` - the name of the host to use for the remote data node. This
+  defaults to `'localhost'`.
```
k-rus: Is it possible to use IP instead of the host name?

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+
+| Column          | Description                                   |
+|-----------------|-----------------------------------------------|
+| `node_name`     | Name of the data node                         |
+| `host`          | Host name for the remote data node            |
+| `port`          | Port for the remote data node                 |
+| `database`      | Name of database used on the remote data node |
+| `username`      | Local username                                |
+| `node_username` | Remote username                               |
+| `created`       | True if a new table was created               |
```
k-rus: There is nothing above in the description of the function indicates that a new table will be created. Isn't it a database created?

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+
+#### Sample usage [](add_data_node-examples)
+
+Let's assume that you have an existing table `conditions` and want to
+use the `time` column as the time partitioning column and the
+`location` column as the space partitioning column. You also want want
```
k-rus: Duplication of `want`

api.md
k-rus: Double capital in `WHen`

mkindahl: @k-rus There is a need for a separate description of what access nodes and data nodes are, but I think this is beyond the scope of this (reference) documentation.

IMHO, nodes as such as not dedicated data or access nodes, so it does not "become" an access node. For a distributed hypertable, there is one access node where the table is accessed as a whole, and data nodes where the pieces of the table is, so a node can be both an access node and a data node depending on how the hypertable was created.

k-rus: I have the same understanding. Thus the phrase

> When executed on the access node

is confusing for me. Can it be better something like:

> Add a data node into the current database to be used with distributed hypertables through this access node.

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+  to use the same user as `local_user`.
+
+* `password` - Password to use on the remote data node. Defaults is to
+  not use a password.
+
+* `bootstrap_database` - The database to use when bootstrapping the
+  server as described above. The default is to use `'postgres'`.
+
+* `bootstrap_user` - User to use when bootstrapping the server as
+  described above. The default is to use the saver provided as
```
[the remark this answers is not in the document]
mfreed: To be honest, I don't think I clearly understand from this writeup the difference between the "bootstrap_user" and the "remote_user"?

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+
+* `bootstrap_password` - Password to use when bootstrapping the server
+  as described above. The default is to use the password as provided
+  as `password`.
+
+#### Returns [](add_data_node-returns)
+
+| Column          | Description                                   |
+|-----------------|-----------------------------------------------|
+| `node_name`     | Name of the data node                         |
```
mfreed: remote data node?

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+| `host`          | Host name for the remote data node            |
+| `port`          | Port for the remote data node                 |
+| `database`      | Name of database used on the remote data node |
+| `username`      | Local username                                |
+| `node_username` | Remote username                               |
+| `created`       | True if a new table was created               |
+
+#### Sample usage [](add_data_node-examples)
+
+Let's assume that you have an existing table `conditions` and want to
```
mfreed: Assume that this should be "hypertable" `conditions"?   And below "chunks of the hypertable"?

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
 SELECT add_dimension('conditions', 'device_id', number_partitions => 2, if_not_exists => true);
 ```
 
+---
+## add_data_node() :community_function: [](add_data_node)
```
mfreed: I do not believe licensing has been finalized.  Would like to pause merging (although can fix below) until finalized, or possibly just remove the decorator until we do (will get back to you today about what to do here).  Sorry for mixed messages. :)

api.md

k-rus: @mkindahl @erimatnor Can you explain which hypertable is meant here?

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+| `bootstrap_database` | Database to bootstrap.                                |
+| `bootstrap_user`     | User to used for bootstrapping.                       |
+| `bootstrap_password` | Password for user use when bootstrapping.             |
+
+##### Parameter Description
+
+* `host` - the name of the host to use for the remote data node. This
+  defaults to `'localhost'`.
+
+* `post` - Port to use for the remote data node. This defaults to the
```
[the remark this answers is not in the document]
k-rus: @mfreed Good point. I feel the call with defaults should always fail. The default host is the local host and the default database is the same database name as current, thus it will try to add the same node/database to itself, which will fail if `if_not_exists` is not set.

api.md
mkindahl: Added.

api.md
mkindahl: Since `getaddrinfo(3)` is used, I would say this is implicit?

api.md
mkindahl: Yes, my fault. Fixed.

api.md

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+  node. Default is to use the current database in use when calling the
+  function.
+
+* `local_user` - Name of the local user. Default is to use the current
+  user.
+
+* `remote_user` - Name of the user on the remote data node. Default is
+  to use the same user as `local_user`.
+
+* `password` - Password to use on the remote data node. Defaults is to
```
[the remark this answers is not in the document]

api.md
```diff
@@ -170,6 +171,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+| `host`               | Host name for the remote data node.                   |
+| `port`               | Port to use on the remote data node.                  |
+| `database`           | Database name where the remote hypertable is created. |
+| `local_user`         | User name on the local node.                          |
+| `remote_user`        | User name on the remote data node.                    |
+| `password`           | Password on the remote data node.                     |
+| `if_not_exists`      | Do not fail if the hyper table already exists         |
+| `bootstrap_database` | Database to bootstrap.                                |
+| `bootstrap_user`     | User to used for bootstrapping.                       |
+| `bootstrap_password` | Password for user use when bootstrapping.             |
```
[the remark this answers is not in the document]

api.md
mkindahl: Well... hmm... here are some confusions. The name is used locally to refer to a remote data node. On another access node, the same data node might have a different name. Changing to "local name to use for data node".

api.md
mkindahl: Ok. Sitting on the commit for the time being then.

api.md
k-rus: It looks like it is common for Postgres documentation to say just host name, meaning both name and IP. So I guess it is obvious for the users.

api.md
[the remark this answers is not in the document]
mfreed: I guess I should have been clearer:  From this discussion, my reaction is that `host` should be *required* not optional.

api.md
mkindahl: Reformulated.

api.md
mkindahl: Does not seem to be easy to add labels arbitrarily.

api.md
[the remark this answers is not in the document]
mkindahl: I think it will be confusing, yes. Since data nodes are really only sensible if elsewhere, we can have the host and port as required arguments. If a user really want to have things locally, she can always provide the local address herself.

api.md
```diff
@@ -170,6 +172,118 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+* `node_name` - Name to use for the node locally.
+
+#### Optional Arguments [](add_data_node-optional-arguments)
+
+| Name                 | Description                                           |
+|----------------------|-------------------------------------------------------|
+| `host`               | Host name for the remote data node.                   |
+| `port`               | Port to use on the remote data node.                  |
+| `database`           | Database name where the remote hypertable is created. |
+| `local_user`         | User name on the local node.                          |
```
erimatnor: Note the now updated interface that simplifies this interface (removes some args)

api.md
```diff
@@ -170,6 +172,118 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+| `remote_user`        | User name on the remote data node.                    |
+| `password`           | Password on the remote data node.                     |
+| `if_not_exists`      | Do not fail if the data node already exists.          |
+| `bootstrap_database` | Database to bootstrap.                                |
```
erimatnor: This is not the database to bootstrap but the database to connect to in order to bootstrap (create) another database (given by the `database` argument.)

api.md
```diff
@@ -170,6 +172,118 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+| `bootstrap_user`     | User to be used for bootstrapping.                    |
```
erimatnor: This user must have CREATEDB permissions on the data node and AFAIK be a  superuser to create the extension in the new db

api.md
```diff
@@ -170,6 +172,118 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+| `bootstrap_password` | Password to be used when bootstrapping.               |
+
+##### Parameter Description
+
+* `host` - the name of the host to use for the remote data node. This
+  defaults to `'localhost'`.
+
+* `port` - Port to use for the remote data node. This defaults to the
```
erimatnor: I think it actually defaults to the port used by the access node, which is usually the default postgres port

api.md
```diff
@@ -170,6 +172,118 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+  default PostgreSQL port.
+
+* `database` - Name of the database to use on the remote data
```
erimatnor: current database name...

api.md
```diff
@@ -170,6 +172,118 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+  node. Default is to use the current database in use when calling the
+  function.
+
+* `local_user` - Name of the local user. Default is to use the current
```
erimatnor: This parameter is no longer present

api.md
```diff
@@ -170,6 +172,118 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+  user.
+
+* `remote_user` - Name of the user on the remote data node. Default is
```
erimatnor: This parameter is no longer present

api.md
```diff
@@ -170,6 +172,118 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+* `bootstrap_password` - Password to be used when bootstrapping the
+  data node as described above. The default is to use the password as
+  provided as `password`.
+
+* `if_not_exists` - Do not fail the operation if the data node already
+  exists. The data node will not be overwritten, but the command will
+  not generate an error and instead generate a notice.
+
+#### Returns [](add_data_node-returns)
+
```
erimatnor: These return columns have been changed

api.md
```diff
@@ -468,6 +582,60 @@ CREATE INDEX ON conditions(time, location) USING brin
...
+##### Parameter Description
+
+* `if_exists` - Do not generate an error if the data node does not
+  exist. Defaults to false.
+
+* `cascade` - Cascade the delete so that all objects dependent on the
+  data node are deleted as well. Defaults to false.
+
+* `force` - Force removal of data nodes from hypertables, potentially
```
erimatnor: Data loss should not be possible here. AFAIK, this is only to force removal even if chunks in some hypertables will be under-replicated (this is the same as for detach, since delete means implicit detaches on all hypertables). We never allow removal in case of data loss as that would leave hypertables in an undefined state (basically broken).

api.md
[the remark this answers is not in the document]
mfreed: I think with port as optional makes sense given default to postgres, was more commenting about host field.

api.md
mkindahl: Changed this.

api.md
mkindahl: From what I understand, the decision was that the decorator should not be there, so removing it.

api.md
mkindahl: That is *really* confusing... :-/

api.md
mkindahl: Not sure what you want changed here.

api.md
mkindahl: Not sure what change you want here.

api.md
mkindahl: Removed.

api.md
mkindahl: Removed.

api.md
mkindahl: Ok. Rechecking `ddl_api.sql`

api.md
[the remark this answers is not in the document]
mkindahl: `remote_user` was removed in the latest iteration. Will update documentation.

api.md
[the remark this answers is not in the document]
mkindahl: Right.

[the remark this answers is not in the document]
mkindahl: Will handle this as a separate PR.

api.md

api.md
mkindahl: Checked the code. You're right, it should only force the operation when there is more than 1 replicated node. Fixed the text.

api.md
mfreed: I'm thinking he's suggesting that rather than defauling to default PostgreSQL port, it's:  ```* `port` - Port to use for the remote data node. This defaults to the
  port used by the access node```
or 
```by the access node on which the function is executed```

File: api.md [hunk 39]
```diff
@@ -170,6 +172,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+* An error will be given if the remote database already exists on the
+  data node and `if_not_exists` is not set.
+
+#### Required Arguments [](add_data_node-required-arguments)
+
+| Name        | Description                         |
+| ----------- | -----------                         |
+| `node_name` | Name of the data node.              |
+
+##### Parameter Description
```
solugebefola: This section is not neccesary

api.md
```diff
@@ -170,6 +172,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+| `host`               | Host name for the remote data node.                   |
+| `port`               | Port to use on the remote data node.                  |
+| `database`           | Database name where the remote hypertable is created. |
+| `password`           | Password on the remote data node.                     |
+| `if_not_exists`      | Do not fail if the data node already exists.          |
+| `bootstrap_database` | Database to use when bootstrapping.                   |
+| `bootstrap_user`     | User to be used for bootstrapping.                    |
+| `bootstrap_password` | Password to be used when bootstrapping.               |
+
+##### Parameter Description
```
solugebefola: These descriptions can be incorporated into the table

api.md
```diff
@@ -170,6 +172,114 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+* `password` - Password to be used on the remote data node. The
+  default is to not use a password.
+
+* `bootstrap_database` - The database to use when bootstrapping the
+  data node as described above.
+  
+  The default is `'postgres'`.
+
+* `bootstrap_user` - User to be used when bootstrapping the data node
+  as described above. The user need to have `CREATEDB` permissions on
```
```suggestion
  as described above. The user needs to have `CREATEDB` permissions on
```

api.md
```diff
@@ -468,6 +578,68 @@ CREATE INDEX ON conditions(time, location) USING brin
...
 ```
 
+## delete_data_node() [](delete_data_node)
+
+This function will remove the data node locally. This will *not*
+affect the remote database in any way, it will just update the local
+index over all existing data nodes.
+
+The data node will be detached from all hypertables that are using
+it. For more information, see [`detach_data_node`].
```
```suggestion
it. For more information, see [`detach_data_node`](#detach_data_node).
```

api.md
```diff
@@ -468,6 +578,68 @@ CREATE INDEX ON conditions(time, location) USING brin
...
+
+#### Errors
+
+* An error will be generated if the data node cannot be detached from
```
```suggestion
An error will be generated if the data node cannot be detached from
```

api.md
```diff
@@ -468,6 +578,68 @@ CREATE INDEX ON conditions(time, location) USING brin
...
+  some hypertable.
```
```suggestion
all attached hypertables.
```

api.md
```diff
@@ -468,6 +578,68 @@ CREATE INDEX ON conditions(time, location) USING brin
...
+
+##### Parameter Description
+
+* `node_name` - Name to use for the node locally.
+
+#### Optional Arguments [](delete_data_node-optional-arguments)
+
+| Name        | Description                                           |
+|-------------|-------------------------------------------------------|
+| `if_exists` | Don't generate error if the data node does not exist. |
```
```suggestion
| `if_exists` | Prevent error if the data node does not exist. |
```

api.md
WireBaron: Went with "The default is the  PostgreSQL port used by the access node on which the function is  executed."

api.md
WireBaron: I assume you were referring to the default value.  It's been fixed.

api.md
WireBaron: fixed

api.md
```diff
@@ -170,6 +172,77 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+## add_data_node() [](add_data_node)
+
+Add a new data node to the database to be used for creating
+distributed hypertables.
+
+Newly created distributed hypertables will be able to make use of this
+data node for distributing data across the cluster. Note that existing
+distributed hypertables will not automatically use the newly added
+data node. For existing distributed hypertables to use added data
+nodes, use [`attach_data_node()`](#attach_data_node).
```
```suggestion
nodes, use [`attach_data_node`](#attach_data_node).
```

api.md
```diff
@@ -170,6 +172,77 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+
+The function will attempt to bootstrap the data node by:
+1. Connecting to a specified postgres instance
+2. Creating the database that will serve as the new data node
+3. Loading the TimescaleDB extension on the new database
+
+#### Errors
+
+An error will be given if:
+* an attempt is made to run the function inside a transaction.
```
```suggestion
* an attempt is made to run the function inside a transaction _or_
```

api.md
```diff
@@ -170,6 +172,77 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+| `port`              | Port for the remote data node                     |
+| `database`          | Name of database used on the remote data node     |
+| `node_created`      | Was the node created                              |
+| `database_created`  | Was the database created on the remote data node  |
+| `extension_created` | Was the extension created on the remote data node |
+
+#### Sample usage [](add_data_node-examples)
+
+Let's assume that you have an existing hypertable `conditions` and
+want to use the `time` column as the time partitioning column and the
```
```suggestion
want to use `time` as the time partitioning column and
```

api.md
```diff
@@ -170,6 +172,77 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+`location` column as the space partitioning column. You also want to
```
```suggestion
`location` as the space partitioning column. You also want to
```

api.md
```diff
@@ -170,6 +172,77 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+distribute the chunks of the hypertable on two data nodes
+`dn1.example.com` and `dn2.example.com`.
+
+To do this, you can just type:
```
```suggestion

```

api.md
```diff
@@ -170,6 +172,77 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+| `database_created`  | Was the database created on the remote data node  |
+| `extension_created` | Was the extension created on the remote data node |
+
+#### Sample usage [](add_data_node-examples)
+
+Let's assume that you have an existing hypertable `conditions` and
+want to use the `time` column as the time partitioning column and the
+`location` column as the space partitioning column. You also want to
+distribute the chunks of the hypertable on two data nodes
+`dn1.example.com` and `dn2.example.com`.
```
```suggestion
`dn1.example.com` and `dn2.example.com`:
```

api.md
WireBaron: done

api.md
```diff
@@ -170,6 +172,75 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+| `node_name` | Name for the data node.             |
+
+#### Optional Arguments [](add_data_node-optional-arguments)
+
+| Name                 | Description                                           |
+|----------------------|-------------------------------------------------------|
+| `host`               | Host name for the remote data node. The default is `'localhost'`. |
+| `port`               | Port to use on the remote data node. The default is the PostgreSQL port used by the access node on which the function is executed. |
+| `database`           | Database name where remote hypertables will be created. The default is the current database name. |
+| `password`           | Password to be used on the remote data node. The default is to not use a password. |
```
k-rus: Will it create new user in the remote database with the name of the current local user and the given remote password? Is it good idea to clarify this? Or is it obvious for PostgreSQL users?

erimatnor: Ideally, it would create the user, but it currently doesn't because we haven't implemented any of the roles/permissions management. So, for now I think we can just add a note that the user is assumed to exist on data nodes and that it is a manual process right now.

api.md
```diff
@@ -468,6 +539,45 @@ CREATE INDEX ON conditions(time, location) USING brin
...
 ```
 
+## delete_data_node() [](delete_data_node)
+
+This function will remove the data node locally. This will *not*
+affect the remote database in any way, it will just update the local
+index over all existing data nodes.
+
+The data node will be detached from all hypertables that are using
+it. For more information, see [`detach_data_node`](#detach_data_node).
```
```suggestion
it if permissions and data integrity requirements are satisfied. For more information, see [`detach_data_node`](#detach_data_node).
```

api.md
```diff
@@ -170,6 +172,75 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
 ```
 
+---
+## add_data_node() [](add_data_node)
+
+Add a new data node to the database to be used for creating
+distributed hypertables.
+
+Newly created distributed hypertables will be able to make use of this
+data node for distributing data across the cluster. Note that existing
```
k-rus:
```suggestion
data node for distributing data across the distributed database. Note that existing
```
As mentioned in another PR it is better to avoid `cluster` due to confusion with PostgreSQL terminology.

api.md
```diff
@@ -170,6 +172,75 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+| `database_created`  | Was the database created on the remote data node  |
+| `extension_created` | Was the extension created on the remote data node |
+
+#### Sample usage [](add_data_node-examples)
+
+Let's assume that you have an existing hypertable `conditions` and
+want to use `time` as the time partitioning column and `location` as
+the space partitioning column. You also want to distribute the chunks
+of the hypertable on two data nodes `dn1.example.com` and
+`dn2.example.com`:
```
k-rus: Isn't it good to mention that the current user should be a superuser to succeed? Otherwise I am confused about superuser requirement for bootstrap user.

api.md
WireBaron: Added a note here "Password to be used when connecting to the remote data node. Note that the user name will be the same as the current user. The default is to not use a password."  There probably is still some room for confusion here, but hopefully this is better.

api.md
WireBaron: changed.

api.md
WireBaron: Changed this to "apportioning data across the distributed database" to make it less repetitive.

api.md
WireBaron: Yeah, I added a note and provided a more fleshed out example.

api.md
```diff
@@ -170,6 +172,85 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+
+#### Returns [](add_data_node-returns)
+
+| Column              | Description                                       |
+|---------------------|---------------------------------------------------|
+| `node_name`         | Local name to use for the data node               |
+| `host`              | Host name for the remote data node                |
+| `port`              | Port for the remote data node                     |
+| `database`          | Name of database used on the remote data node     |
+| `node_created`      | Was the node created                              |
```
```suggestion
| `node_created`      | Was the data node created locally                              |
```

WireBaron: changed

api.md
```diff
@@ -170,6 +172,85 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+| `bootstrap_password` | Password to be used when bootstrapping. The default is the password provided in `password`. |
+
+#### Returns [](add_data_node-returns)
+
+| Column              | Description                                       |
+|---------------------|---------------------------------------------------|
+| `node_name`         | Local name to use for the data node               |
+| `host`              | Host name for the remote data node                |
+| `port`              | Port for the remote data node                     |
+| `database`          | Name of database used on the remote data node     |
```
k-rus:
```suggestion
| `database`          | Database name used on the remote data node     |
```
Will look more consistent with the above. Otherwise it is good to use article. But I am not so good in English.

WireBaron: changed

api.md
```diff
@@ -170,6 +172,85 @@ SELECT add_dimension('conditions', 'device_id', number_partitions => 2);
...
+```
+
+Note that the previous example will only succeed if the current user
+has `SUPERUSER` permission on the two remote systems and doesn't
+require a password to connect.  If the current user doesn't have the
+appropriate permissions, you'll have to provide a bootstrap user as
+described above:
+
+```sql
+SELECT add_data_node('dn1', host => 'dn1.example.com', password => 'mydn1password', bootstrap_user => 'dn1_superuser', bootstrap_password => 'dn1_su_password');
```
k-rus: May be split into several lines?

api.md
```diff
@@ -468,6 +549,46 @@ CREATE INDEX ON conditions(time, location) USING brin
...
+| Name        | Description            |
+| ----------- | -----------            |
+| `node_name` | Name of the data node. |
+
+#### Optional Arguments [](delete_data_node-optional-arguments)
+
+| Name        | Description                                           |
+|-------------|-------------------------------------------------------|
+| `if_exists` | Prevent error if the data node does not exist. Defaults to false. |
+| `cascade`   | Cascade the delete so that all objects dependent on the data node are deleted as well. Defaults to false. |
```
k-rus: I have no idea what the dependent objects. Will it be important for users to know?

PR #173, opened 2019-07-08
Update drop_chunks return behavior

spolcyn:
Previously, drop_chunks returned an empty table. As of TimescaleDB PR 1325, it returns the internal descriptors of the chunks that it dropped.

### DIFF
--- a/api.md
+++ b/api.md
@@ -554,6 +554,8 @@ SELECT detach_tablespaces('conditions');
 
 Removes data chunks whose time range falls completely before (or after) a
 specified time, operating either across all hypertables or for a specific one.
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -613,13 +615,17 @@ Drop all chunks older than 3 months ago:
 SELECT drop_chunks(interval '3 months');
 ```
 
-The expected output:
+An example output:
```
svenklemm: I would drop the `An` here and just name it `Example output:`

svenklemm:
Dont forget to label with version you want this to go in

api.md
solugebefola: I agree

PR #183, opened 2019-07-11
Document `create_distributed_hypertable`

WireBaron:
This change adds the `create_distributed_hypertable` command to the
api.md file.  It also adds two new optional parameters to the
`create_hypertable` command.

### DIFF
--- a/api.md
+++ b/api.md
@@ -10,6 +10,7 @@
 > - [attach_tablespace](#attach_tablespace)
 > - [chunk_relation_size](#chunk_relation_size)
 > - [chunk_relation_size_pretty](#chunk_relation_size_pretty)
+> - [create_distributed_hypertable](#create_distributed_hypertable)
 > - [create_hypertable](#create_hypertable)
 > - [create index (transaction per chunk)](#create_index)
 > - [create view (continuous aggregate)](#continuous_aggregate-create_view)
@@ -255,6 +256,8 @@ still work on the resulting hypertable.
 | `associated_table_prefix` | Prefix for internal hypertable chunk names. Default is "_hyper". |
 | `migrate_data` | Set to `true` to migrate any existing `main_table` data to chunks in the new hypertable. A non-empty table will generate an error without this option. Note that, for large tables, the migration might take a long time. Defaults to false. |
 | `time_partitioning_func` | Function to convert incompatible primary time column values to compatible ones. The function must be `IMMUTABLE`. |
+| `replication_factor` | If set to 1 or greater, will create a distributed hypertable. Values greater than 1 are currently not recommended. The default value is NULL. When creating a distributed hypertable, consider using [`create_distributed_hypertable`](#create_distributed_hypertable) in place of `create_hypertable`. |
+| `data_nodes` | This is the set of data nodes that will be used for this table if it is distributed. This has no impact on non-distributed hypertables. If no data nodes are specified, a distributed hypertable will use all data nodes known by this instance. |
 | `chunk_target_size` | DEPRECATED - The target size of a chunk (including indexes) in `kB`, `MB`, `GB`, or `TB`. Setting this to `estimate` or a non-zero chunk size, e.g., `2GB` will enable adaptive chunking (a DEPRECATED feature). The `estimate` setting will estimate a target chunk size based on system information. Adaptive chunking is disabled by default. |
 | `chunk_sizing_func` | DEPRECATED - Allows setting a custom chunk sizing function for adaptive chunking (a DEPRECATED feature). The built-in chunk sizing function will be used by default. Note that `chunk_target_size` needs to be set to use this function.  |
 
@@ -431,9 +434,96 @@ PostGIS geospatial indexes).  During testing, you might check your
 total chunk sizes via the [`chunk_relation_size`](#chunk_relation_size)
 function.
 
-**Space partitions**: In most cases, it is advised for users not to use
+**Space partitions:** In most cases, it is advised for users not to use
 space partitions. The rare cases in which space partitions may be useful
 are described in the [add dimension][] section.
+
+---
+
+## create_distributed_hypertable() [](create_distributed_hypertable)
+
+Creates a TimescaleDB hypertable distributed across a multinode
+environment.  Use this function in place of [`create_hypertable`](#create_hypertable)
+when creating distributed hypertables.
+
+#### Required Arguments [](create_distributed_hypertable-required-arguments)
+
+|Name|Description|
+|---|---|
+| `main_table` | Identifier of table to convert to hypertable |
+| `time_column_name` | Name of the column containing time values as well as the primary column to partition by. |
+
+#### Optional Arguments [](create_distributed_hypertable-optional-arguments)
+
+|Name|Description|
+|---|---|
+| `partitioning_column` | Name of an additional column to partition by. If provided, the `number_partitions` argument must also be provided. |
+| `number_partitions` | Number of hash partitions to use for `partitioning_column`. Must be > 0. |
+| `chunk_time_interval` | Interval in event time that each chunk covers. Must be > 0. As of TimescaleDB v0.11.0, default is 7 days, unless adaptive chunking (DEPRECATED)  is enabled, in which case the interval starts at 1 day. For previous versions, default is 1 month. |
+| `create_default_indexes` | Boolean whether to create default indexes on time/partitioning columns. Default is TRUE. |
+| `if_not_exists` | Boolean whether to print warning if table already converted to hypertable or raise exception. Default is FALSE. |
+| `partitioning_func` | The function to use for calculating a value's partition.|
+| `associated_schema_name` | Name of the schema for internal hypertable tables. Default is "_timescaledb_internal". |
+| `associated_table_prefix` | Prefix for internal hypertable chunk names. Default is "_hyper". |
+| `migrate_data` | Set to `true` to migrate any existing `main_table` data to chunks in the new hypertable. A non-empty table will generate an error without this option. Note that, for large tables, the migration might take a long time. Defaults to false. |
+| `time_partitioning_func` | Function to convert incompatible primary time column values to compatible ones. The function must be `IMMUTABLE`. |
+| `replication_factor` | This will determine the number of data nodes each incoming write is written to.  The default value is 1 and this is required to be 1 or greater, but note that values greater than 1 are not recommended in the current implementation. |
+| `data_nodes` | This is the set of data nodes that will be used for this table.  If this is not present then all data nodes known by this instance will be used to distribute the hypertable. | 
+| `chunk_target_size` | DEPRECATED - The target size of a chunk (including indexes) in `kB`, `MB`, `GB`, or `TB`. Setting this to `estimate` or a non-zero chunk size, e.g., `2GB` will enable adaptive chunking (a DEPRECATED feature). The `estimate` setting will estimate a target chunk size based on system information. Adaptive chunking is disabled by default. |
+| `chunk_sizing_func` | DEPRECATED - Allows setting a custom chunk sizing function for adaptive chunking (a DEPRECATED feature). The built-in chunk sizing function will be used by default. Note that `chunk_target_size` needs to be set to use this function.  |
+
+#### Returns
+
+|Column|Description|
+|---|---|
+| `hypertable_id` | ID of the hypertable in TimescaleDB. |
+| `schema_name` | Schema name of the table converted to hypertable. |
+| `table_name` | Table name of the table converted to hypertable. |
+| `created` | True if the hypertable was created, false when `if_not_exists` is true and no hypertable was created. |
+
+#### Sample Usage [](create_distributed_hypertable-examples)
+
+Create a table `conditions` which will be partitioned across data
+nodes by the 'location' column.  Note that the number of space
+partitions is automatically equal to the number of data nodes assigned
+to this hypertable (all configured data nodes in this case, as
+`data_nodes` is not specified).
+```sql
+SELECT create_distributed_hypertable('conditions', 'time', 'location');
+```
+
+Create a table `conditions` using a specific set of data nodes.
+```sql
+SELECT create_distributed_hypertable('conditions', 'time', 'location',
+    data_nodes => '{ "data_node_1", "data_node_2", "data_node_4", "data_node_7" }');
+```
+
+#### Best Practices [](create_distributed_hypertable-best-practices)
+
+**Space partitions:** As opposed to the normal [`create_hypertable` best practices](#create_hypertable-best-practices),
+space partitions are highly recommended for distributed hypertables.
+Incoming data will be divided among data nodes based upon the space
+partition (the first one if multiple space partitions have been
+defined).  If there is no space partition, all the data for each time
+slice will be written to a single data node.
+
+**Time intervals:** Follow the same guideline in setting the `chunk_time_interval`
+as with [`create_hypertable`](#create_hypertable-best-practices),
+bearing in mind that the calculation needs to be based on the memory
+capacity of the data nodes.  However, one additional thing to
+consider, assuming space partitioning is being used, is that the
+hypertable will be evenly distributed across the data nodes, allowing
+a larger time interval.
+
+For example, assume you are ingesting 10GB of data per day and you
+have five data nodes, each with 64GB of memory.  If this is the only
+table being served by these data nodes, then you should use a time 
+interval of 1 week (7 * 10GB / 5 * 64GB ~= 22% main memory used for
+most recent chunks).
+
+If space partitioning is not being used, the `chunk_time_interval`
+should be the same as the non-distributed case, as all of the incoming
+data will be handled by a single node.
 
 ---
 


api.md
```diff
@@ -255,6 +256,8 @@ still work on the resulting hypertable.
 | `associated_table_prefix` | Prefix for internal hypertable chunk names. Default is "_hyper". |
 | `migrate_data` | Set to `true` to migrate any existing `main_table` data to chunks in the new hypertable. A non-empty table will generate an error without this option. Note that, for large tables, the migration might take a long time. Defaults to false. |
 | `time_partitioning_func` | Function to convert incompatible primary time column values to compatible ones. The function must be `IMMUTABLE`. |
+| `replication_factor` | If non-zero, will create a distributed hypertable with each chunk being written to `replication_factor` different data nodes. The default value is 0. When creating a distributed hypertable, consider using [`create_distributed_hypertable`](#create_distributed_hypertable) in place of `create_hypertable`.|
```
dianasaur323: Do you want to specify that we don't really support >1 right now? Or that it's not recommended?

api.md
```diff
@@ -431,12 +434,72 @@ PostGIS geospatial indexes).  During testing, you might check your
...
+if multiple space partitions have been defined).  However, if there is
+no space partition, all the data for a time slice will be written to a
+single data node, negating much of the benefit of distributing the
+hypertable.
+
+**Time intervals:** Follow the same guideline in setting the `chunk_time_interval`
+as with [`create_hypertable`](#create_hypertable-best-practices),
+bearing in mind that the calculation needs to be based on the memory
+capacity of the data nodes.  However, one additional thing to
+consider, assuming space paritioning is being used, is that the
```
dianasaur323: Small spelling nit - "space partitioning"

api.md
```diff
@@ -431,12 +434,72 @@ PostGIS geospatial indexes).  During testing, you might check your
...
+hypertable will be evenly distributed across the data nodes, allowing
+a larger time interval.  In this case you can consider the total
+memory across all the data nodes, though the incoming rate should also
+be multiplied by the `replication_factor` if greater than 1.
+
+If space partioning is not being used, the `chunk_time_interval`
```
dianasaur323: Haha, space partitioning is a struggle! Spelling nit.

api.md
```diff
@@ -431,12 +434,72 @@ PostGIS geospatial indexes).  During testing, you might check your
...
+
+**Time intervals:** Follow the same guideline in setting the `chunk_time_interval`
+as with [`create_hypertable`](#create_hypertable-best-practices),
+bearing in mind that the calculation needs to be based on the memory
+capacity of the data nodes.  However, one additional thing to
+consider, assuming space paritioning is being used, is that the
+hypertable will be evenly distributed across the data nodes, allowing
+a larger time interval.  In this case you can consider the total
+memory across all the data nodes, though the incoming rate should also
+be multiplied by the `replication_factor` if greater than 1.
```
dianasaur323: I find this part to be slightly confusing, and wonder if we want to keep it this way. Should we just tell people to run chunk size pretty to define the chunk interval instead?

api.md
```diff
@@ -431,12 +434,72 @@ PostGIS geospatial indexes).  During testing, you might check your
...
 ---
 
+## create_distributed_hypertable() [](create_distributed_hypertable)
+
+A specialized implementation of [`create_hypertable`](#create_hypertable)
+designed for creating hypertables distributed across multinode 
+environments.  The required and optional arguments are the same as for
+`create_hypertable`, as is the returned value.  The main difference is
+that this command will default the `replication_factor` optional
+parameter to a value of 1.
```
mkindahl: Also, the replication factor cannot be explicitly set to zero (or NULL), which it can with `create_hypertable`.

api.md
```diff
@@ -431,12 +434,72 @@ PostGIS geospatial indexes).  During testing, you might check your
...
+partitions is automatically equal to the number of data nodes in this
+case.
+```sql
+SELECT create_distributed_hypertable('conditions', 'time', 'location');
+```
+
+Create a table `conditions` using a specific set of data nodes.  Each
+chunk will be written to two data nodes.
+```sql
+SELECT create_distributed_hypertable('conditions', 'time', 'location', replication_factor => 2,
```
mkindahl: Not sure if we have some line length restrictions, but this line looks long.

api.md
mkindahl: Not sure you need to mention this, you have already said above that if there is no space partitioning, there is no partitioning of a time slice.

api.md
```diff
@@ -431,12 +434,72 @@ PostGIS geospatial indexes).  During testing, you might check your
...
+
+#### Best Practices [](create_distributed_hypertable-best-practices)
+
+**Space partitions:** As opposed to the normal [`create_hypertable` best practices]
+(#create_hypertable-best-practices), space partitions are highly
+recommended for distributed hypertables.  Incoming data will be
+divided among data nodes based upon the space partition (the first one
+if multiple space partitions have been defined).  However, if there is
+no space partition, all the data for a time slice will be written to a
+single data node, negating much of the benefit of distributing the
```
erimatnor: There might be other upsides to avoiding space partitioning, so I think we should be careful about giving too much of on opinion here without having tested pros and cons of different setups.  I am happy to see us recommend space partitioning, but we probably shouldn't say much about other setups until we know more.

For instance, we could recommend smaller chunk time intervals in case people choose to not use space partitioning? With small chunks, you would quite quickly be able to use all nodes and you would also get better parallelism for queries that e.g., aggregate a single device (like TSBS single-groupby-1).

api.md
WireBaron: I added a note here, though take a look as it feels this description is getting a little unwieldy.

api.md
WireBaron: Fixed

api.md
WireBaron: For some reason my fingers always trip up on this word, fixed.

api.md
WireBaron: Doesn't chunk size pretty just show the sizes of the existing chunks?  It doesn't seem like that would be useful to make an a priori estimate of what the chunk size should be set to.

I'm not really sure the best way to improve this section.  Do you think that it's confusingly worded, or just too mathy?  If it's the second, I could add a concrete example, similar to what we have for create_hypertable.

dianasaur323: Hmm.. yeah, I agree - let's add a concrete example and call it a day?

api.md
WireBaron: Also updated the default

api.md
```diff
@@ -255,6 +256,8 @@ still work on the resulting hypertable.
 | `associated_table_prefix` | Prefix for internal hypertable chunk names. Default is "_hyper". |
 | `migrate_data` | Set to `true` to migrate any existing `main_table` data to chunks in the new hypertable. A non-empty table will generate an error without this option. Note that, for large tables, the migration might take a long time. Defaults to false. |
 | `time_partitioning_func` | Function to convert incompatible primary time column values to compatible ones. The function must be `IMMUTABLE`. |
+| `replication_factor` | If non-zero, will create a distributed hypertable with each chunk being written to `replication_factor` different data nodes. The default value is 0. When creating a distributed hypertable, consider using [`create_distributed_hypertable`](#create_distributed_hypertable) in place of `create_hypertable`.  Note that values greater than 1 are not recommended in the current implementation.|
```
solugebefola: The use of `replication_factor` in the description is confusing.  Consider "If non-zero, will create a distributed hypertable with each chunk being written to the number of data nodes corresponding to `replication_factor`"

api.md
WireBaron: Added example.

api.md
```diff
@@ -431,12 +434,72 @@ PostGIS geospatial indexes).  During testing, you might check your
...
 function.
 
-**Space partitions**: In most cases, it is advised for users not to use
+**Space partitions:** In most cases, it is advised for users not to use
 space partitions. The rare cases in which space partitions may be useful
 are described in the [add dimension][] section.
 
 ---
 
+## create_distributed_hypertable() [](create_distributed_hypertable)
```
solugebefola: Although this shares arguments with `create_hypertable` it should have its own arguments section, etc., even if is a duplicate.  This prevents the user from having to go back and forth on the page to utilize the api.

api.md
```diff
@@ -431,12 +434,72 @@ PostGIS geospatial indexes).  During testing, you might check your
...
+chunk will be written to two data nodes.
+```sql
+SELECT create_distributed_hypertable('conditions', 'time', 'location', replication_factor => 2,
+    data_nodes => '{ "data_node_1", "data_node_2", "data_node_4", "data_node_7" }');
+```
+
+#### Best Practices [](create_distributed_hypertable-best-practices)
+
+**Space partitions:** As opposed to the normal [`create_hypertable` best practices]
+(#create_hypertable-best-practices), space partitions are highly
```
solugebefola: You cannot have a line break between the two parts of the link here 
> [`create_hypertable` best practices] (#create_hypertable-best-practices)

api.md
WireBaron: Added a note to that effect.

api.md
WireBaron: I'm not sure either, but I did break the data_nodes argument down to their own line.  This remainder here is still less than other sql lines we have in this file.

api.md
WireBaron: I did want to explicitly call this out, as it may not be immediately obvious to someone first playing around with this.

api.md
```diff
@@ -255,6 +256,8 @@ still work on the resulting hypertable.
 | `associated_table_prefix` | Prefix for internal hypertable chunk names. Default is "_hyper". |
 | `migrate_data` | Set to `true` to migrate any existing `main_table` data to chunks in the new hypertable. A non-empty table will generate an error without this option. Note that, for large tables, the migration might take a long time. Defaults to false. |
 | `time_partitioning_func` | Function to convert incompatible primary time column values to compatible ones. The function must be `IMMUTABLE`. |
+| `replication_factor` | If non-zero, will create a distributed hypertable with each chunk being written to `replication_factor` different data nodes. The default value is 0. When creating a distributed hypertable, consider using [`create_distributed_hypertable`](#create_distributed_hypertable) in place of `create_hypertable`.|
+| `data_nodes` | This is the set of data nodes that will be used for this table if it is distributed. This has no impact if `replication_factor` is 0. If `replication factor is not 0, and this is not present, all data nodes known by this instance will be used to distribute the hypertable. |
```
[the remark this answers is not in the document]
WireBaron: Went with not set or 0.

api.md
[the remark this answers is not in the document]
WireBaron: changed.

api.md
```diff
@@ -431,12 +434,72 @@ PostGIS geospatial indexes).  During testing, you might check your
 total chunk sizes via the [`chunk_relation_size`](#chunk_relation_size)
 function.
 
-**Space partitions**: In most cases, it is advised for users not to use
+**Space partitions:** In most cases, it is advised for users not to use
```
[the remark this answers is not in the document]
WireBaron: Perhaps, but I'm not sure I want to tackle this right now.

api.md
```diff
@@ -431,12 +434,72 @@ PostGIS geospatial indexes).  During testing, you might check your
...
+`create_hypertable`, as is the returned value.  The main difference is
+that this command will default the `replication_factor` optional
+parameter to a value of 1.
+
+#### Sample Usage [](create_distributed_hypertable-examples)
+
+Create a table `conditions` spread across the entire distributed
+database.
+```sql
+SELECT create_distributed_hypertable('conditions', 'time');
```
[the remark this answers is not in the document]
WireBaron: Removed.

api.md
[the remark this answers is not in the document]
WireBaron: Removed replication factor from this example.

api.md
WireBaron: Removed the "negating much of the benefit..." section.

api.md
```diff
@@ -431,12 +434,72 @@ PostGIS geospatial indexes).  During testing, you might check your
...
-**Space partitions**: In most cases, it is advised for users not to use
+**Space partitions:** In most cases, it is advised for users not to use
 space partitions. The rare cases in which space partitions may be useful
 are described in the [add dimension][] section.
 
 ---
 
+## create_distributed_hypertable() [](create_distributed_hypertable)
+
+A specialized implementation of [`create_hypertable`](#create_hypertable)
```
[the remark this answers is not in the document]
WireBaron: Reworded this section.

api.md
WireBaron: Changed as you suggested

api.md
WireBaron: I was wondering if I could get away with this.  Copied the sections down and cleaned up the wording of replication_factor and data_nodes (which become much cleaner in this context).

api.md
WireBaron: fixed

WireBaron:
@solugebefola 
I had actually originally based this off of master.  I noticed that there was a clustering branch I should have used while looking at another PR and rebased to that branch as part of my most recent change.

api.md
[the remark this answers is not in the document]
dianasaur323: Ha, let's create a pivotal issue so we don't forget. I'm on it.

api.md
```diff
@@ -431,12 +434,104 @@ PostGIS geospatial indexes).  During testing, you might check your
...
+
+**Time intervals:** Follow the same guideline in setting the `chunk_time_interval`
+as with [`create_hypertable`](#create_hypertable-best-practices),
+bearing in mind that the calculation needs to be based on the memory
+capacity of the data nodes.  However, one additional thing to
+consider, assuming space partitioning is being used, is that the
+hypertable will be evenly distributed across the data nodes, allowing
+a larger time interval.  In this case you can consider the total
+memory across all the data nodes, though the incoming rate should also
+be multiplied by the `replication_factor` if greater than 1.
```
erimatnor: Again, I think we should not mention replication_factor here. Let's add that when the feature is ready.

api.md
```diff
@@ -431,12 +434,104 @@ PostGIS geospatial indexes).  During testing, you might check your
...
+| `hypertable_id` | ID of the hypertable in TimescaleDB. |
+| `schema_name` | Schema name of the table converted to hypertable. |
+| `table_name` | Table name of the table converted to hypertable. |
+| `created` | True if the hypertable was created, false when `if_not_exists` is true and no hypertable was created. |
+
+#### Sample Usage [](create_distributed_hypertable-examples)
+
+Create a table `conditions` which will be partitioned across data
+nodes by the 'location' column.  Note that the number of space
+partitions is automatically equal to the number of data nodes in this
```
erimatnor: number of configured data nodes in the database

api.md
```diff
@@ -431,12 +434,104 @@ PostGIS geospatial indexes).  During testing, you might check your
...
 
 ---
 
+## create_distributed_hypertable() [](create_distributed_hypertable)
+
+Creates a TimescaleDB hypertable distributed across a multinode
+environment.  This is very similar to [`create_hypertable`](#create_hypertable),
+and in fact has all of the same required and optional arguments.  The
+main difference is that this command will default the
+`replication_factor` optional parameter to a value of 1, and will
```
erimatnor: Do not go into replication_factor here. IMO, there's generally too much mention of replication_factor here considering replication is not a feature we're launching at this point.

I would suggest something simple like this:


"Creates a distributed hypertable across multiple data nodes. Use this function in place of `create_hypertable` when creating distributed hypertables."

api.md
```diff
@@ -255,6 +256,8 @@ still work on the resulting hypertable.
 | `associated_table_prefix` | Prefix for internal hypertable chunk names. Default is "_hyper". |
 | `migrate_data` | Set to `true` to migrate any existing `main_table` data to chunks in the new hypertable. A non-empty table will generate an error without this option. Note that, for large tables, the migration might take a long time. Defaults to false. |
 | `time_partitioning_func` | Function to convert incompatible primary time column values to compatible ones. The function must be `IMMUTABLE`. |
+| `replication_factor` | If set to 1 or greater, will create a distributed hypertable with each chunk being written to the number of data nodes corresponding to `replication_factor`. The default value is NULL. When creating a distributed hypertable, consider using [`create_distributed_hypertable`](#create_distributed_hypertable) in place of `create_hypertable`.  Note that values greater than 1 are not recommended in the current implementation.|
```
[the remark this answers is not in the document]
WireBaron: I'm making the suggested change, but it doesn't leave me with particularly warm fuzzy feeling.  It seems wrong that we're not explicitly disallowing this to be set higher than 1, we're just saying it's not recommended and not telling them what it does.

api.md
```diff
@@ -255,6 +256,8 @@ still work on the resulting hypertable.
...
+| `data_nodes` | This is the set of data nodes that will be used for this table if it is distributed. This has no impact if `replication_factor` is not set or 0. If `replication factor` is not 0, and this is not present, all data nodes known by this instance will be used to distribute the hypertable. |
```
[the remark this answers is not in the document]
WireBaron: Nice, using your wording here.

api.md
WireBaron: Modified as you suggested.  We should look at just removing replication_factor from the exposed API here for the public beta, unless we're planning on building up more support by then.

api.md
WireBaron: Went a little more verbose with "number of data nodes assigned to this hypertable (all configured data nodes in this case, as `data_nodes` is not specified)."

api.md
WireBaron: This still feels a bit like we're setting the users up to fail if they play around with the replication_factor rather than simply preventing them from doing so.  Removing this line, but we need to make this more coherent before public beta.

PR #190, opened 2019-07-15
Clarify docker-compose method

spolcyn:
Previous instructions didn't explicitly specify the need to change the YAML file


solugebefola:
Add label please

solugebefola:
Is this only for version `1.4` and later?

solugebefola:
It seems like something that should apply to docs for all versions?

PR #200, opened 2019-07-17
Adjust documentation regarding metadata table

svenklemm:


using-timescaledb/telemetry.md
```diff
@@ -48,8 +48,9 @@ the `UUID`s contain the current timestamp. For full transparency, we expose a
 new API function, [`get_telemetry_report`][get_telemetry_report], that returns
 a text string of the exact JSON that is sent to our servers.
 
-Additionally any content of the table `_timescaledb_catalog.telemetry_metadata`
-and the value of `timescaledb_telemetry.cloud` will be included in the telemetry report.
+Additionally any content of the table `_timescaledb_catalog.metadata` which has
```
solugebefola: Are these items not part of what gets returned from `get_telemetry_report`?

svenklemm: yes they are but the actual items where not changed only the table name changed so the output is the same as before

PR #204, opened 2019-07-17
Add CASCADE option to delete_data_node and warning

k-rus:
It is important to set CASCADE option to true, thus it is added in the example and warning is given. In sync with `delete_data_node` API doc.

### DIFF
--- a/getting-started/scaling-out.md
+++ b/getting-started/scaling-out.md
@@ -86,7 +86,7 @@ foreign data wrapper and any data node (server) objects.
 Deleting a data node is done by calling `delete_data_node`:
 
 ```sql
-SELECT delete_data_node('node1', cascade=>'true');
+SELECT delete_data_node('node1', cascade=>true);
 ```
 >:TIP: Note that a data node cannot be deleted if it contains data for a
 hypertable, since otherwise data would be lost.
--- a/getting-started/scaling-out.md
+++ b/getting-started/scaling-out.md
@@ -86,11 +86,14 @@ foreign data wrapper and any data node (server) objects.
 Deleting a data node is done by calling `delete_data_node`:
 
 ```sql
-SELECT delete_data_node('node1');
+SELECT delete_data_node('node1', cascade=>'true');
 ```
 >:TIP: Note that a data node cannot be deleted if it contains data for a
 hypertable, since otherwise data would be lost.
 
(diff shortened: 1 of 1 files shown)


getting-started/scaling-out.md
```diff
@@ -86,11 +86,14 @@ foreign data wrapper and any data node (server) objects.
...
 
 ```sql
-SELECT delete_data_node('node1');
+SELECT delete_data_node('node1', cascade=>'true');
 ```
 >:TIP: Note that a data node cannot be deleted if it contains data for a
 hypertable, since otherwise data would be lost.
 
+>:WARNING: Although the `cascade` parameter is not strictly a required argument,
+you *MUST* set it to `true` to avoid breaking necessary user mappings.
```
dianasaur323: Per Brian's comment, I was going to remove the last portion and just go with >:WARNING: Although the `cascade` parameter is not strictly a required argument, you *MUST* set it to `true`?

So perhaps update that here too so that they match?

k-rus: I will sync with the actual change if it will be in https://github.com/timescale/docs.timescale.com-content/pull/203

dianasaur323: Actually, I kind of like perhaps : you MUST set it to true in this current release to avoid putting TimescaleDB in an error state?

k-rus: Synced

getting-started/scaling-out.md
```diff
@@ -86,11 +86,15 @@ foreign data wrapper and any data node (server) objects.
 Deleting a data node is done by calling `delete_data_node`:
 
 ```sql
-SELECT delete_data_node('node1');
+SELECT delete_data_node('node1', cascade=>'true');
```
svenklemm: should remove the quotes around true

k-rus: Sorry, blindly copied from #203. Fixed

k-rus: Actually it works correctly with quotes too, so it doesn't require to fix #203. But ofc it is cleaner without them.
Thanks for quick approval.

svenklemm: yes it works with quotes as well but since they are not needed they should be removed from #203 too i think

dianasaur323:
@svenklemm okay I'll fix it

PR #214, opened 2019-07-30
Fix CentOS 7 install sudo error

spolcyn:
Running just `cat` with superuser privileges results in a permissions error. Using `sudo tee` fixes this.


solugebefola:
Add @RobAtticus as a reviewer?

PR #216, opened 2019-08-06
Update install prereqs to reflect latest

RobAtticus:
Since some of our binary builds are built only against 10.9+ and
11.4+, we update the pre-requisites to reflect that. In addition,
we now support Debian 10 (buster).

### DIFF
--- a/getting-started/installation-apt-debian.md
+++ b/getting-started/installation-apt-debian.md
@@ -2,11 +2,11 @@
 
 This will install TimescaleDB via `apt` on Debian distros.
 
-**Note: TimescaleDB requires PostgreSQL 9.6.3+, 10.2+, or 11.0+**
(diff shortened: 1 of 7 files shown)


getting-started/installation-apt-debian.md
```diff
@@ -2,11 +2,11 @@
 
 This will install TimescaleDB via `apt` on Debian distros.
 
-**Note: TimescaleDB requires PostgreSQL 9.6.3+, 10.2+, or 11.0+**
+**Note: TimescaleDB requires PostgreSQL 9.6.3+, 10.9+, or 11.4+**
```
svenklemm: why is 9.6 not affected? did you adjust package metadata as well?

RobAtticus: I did not update the version I built against for 9.6 (not sure why), so as far as I know, it should still work back to 9.6.3 on platforms with that version

PR #223, opened 2019-09-12
Clarify that trial does not require contacting sales

mfreed:


solugebefola:
@mfreed label versions?

PR #226, opened 2019-10-04
Update documentation for attach_data_node and detach_data_node

mkindahl:
The documentation for `attach_data_node` and `detach_data_node` is
updated to match the implementation.

### DIFF
--- a/api.md
+++ b/api.md
@@ -253,7 +253,7 @@ data node by:
 2. Creating the database given in `database` that will serve as the
    new data node.
 3. Loading the TimescaleDB extension in the new database.
-4. Setting metadata to make data node part of the distributed
+4. Setting metadata to make the data node part of the distributed
    database.
 
 #### Errors
--- a/api.md
+++ b/api.md
@@ -202,7 +202,8 @@ function.
 
 | Name              | Description                                   |
 |-------------------|-----------------------------------------------|
-| `if_not_attached` | Prevents error if data node is already attached to the hypertable. A notice will be printed that the data node is attached. |
+| `if_not_attached` | Prevents error if the data node is already attached to the hypertable. A notice will be printed that the data node is attached. Defaults to `FALSE`. |
+| `repartition`     | Change the configuration so that all the attached data nodes are used. Defaults to `TRUE`. |
 
 #### Returns
 
@@ -252,7 +253,7 @@ data node by:
 2. Creating the database given in `database` that will serve as the
    new data node.
 3. Loading the TimescaleDB extension in the new database.
-4. Setting metadata to make the data node part of the distributed
+4. Setting metadata to make data node part of the distributed
    database.
 
 #### Errors
@@ -778,11 +779,7 @@ partition across
 
 #### Returns
 
-| Column               | Description                              |
-|----------------------|------------------------------------------|
-| `hypertable_id`      | Hypertable id of the modified hypertable |
-| `node_hypertable_id` | Hypertable id on the remote data node    |
-| `node_name`          | Name of the attached data node     |
+The number of hypertables the data node was detached from.
 
 #### Errors
 
--- a/api.md
+++ b/api.md
@@ -203,7 +203,7 @@ function.
 | Name              | Description                                   |
 |-------------------|-----------------------------------------------|
 | `if_not_attached` | Prevents error if the data node is already attached to the hypertable. A notice will be printed that the data node is attached. Defaults to `FALSE`. |
-| `repartition`     | Change the configuration so that all the attached data nodes are used. Defaults to `TRUE`. |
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -202,7 +202,8 @@ function.
 
 | Name              | Description                                   |
 |-------------------|-----------------------------------------------|
-| `if_not_attached` | Prevents error if data node is already attached to the hypertable. A notice will be printed that the data node is attached. |
+| `if_not_attached` | Prevents error if data node is already attached to the hypertable. A notice will be printed that the data node is attached. Defaults to `FALSE`. |
+| `repartition`     | Expand the number of partitions in the space dimension to the number of data nodes after attaching this data node. Defaults to `TRUE`. |
```
erimatnor: Maybe we should say something like:

```
Allow updates to the hypertable's partitioning configuration to ensure the attached data node will be utilized.
```

This gives more of the _why_ as opposed to the _how_.

api.md
```diff
@@ -778,11 +779,7 @@ partition across
 
 #### Returns
 
-| Column               | Description                              |
-|----------------------|------------------------------------------|
-| `hypertable_id`      | Hypertable id of the modified hypertable |
-| `node_hypertable_id` | Hypertable id on the remote data node    |
-| `node_name`          | Name of the attached data node     |
+The number of data nodes detached.
```
erimatnor: Hmmm, we're only detaching one specifc data node so this seems odd. I suggest writing something like:

```
The number of hypertables this data node was detached from.
```

api.md
mkindahl: Done, approximately. PTAL.

api.md
```diff
@@ -252,7 +253,7 @@ data node by:
 2. Creating the database given in `database` that will serve as the
    new data node.
 3. Loading the TimescaleDB extension in the new database.
-4. Setting metadata to make the data node part of the distributed
```
k-rus: I don't think the removing `the` is appropriate, but I am not an English native person.

api.md
```diff
@@ -202,7 +202,8 @@ function.
 
 | Name              | Description                                   |
 |-------------------|-----------------------------------------------|
-| `if_not_attached` | Prevents error if data node is already attached to the hypertable. A notice will be printed that the data node is attached. |
+| `if_not_attached` | Prevents error if data node is already attached to the hypertable. A notice will be printed that the data node is attached. Defaults to `FALSE`. |
+| `repartition`     | Change the configuration so that all attached data nodes are used. Defaults to `TRUE`. |
```
k-rus: I haven't heard it is implemented. I guess you know better, since you were working on this.

api.md
mkindahl: One of my more common typos. Thanks for catching it.

api.md
mkindahl: It was there when I looked at the code. Just documenting what is there.

k-rus: If it is not implemented, but only provided to the user, I think it is good to have a note in the documentation now. In a sense it is the part of documentation limitations.

mkindahl: Not sure I follow. Looking at the code I see a reconfiguration of the hypertable in the code when `repartition` is true. It looks like this:

```
		if (repartition)
		{
			ts_dimension_set_number_of_slices(dim, num_nodes & 0xFFFF);

			ereport(NOTICE,
					(errmsg("the number of partitions in dimension \"%s\" was increased to %u",
							NameStr(dim->fd.column_name),
							num_nodes),
					 errdetail("To make use of all attached data nodes, a distributed "
							   "hypertable needs at least as many partitions in the first "
							   "closed (space) dimension as there are attached data nodes.")));
		}
```

Not sure what changes you want to have in the documentation.

api.md
```diff
@@ -202,7 +202,8 @@ function.
 
 | Name              | Description                                   |
 |-------------------|-----------------------------------------------|
-| `if_not_attached` | Prevents error if data node is already attached to the hypertable. A notice will be printed that the data node is attached. |
+| `if_not_attached` | Prevents error if the data node is already attached to the hypertable. A notice will be printed that the data node is attached. Defaults to `FALSE`. |
+| `repartition`     | Change the configuration so that all the attached data nodes are used. Defaults to `TRUE`. |
```
```suggestion
| `repartition`     | Change the partitioning configuration so that all the attached data nodes are used. Defaults to `TRUE`. |
```

api.md
```diff
@@ -252,7 +253,7 @@ data node by:
 2. Creating the database given in `database` that will serve as the
    new data node.
 3. Loading the TimescaleDB extension in the new database.
-4. Setting metadata to make the data node part of the distributed
+4. Setting metadata to make data node part of the distributed
```
```suggestion
4. Setting metadata to make the data node part of the distributed
```

api.md
erimatnor: Yeah, think `the` should be there.

PR #232, opened 2019-10-22
Document password authentication for multinode

mkindahl:
Information about how to set up password authentication for data nodes.

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -1,15 +1,17 @@
 # Getting Started
 
-A TimescaleDB behaves in many respects like a standard PostgreSQL database. It:
-- Coexists with other TimescaleDBs and PostgreSQL databases on a PostgreSQL
-server.
+A TimescaleDB database behaves in many respects like a standard PostgreSQL
+database. It:
+
+- Coexists with other TimescaleDB databases and PostgreSQL databases
+  on a PostgreSQL server.
 - Uses SQL as its interface language.
 - Contains standard database objects like tables, indexes, and triggers.
 - Uses common PostgreSQL connectors to third-party tools.
 
-The way the database accomplishes this synchronicity is through its packaging
-as a PostgreSQL extension, whereby a standard PostgreSQL database is
-transformed into a TimescaleDB.
+The way the database accomplishes this synchronicity is through its
+packaging as a PostgreSQL extension, whereby a standard PostgreSQL
+database is transformed into a TimescaleDB database.
 
 <img class="main-content__illustration" style="width: 100%" src="//assets.iobeam.com/images/docs/illustration-hierarchy.svg" alt="hierarchy illustration"/>
 
@@ -26,9 +28,11 @@ Getting started with TimescaleDB involves a few steps:
 1. [Creating a standard PostgreSQL database and enabling the TimescaleDB extension on the database.][setup]
 1. [Creating a hypertable to store your data.][create-hypertables]
 1. [Migrating your data to a hypertable][migrate] (optional)
+1. [Setting up data node authentication for distributed hypertables][distdb-auth]
 
 [hypertables]: /introduction/architecture#hypertables
 [install]: /getting-started/installation
 [setup]: /getting-started/setup
 [create-hypertables]: /getting-started/creating-hypertables
+[distdb-auth]: /getting-started/data-node-authentication
 [migrate]: /getting-started/migrating-data
--- a/getting-started/data-node-authentication.md
+++ b/getting-started/data-node-authentication.md
@@ -1,0 +1,99 @@
+# Data Node Authentication [](distdb-auth)
+
+In a multi-node environment, it is necessary to ensure that the access
+node have the necessary roles and permissions on all the data nodes
+where distributed hypertables are placed.
+
+This is critical since when TimescaleDB is executing a query or
+command it will use the same role on the data nodes as on the access
+node. This means that the same roles need to exist on the data nodes
+and the access node need to have authentication set up for those
+roles.
+
+In a typical setup, you will create the distributed hypertable in a
+database and use the same database on all the data nodes. This means
+that you need to ensure that a user that is connected to the access
+node can access the same databases on the data nodes.
+
+>:NOTE: Setting up a secure system is a complex task and this section
+> should not be read as recommending any particular security measures
+> for securing your system. The chapter contains technical
+> instructions for how to set up the different authentication methods.
+
+## Setting up Password Authentication [](distdb-auth-pass)
+
+The easiest authentication to set up is password
+authentication. Configuring password authentication between the access
+node and data nodes is a two step process:
+
+1. Each data node needs to be configured to accept and authenticate
+   connections using passwords.
+
+2. The access node needs data node passwords for each role using the
+   access node. We will deal with each step separately below.
+
+In this section, the focus is on setting up SCRAM SHA-256 password
+authentication. For other password authentication methods, [see the
+PostgreSQL manual][auth-password].
+
+>:NOTE: The `password` method should not be used since it sends the
+> password in cleartext, and the `md5` is using a cryptographic hash
+> function that is no longer considered secure.
+
+### Setting up Password Authentication on the Data Node [](distdb-auth-pass-data-node)
+
+In order to set up password authentication for a user `new_user`, it
+is necessary to update the HBA (Host-Based Authentication) file on the
+data nodes to enable password authentication for `new_user`. The
+default name of the HBA file is `pg_hba.conf`. Details of [the
+`pg_hba.conf` format can be found in the PostgreSQL
+manual.][postgresql-hba].
+
+In this case, you want to give `new_user` in the `example.com` domain
+access to all databases on the data node, so add the following line to
+the `pg_hba.conf`.
+
+```
+host    all    new_user    .example.com    scram-sha-256
+```
+
+Once you have set up the authentication method on the data node, you
+need to check the configuration file (normally `postgresql.conf` in
+the data directory of the data node server) and make sure that the
+`hba_file` option is correctly set.
+
+### Setting up Passwords on the Access Node [](distdb-auth-pass-access-node)
+
+To execute queries across data nodes, the access node needs to connect
+to data nodes as the user executing the query. This means that the
+access node needs passwords for the current user on each data node,
+and these are stored in a password file on the access node.
+
+The default name of the password file is `passfile` and it is located
+in the data directory for the PostgreSQL server. You can set the name
+to something else in the `postgresql.conf` file for the access node
+using the option `timescaledb.passfile`.
+
+```
+timescaledb.passfile = 'pgpass.conf'
+```
+
+**Password file format.** The password file format is [described in
+the PostgreSQL manual][passfile-format] and consists of lines of the
+format:
+
+```
+hostname:port:database:username:password
+```
+
+So, in order to add a password for our user `new_user` and use
+password `xyzzy`, you can create a password file with the following
+contents:
+
+```
+*:*:*:new_user:xyzzy
+```
+
+[postgresql-hba]: https://www.postgresql.org/docs/current/auth-pg-hba-conf.html
+[auth-password]: https://www.postgresql.org/docs/current/auth-password.html
+[passfile-format]: https://www.postgresql.org/docs/current/libpq-pgpass.html
--- a/page-index/page-index.js
+++ b/page-index/page-index.js
@@ -61,6 +61,10 @@ const pageIndex = [
                         href: "migrate-from-postgresql",
                         to: "/getting-started/migrating-data"
                     }, {
+			Title: "Data Node Authentication",
+			type: PAGE,
+                        href: "data-node-authentication",
+                    }, {
                         href: "scaling-out",
                         to: "/getting-started/scaling-out"
                     },


getting-started.md
```diff
@@ -26,9 +27,12 @@ Getting started with TimescaleDB involves a few steps:
 1. [Creating a standard PostgreSQL database and enabling the TimescaleDB extension on the database.][setup]
 1. [Creating a hypertable to store your data.][create-hypertables]
 1. [Migrating your data to a hypertable][migrate] (optional)
+1. [Setting up data node authentication for distributed hypertables][data-node-authentication]
 
 [hypertables]: /introduction/architecture#hypertables
 [install]: /getting-started/installation
 [setup]: /getting-started/setup
 [create-hypertables]: /getting-started/creating-hypertables
+[create-distributed-hypertables]: /getting-started/creating-distributed-hypertables
```
solugebefola: Is there a reference to this link somewhere in the main text?

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,70 @@
+# Data Node Authentication
+
+In a multi-node environment, it is necessary to ensure that the access
+node have the necessary roles and permissions on all the data nodes
+where distributed hypertables are placed.
+
+This is critical since when TimescaleDB is executing a query or
+command it will use the same role on the data nodes for the
+fragments. This means that the same role need to exist on the data
+nodes and authenticate to them.
```
solugebefola:
```suggestion
nodes for appropriate authentication.
```
Not sure what you were going for, so I took a guess?

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,70 @@
...
+
+In a typical setup, you will deploy the distributed hypertable in a
+database and use the same database on all the data nodes. This means
+that we need to ensure that a user that is connected to the access
```
solugebefola:
```suggestion
that you need to ensure that a user that is connected to the access
```
to stay consistent

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,70 @@
...
+```
+host    all    new_user    .example.com    scram-sha-256
+```
+
+The assumption is that all the nodes are in the =*.example.com=
+domain.
+
+### Setting up passwords
+
+Passwords are not passed on the terminal but are added to a special
```
solugebefola:
```suggestion
Passwords are not passed through the command line, but are added to a special
```

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,70 @@
...
+password file. The default name of the password file is `passfile`
+located in the data directory for the PostgreSQL server, but it be set
```
solugebefola:
```suggestion
located in the data directory for the PostgreSQL server, but it can be set
```

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,70 @@
...
+in the `postgresql.conf` file for the access node using the option
+`timescaledb.passfile`. You also want to set the name of the HBA file
```
solugebefola:
```suggestion
`timescaledb.passfile`. You should also set the name of the HBA file
```

getting-started.md
```diff
@@ -1,15 +1,16 @@
...
 - Uses SQL as its interface language.
 - Contains standard database objects like tables, indexes, and triggers.
 - Uses common PostgreSQL connectors to third-party tools.
 
-The way the database accomplishes this synchronicity is through its packaging
-as a PostgreSQL extension, whereby a standard PostgreSQL database is
-transformed into a TimescaleDB.
+The way the database accomplishes this synchronicity is through its
+packaging as a PostgreSQL extension, whereby a standard PostgreSQL
+database is transformed into a TimescaleDB.
```
erimatnor:
```suggestion
database is transformed into a TimescaleDB database.
```

FWIW, I find this sentence odd and superfluous so could just as well remove altogether, imo.

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,70 @@
+# Data Node Authentication
+
+In a multi-node environment, it is necessary to ensure that the access
+node have the necessary roles and permissions on all the data nodes
+where distributed hypertables are placed.
+
+This is critical since when TimescaleDB is executing a query or
+command it will use the same role on the data nodes for the
+fragments. This means that the same role need to exist on the data
```
erimatnor:
```suggestion
This means that the same role needs to exist on the data
```
not sure I understand "fragments". We've never used this terminology before and it is confusing to me.

I think you can just delete "for the fragments".

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,70 @@
+# Data Node Authentication
+
+In a multi-node environment, it is necessary to ensure that the access
+node have the necessary roles and permissions on all the data nodes
+where distributed hypertables are placed.
+
+This is critical since when TimescaleDB is executing a query or
+command it will use the same role on the data nodes for the
```
erimatnor:
```suggestion
command it will use the same role on the data nodes.
```

getting-started/data-node-authentication.md
erimatnor: FWIW, I think @solugebefola's suggestion is good.

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,70 @@
...
+fragments. This means that the same role need to exist on the data
+nodes and authenticate to them.
+
+In a typical setup, you will deploy the distributed hypertable in a
```
erimatnor:
```suggestion
In a typical setup, you will create the distributed hypertable in a
```

I don't think your "deploying" hypertables.

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,70 @@
...
+is necessary to update the HBA (Host-Based Authentication) file,
+typically named `pg_hba.conf`, on the data nodes to enable password
+authentication for `new_user`. Details of the format of `pg_hba.conf`
+can be found in the [PostgreSQL manual, Chapter 20.1][postgresql-hba].
+
+In order to give `new_user` access to all databases on the data node,
+add the following line to the `pg_hba.conf` on the data node. This
+will use SCRAM SHA-256 password authentication. For other password
+authentication methods, see [PostgreSQL manual, Chapter
+20.5][auth-password].
```
erimatnor: I would skip the chapter number reference since this is self-evident by the link.

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,70 @@
...
+command it will use the same role on the data nodes for the
+fragments. This means that the same role need to exist on the data
+nodes and authenticate to them.
+
+In a typical setup, you will deploy the distributed hypertable in a
+database and use the same database on all the data nodes. This means
+that we need to ensure that a user that is connected to the access
+node can access the same databases on the data nodes.
+
+## Setting up password authentication
```
erimatnor:
```suggestion
## Setting up Password Authentication
```

This is the style in other files

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,70 @@
...
+
+In order to set up password authentication for a user `new_user`, it
```
erimatnor: I think this section would benefit from being split into subsections about what needs to be done on (1) data nodes, and (2) the access node

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,70 @@
...
+20.5][auth-password].
+
+```
+host    all    new_user    .example.com    scram-sha-256
+```
+
+The assumption is that all the nodes are in the =*.example.com=
+domain.
+
+### Setting up passwords
```
erimatnor:
```suggestion
### Setting up Passwords on the Access Node
```

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,70 @@
...
+Passwords are not passed on the terminal but are added to a special
+password file. The default name of the password file is `passfile`
+located in the data directory for the PostgreSQL server, but it be set
+in the `postgresql.conf` file for the access node using the option
+`timescaledb.passfile`. You also want to set the name of the HBA file
+using the `hba_file` option:
+
+```
+timescaledb.passfile = pgpass.conf
+hba_file = pg_hba.conf
```
erimatnor: While this is not wrong, the `pg_hba.conf` configured above only needs to be on the data nodes. The access node can be governed by another `pg_hba.conf` with different settings, but this makes it seem like the `pg_hba.conf` from the previous step needs to be on the access node.

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+**Password file format.* The password file format is described in
+[Chapter 34.15 in the PostgreSQL manual][passfile-format] and consists
```
erimatnor: Again, I don't think we should reference chapter numbers here. It is self-evident when clicking the link and chapter numbers might change.

getting-started.md
```diff
@@ -1,15 +1,16 @@
 # Getting Started
 
-A TimescaleDB behaves in many respects like a standard PostgreSQL database. It:
-- Coexists with other TimescaleDBs and PostgreSQL databases on a PostgreSQL
-server.
+A TimescaleDB behaves in many respects like a standard PostgreSQL
+database. It:
+
+- Coexists with other TimescaleDBs and PostgreSQL databases on a PostgreSQL server.
```
WireBaron: In line with Erik's suggestion, this should also be `TimescaleDB databases`.

getting-started/data-node-authentication.md
WireBaron: These need to be quoted strings for the config file, so `timescaledb.passfile = 'pgpass.conf'`.  Also, `pg_hba.conf` seemed to be the default value for `hba_file` when I was trying this earlier, so this might not be necessary.  Actually it may be worth telling users to check the value for this setting and then updating the appropriate file above rather than just defaulting to `pg_hba.conf`.

getting-started.md
mkindahl: Nope, temporarily added another list item but forgot to remove this one.

getting-started/data-node-authentication.md
mkindahl: Elaborated a bit to make it more clear.

getting-started/data-node-authentication.md
mkindahl: Agree. There were another one that I also replaced with a "you".

getting-started/data-node-authentication.md
mkindahl: Well... I was using "command-line" first, but it might not be the command-line in the sense that the user types it. Would like to have a different wording here for this. Not entirely sure what.

getting-started/data-node-authentication.md
mkindahl: Split it and described the setup for access node and data node separately.

mkindahl: Done. PTAL.

getting-started/data-node-authentication.md
mkindahl: Split. Describing the setup for the data node first.

mkindahl:
> I followed these steps on a clean postgres cluster (started with a ec2 instance and just added postgres11, timescaledb-tune, and built timescale from latest clustering). Distributed operations were still failing for me with `fe_sendauth: no password supplied` for scram-sha-256 or password authentication. So it seems like there might still be a step missing in here.

Try the instructions again. It was not very clear what needed to be done on the data node and the access node respectively.

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,89 @@
...
+So, in order to add a password for our user `new_user` and use
+password `xyzzy`, you can create a password file with the following
+contents:
+
+```
+*:*:*:new_user:xyzzy
+```
+
+[postgresql-hba]: https://www.postgresql.org/docs/11/auth-pg-hba-conf.html
+[auth-password]: https://www.postgresql.org/docs/11/auth-password.html
```
erimatnor: Let's use `current` instead of `11` here. Then we need not update with every PG release.

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,89 @@
...
+## Setting up Password Authentication [](distdb-auth.pass)
+
+The easiest authentication to set up is password authentication. It is
+not as strong as certificate authentication, but can be used for an
+initial setup or when you have good perimeter security. In this
+section, the focus is on setting up SCRAM SHA-256 password
+authentication. For other password authentication methods, [see the
+PostgreSQL manual][auth-password].
+
+> NOTE: The `passwd` method should not be used since it sends the
```
erimatnor: What's the `passwd` method? Does it refer to the `passwd` command on UNIX OSes? The PostgreSQL docs has a method called simply `password`. Is that what this is referring to?

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,89 @@
...
+node can access the same databases on the data nodes.
+
+> NOTE: Setting up a secure system is a complex task and this section
+> should not be read as recommending any particular security measures
+> for securing your system. The chapter contains technical
+> instructions for how to set up the different authentication methods.
+
+## Setting up Password Authentication [](distdb-auth.pass)
+
+The easiest authentication to set up is password authentication. It is
```
erimatnor: I think this lacks some initial context. I would start this section with something like:

`
An access node needs to connect to and authenticate with data nodes on behalf of the clients it serves. The easiest authentication mechanism to configure is password authentication. Configuring password authentication between the access node and data nodes is a two step process: (1) each data node needs to be configured to accept and authenticate connections using passwords, and (2) the access node needs a list of passwords to use for each user (role) that executes queries and commands across data nodes. We will deal with each step separately below.
`

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,89 @@
...
+not as strong as certificate authentication, but can be used for an
```
erimatnor: I would skip this comparison against other methods and/or move it to the `NOTE` below.

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+Once you have set up the authentication method on the data node, you
+need to check the configuration file (normally `postgresql.conf` in
+the data directory of the data node server) and make sure that the
+`hba_file` option is correctly set.
+
+### Setting up Passwords on the Access Node [](distdb-auth.pass.access-node)
+
+Passwords are not passed through the command line, but are added to a special
```
erimatnor: I suggest starting with more context:
`
To securely execute queries across data nodes, the access node needs to connect to data nodes as the user executing the query on the access node, preferably with a unique password. The access node therefore needs to obtain the password to use for each user. The necessary user-password mapping is currently stored in a password file on the access node. Note that the password for a user need not be the same as the password used by the client to authenticate with the access node. But it _must_ be the same as the user's password on the data nodes.
`

getting-started/data-node-authentication.md
mkindahl: Done.

getting-started/data-node-authentication.md
mkindahl: Yes, typo. Fixed.

getting-started/data-node-authentication.md
mkindahl: IMHO, there is context earlier in the chapter, but outlining the steps before describing them in detail can be added.

mkindahl: Done.

getting-started/data-node-authentication.md
mkindahl: Removed.

getting-started/data-node-authentication.md
mkindahl: Did something along these lines. PTAL.

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,99 @@
...
+Once you have set up the authentication method on the data node, you
+need to check the configuration file (normally `postgresql.conf` in
+the data directory of the data node server) and make sure that the
+`hba_file` option is correctly set.
+
+### Setting up Passwords on the Access Node [](distdb-auth.pass.access-node)
+
+To execute queries across data nodes, the access node needs to connect
+to data nodes as the user executing the query. This means that the
+access node needs passwords for each data node, which are stored in a
```
erimatnor:
```suggestion
access node needs passwords for each user, which are stored in a
```

mkindahl: Hmmmm, it is not clear if it is a user on a data node or on the access node. Will take a shot.

mkindahl: Done. PTAL.

page-index/page-index.js
```diff
@@ -60,6 +60,9 @@ const pageIndex = [
                         type: HIDDEN_REDIRECT,
                         href: "migrate-from-postgresql",
                         to: "/getting-started/migrating-data"
+                    }, {
+                        href: "distdb-auth",
```
solugebefola: You need to add `type: PAGE,` and `Title: <your title>` to make it visible

solugebefola: `href` needs to be the filename

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,99 @@
...
+database and use the same database on all the data nodes. This means
+that you need to ensure that a user that is connected to the access
+node can access the same databases on the data nodes.
+
+>:NOTE: Setting up a secure system is a complex task and this section
+> should not be read as recommending any particular security measures
+> for securing your system. The chapter contains technical
+> instructions for how to set up the different authentication methods.
+
+## Setting up Password Authentication [](distdb-auth.pass)
```
solugebefola: You cannot use periods in the anchor name
```suggestion
## Setting up Password Authentication [](distdb-auth-pass)
```

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,99 @@
...
+
+In this section, the focus is on setting up SCRAM SHA-256 password
+authentication. For other password authentication methods, [see the
+PostgreSQL manual][auth-password].
+
+>:NOTE: The `password` method should not be used since it sends the
+> password in cleartext, and the `md5` is using a cryptographic hash
+> function that is no longer considered secure.
+
+### Setting up Password Authentication on the Data Node [](distdb-auth.pass.data-node)
```
solugebefola:
```suggestion
### Setting up Password Authentication on the Data Node [](distdb-auth-pass-data-node)
```

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,99 @@
...
+```
+host    all    new_user    .example.com    scram-sha-256
+```
+
+Once you have set up the authentication method on the data node, you
+need to check the configuration file (normally `postgresql.conf` in
+the data directory of the data node server) and make sure that the
+`hba_file` option is correctly set.
+
+### Setting up Passwords on the Access Node [](distdb-auth.pass.access-node)
```
solugebefola:
```suggestion
### Setting up Passwords on the Access Node [](distdb-auth-pass-access-node)
```

getting-started/data-node-authentication.md
```diff
@@ -0,0 +1,99 @@
...
+   connections using passwords.
+
+2. The access node needs data node passwords for each role using the
+   access node. We will deal with each step separately below.
+
+In this section, the focus is on setting up SCRAM SHA-256 password
+authentication. For other password authentication methods, [see the
+PostgreSQL manual][auth-password].
+
+>:NOTE: The `password` method should not be used since it sends the
```
pmwkaa: I am slightly concerned about this section, since SCRAM authentication is a really a new way to do it and it does not really supported by a connection poolers yet. Most of folks live with `md5` and use it likely with TLS. I hope it will not confuse people

mkindahl: Yeah, good point, but I think there are dangers in using MD5 in our examples if people copy it and think it is safe. If they run into this problem and think a little, they should be fine.

PR #233, opened 2019-10-22
Update the scaling out section

erimatnor:
This updates the scaling out section to be up-to-date with the latest
multinode APIs and functionality.

### DIFF
--- a/getting-started/scaling-out.md
+++ b/getting-started/scaling-out.md
@@ -1,17 +1,16 @@
 # Scaling out using Distributed Hypertables
 
->:WARNING: Running TimescaleDB in a multi-node setup is currently in PRIVATE BETA.
-This method of deployment is not meant for production use. For more information, please
+>:WARNING: Distributed hypertables are currently in PRIVATE BETA and
+are not yet meant for production use. For more information, please
 [contact us][contact].
 
-TimescaleDB supports multi-node clustering by leveraging the hypertable and chunk primitives.
-Hypertables are used to handle large amount of data by breaking it up
-into smaller pieces (chunks), allowing operations to execute efficiently. When
-the amount of data is expected to be beyond what a single machine can handle,
-you can distribute the data chunks over several machines by using
-*distributed hypertables*.
-Distributed hypertables are similar to normal hypertables, but they
-add an additional layer of hypertable partitioning by distributing chunks
+Hypertables handle large amounts of data by breaking it up into
+smaller pieces (chunks), allowing operations to execute
+efficiently. When the amount of data is expected to be beyond what a
+single machine can handle, you can distribute the data chunks over
+several machines by using *distributed hypertables*.  Distributed
+hypertables are similar to normal hypertables, but they add an
+additional layer of hypertable partitioning by distributing chunks
 across *data nodes*.
 
 Data nodes together with an *access node* constitute a distributed database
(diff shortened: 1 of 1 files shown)


getting-started/scaling-out.md
```diff
@@ -1,17 +1,16 @@
...
-*distributed hypertables*.
-Distributed hypertables are similar to normal hypertables, but they
-add an additional layer of hypertable partitioning by distributing chunks
+Hypertables handle large amounts of data by breaking it up into
+smaller pieces (chunks), allowing operations to execute
+efficiently. When the amount of data is expected to be beyond what a
+single machine can handle, you can distribute the data chunks over
+several machines by using *distributed hypertables*.  Distributed
+hypertables are similar to normal hypertables, but they add an
+additional layer of hypertable partitioning by distributing chunks
```
mkindahl: Suggestion: "... but they allow partitioning the hypertable further by distributing..."

getting-started/scaling-out.md
```diff
@@ -30,72 +29,65 @@ and removed using [`delete_data_node`][delete_data_node].
...
-  existing PostgreSQL instance.
+* A data node is added as a local object on the access node and
+  contains the configuration needed to connect to a database on a
+  PostgreSQL instance.
 
-* You should already have a running PostgreSQL server on the data node host.
+* You should already have a running PostgreSQL server on an instance
+  that will host a data node database. The data node's database will
+  be created when executing the `add_data_node` command on the access
+  node and should _not_ exist prior to adding the data node.
```
mkindahl: It can exist, it will not fail, but it might be good to not encourage it.

```
	/* Create the database with the user as owner. This command might fail
	 * if the database already exists, but we catch the error and continue
	 * with the bootstrapping if it does. */
	res = remote_connection_execf(conn,
								  "CREATE DATABASE %s OWNER %s",
								  quote_identifier(dbname),
								  quote_identifier(username));
```

getting-started/scaling-out.md
```diff
@@ -30,72 +29,65 @@ and removed using [`delete_data_node`][delete_data_node].
...
 
-* Ensure that the data node has password authentication enabled
-  in their `pg_hba.conf` files for any non-superusers.
+* PostgreSQL instances that will act as data nodes are assumed to
+  contain the same roles and permissions as the access node instance
+  (or, at a minimum, the roles and permissions associated with
+  distributed hypertables). Currently, such roles and permissions need
```
mkindahl: Even though correct, I suggest that you remove the comment in the parentheses. It's a good way to distract the reader from the main focus.

getting-started/scaling-out.md
```diff
@@ -30,72 +29,65 @@ and removed using [`delete_data_node`][delete_data_node].
...
+  to be created manually, although there is a [utility
+  command][distributed_exec] that can be used to create roles and
+  permissions across all (or a subset of) data nodes.
```
mkindahl: Suggest to remove "all (or a subset of) ".

getting-started/scaling-out.md
```diff
@@ -30,72 +29,65 @@ and removed using [`delete_data_node`][delete_data_node].
...
-  hypertable should be stored.
+* Run `add_data_node` as a superuser.
 
-* Provide the remote password, which will be used by the current user
-  during access to the created remote data node.
+* Ensure that the access node's superuser has a way to
+  [authenticate][data-node-authentication] with the data node
+  instance. This can be done either by setting up password or
+  certificate authentication. Other ways of authenticating may be
+  supported in the future.
```
mkindahl: Suggest to focus on what we have now, the last sentence is redundant.

getting-started/scaling-out.md
```diff
@@ -30,72 +29,65 @@ and removed using [`delete_data_node`][delete_data_node].
...
+* Provide a name to use when referring to the data node from
+  the access node database.
 
-* Ensure that the bootstrap user used for connecting to the data node
-  is a superuser.
-  This is necessary since
-  `add_data_node` expects to be able to create a
-  database on the remote data node and create
-  a TimescaleDB extension within it.
+* Provide the host name (and optionally port) of the PostgreSQL
```
mkindahl: Suggest "...host name, and optionally port, of the PostgreSQL..."

getting-started/scaling-out.md
```diff
@@ -30,72 +29,65 @@ and removed using [`delete_data_node`][delete_data_node].
...
-  foreign data wrapper on the access node.
+After creating the data node:
 
-```sql
-SELECT add_data_node('node1', host => 'dn1.example.com',
-  password=>'<remote_password>', bootstrap_user=>'<superuser>',
-  bootstrap_password=>'<superuser_password>');
+* Ensure that non-superusers have `USAGE` privileges on the
+  `timescaledb_fdw` foreign data wrapper and any
+  data node objects they will use on the access node.
```
mkindahl: Hmmm.... maybe something we might consider to do automatically when adding the data node. No action here though.

getting-started/scaling-out.md
```diff
@@ -172,15 +162,22 @@ hypertable, you can use [`detach_data_node`][detach_data_node].
 SELECT detach_data_node('node1', hypertable => 'conditions');
 ```
 
-If a data node is storing data for a hypertable,
-then you will get an error.
+Note that you cannot detach a data node that still holds data for the
+hypertable. A common way to solve this issue is to use a retention
```
mkindahl: "common"? We have not really had it in production. Suggest "You can solve this by using the retention period of the data and make sure..."

I think you should also have a link to a description about the retention period.

getting-started/scaling-out.md
```diff
@@ -30,72 +29,65 @@ and removed using [`delete_data_node`][delete_data_node].
...
 
-* Data nodes are added as objects locally and connect to an
-  existing PostgreSQL instance.
+* A data node is added as a local object on the access node and
+  contains the configuration needed to connect to a database on a
+  PostgreSQL instance.
 
-* You should already have a running PostgreSQL server on the data node host.
+* You should already have a running PostgreSQL server on an instance
+  that will host a data node database. The data node's database will
```
solugebefola:
```suggestion
  that will host a data node. The data node's database will
```

getting-started/scaling-out.md
```diff
@@ -30,72 +29,65 @@ and removed using [`delete_data_node`][delete_data_node].
...
 
-SELECT add_data_node('node2', host => 'dn2.example.com',
-  password=>'<remote_password>', bootstrap_user=>'<superuser>',
-  bootstrap_password=>'<superuser_password>');
-```
-
-Any additional users that will access a distributed hypertable currently
-need their own user mappings per data node with a `user` and `password` option.
-A user mapping can be created for a data node as follows:
+* Ensure that each user of a distributed hypertable have a way to
```
solugebefola:
```suggestion
* Ensure that each user of a distributed hypertable has a way to
```

getting-started/scaling-out.md
```diff
@@ -172,15 +162,22 @@ hypertable, you can use [`detach_data_node`][detach_data_node].
...
+achieved by first blocking new chunks on the data node:
 
 ```sql
-SELECT detach_data_node('node1', hypertable => 'conditions');
-ERROR:  detaching data node "node1" would mean a data-loss for hypertable "conditions" since data node has the only data replica
-HINT:  Ensure the data node "node1" has no non-replicated data before detaching it.
+SELECT block_new_chunks('node1', hypertable => 'conditions');
 ```
 
+This will prohibit any new data to be stored on the data node. Once
```
solugebefola:
```suggestion
This will prohibit any new data from being stored on the data node. Once
```

getting-started/scaling-out.md
erimatnor: Question is if we should delve into the subtleties of this here. I do say _should_ and not _must_. I figure we'd keep it simple here.

getting-started/scaling-out.md
erimatnor: I am not too fond of this suggestion because I don't think distributed hypertables allow further partitioning. In fact, you can partition the same way on a single node and the only difference is where chunks are stored.

Also, FWIW, this is not new content, just updated because of formatting (line breaks).

getting-started/scaling-out.md
erimatnor: Not sure what you mean by `description about the retention period.`? This is more of a concept enabled by drop_chunks.

erimatnor: See update.

getting-started/scaling-out.md
```diff
@@ -30,72 +29,64 @@ and removed using [`delete_data_node`][delete_data_node].
...
-  database on the remote data node and create
-  a TimescaleDB extension within it.
+* Ensure that non-superusers have `USAGE` privileges on the
+  `timescaledb_fdw` foreign data wrapper and any
+  data node objects they will use on the access node.
 
-* Ensure that the local user has `USAGE` privileges on the `timescaledb_fdw`
-  foreign data wrapper on the access node.
+* Ensure that each user of a distributed hypertable has a way to
+  [authenticate][data-node-authentication] with the data nodes they
```
solugebefola: I'm assuming the `page-index.js` will have been modified to accommodate this page (from the other PR)?

erimatnor: I assume that too :)

PR #235, opened 2019-10-22
Add distributed_exec to the API docs

erimatnor:
This change adds documentation for the `distributed_exec` API function
used to execute commands across the data nodes of a distributed
database.

### DIFF
--- a/api.md
+++ b/api.md
@@ -926,7 +926,7 @@ SELECT distributed_exec($$CREATE USER davide WITH PASSWORD 'jw8s0F4'$$, node_lis
 
 ```
 
---
+---
 
 ## drop_chunks() [](drop_chunks)
 
--- a/api.md
+++ b/api.md
@@ -20,6 +20,7 @@
 > - [delete_data_node](#delete_data_node)
 > - [detach_tablespace](#detach_tablespace)
 > - [detach_tablespaces](#detach_tablespaces)
+> - [distributed_exec](#distributed_exec)
 > - [drop_chunks](#drop_chunks)
 > - [drop view (continuous aggregate)](#continuous_aggregate-drop_view)
 > - [first](#first)
@@ -222,7 +223,7 @@ previously created with
 ```sql
 SELECT * FROM attach_data_node('dn3','conditions');
 
-hypertable_id | node_hypertable_id |  node_name  
+hypertable_id | node_hypertable_id |  node_name
 --------------+--------------------+-------------
             5 |                  3 | dn3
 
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -887,12 +888,51 @@ SELECT detach_tablespaces('conditions');
...
+|Name|Description|
+|---|---|
+| `node_list` | An array of data node names to execute the command across. Defaults to all data nodes if not specified. |
+
+#### Sample Usage [](distributed_exec-examples)
+
+Create the role `davide` across all data nodes in a distributed database:
+
+```sql
+SELECT distributed_exec('CREATE USER davide WITH PASSWORD ''jw8s0F4''');
```
feikesteenbergen: Nit, probably a matter of taste, but I like to use the dollar quoting syntax for this kind of stuff
```suggestion
SELECT distributed_exec($sql$CREATE USER davide WITH PASSWORD 'jw8s0F4'$sql$);
```

api.md
```diff
@@ -887,12 +888,51 @@ SELECT detach_tablespaces('conditions');
...
+distributed database. A common use case is creating the roles and
+permissions needed in a distributed database.
+
+Note that the command is _not_ executed on the access node itself.
+
+#### Required Arguments [](distributed_exec-required-arguments)
+
+|Name|Description|
+|---|---|
+| `query` | The command to execute on data nodes. |
```
feikesteenbergen: For me it is unclear if I can chain commands here, can i runthe following for example:
```sql
SELECT distributed_exec($sql$CREATE USER davide; GRANT myapp TO davide;$sql$);
```

api.md
```diff
@@ -887,12 +888,51 @@ SELECT detach_tablespaces('conditions');
 ```
 
 ---
+## distributed_exec() [](distributed_exec)
+
+This command is used on an access node to execute a SQL command across
+the data nodes of a distributed database. This command can be used to,
+e.g., create cluster-wide objects across all instances of a
```
mkindahl: For stylistic reasons, I avoid "e.g." and "i.e." in text and write "for example" and "that is" instead. Easier on the eye.

mkindahl: Think we should reserve "cluster" for "PostgreSQL cluster" and use "distributed database". In this case, you can just remove "cluster-wide" and just write "...create objects on all connected data nodes..." and avoid the extra note below.

api.md
```diff
@@ -887,12 +888,51 @@ SELECT detach_tablespaces('conditions');
...
+distributed database. A common use case is creating the roles and
```
mkindahl: I prefer to not use words "common", "often", and similar subjective words in formal texts since it can give the wrong impression.

api.md
```diff
@@ -887,12 +888,51 @@ SELECT detach_tablespaces('conditions');
...
+permissions needed in a distributed database.
+
+Note that the command is _not_ executed on the access node itself.
```
mkindahl: Not needed if you reformulate the above sentence.

api.md
```diff
@@ -887,12 +888,51 @@ SELECT detach_tablespaces('conditions');
...
+
+|Name|Description|
+|---|---|
+| `query` | The command to execute on data nodes. |
+
+#### Optional Arguments [](distributed_exec-optional-arguments)
+
+|Name|Description|
+|---|---|
+| `node_list` | An array of data node names to execute the command across. Defaults to all data nodes if not specified. |
```
mkindahl: Suggest: "An array of data nodes where the command should be executed."

api.md
solugebefola: cc @mtgentry useful for postgrescheatsheet.com?

erimatnor: @feikesteenbergen Would you mind just `$$`?

solugebefola:
I'll review as soon as current comments are addressed

api.md
erimatnor: This isn't currently supported. Not sure it can be supported either, but might be worth investigating. Will make clear.

api.md
erimatnor: FWIW, `cluster-wide` here refers to the PostgreSQL terminology used in the official docs, like here: https://www.postgresql.org/docs/12/catalog-pg-authid.html. But yeah, the term is very overloaded.

api.md
feikesteenbergen: No, I don't mind just $$

api.md
```diff
@@ -887,12 +888,51 @@ SELECT detach_tablespaces('conditions');
...
+```
+
+Create the role `davide` on two specific data nodes:
+
+```sql
+SELECT distributed_exec($$CREATE USER davide WITH PASSWORD 'jw8s0F4'$$, node_list => '{ "dn1", "dn2" }');
+
+```
+
+--
```
```suggestion
---
```
needs three dashes

api.md
```diff
@@ -887,12 +888,51 @@ SELECT detach_tablespaces('conditions');
...
 
 ## drop_chunks() [](drop_chunks)
 
 Removes data chunks whose time range falls completely before (or after) a
 specified time, operating either across all hypertables or for a specific one.
-Shows a list of the chunks that were dropped in the same style as the 
+Shows a list of the chunks that were dropped in the same style as the
```
```suggestion
Shows a list of the chunks that were dropped, in the same style as the
```

PR #238, opened 2019-10-23
Update the examples in the multinode tutorial

erimatnor:
This simplifies and expands the tutorial for distributed hypertables
in the multinode tutorial.

### DIFF
--- a/tutorials/clustering.md
+++ b/tutorials/clustering.md
@@ -63,37 +63,54 @@ using our informational views:
 SELECT * FROM timescaledb_information.data_node;
 ```
 
-Now that we've created a database and connected a couple data nodes, let's go ahead
-and create a distributed hypertable. With the following commands, we'll create
-a hypertable and then distribute it:
+Now that we've created a database and added a couple data nodes,
+let's go ahead and create a distributed hypertable. With the following
+commands, we'll create a distributed hypertable for collecting
+temperature readings from sensors:
 
 ```sql
-CREATE TABLE distributed (time timestamptz NOT NULL, id integer, value numeric);
+CREATE TABLE conditions (time timestamptz NOT NULL, device integer, temp float);
 
-SELECT create_distributed_hypertable('distributed','time', 'id');
+SELECT create_distributed_hypertable('conditions', 'time', 'device');
 ```
 
-To check how your data is being distributed, let's generate some sample data:
+This will partition the `conditions` table along one time dimension
+(`time`) and one space dimension (`device`). By default, all
+the data nodes are used.
+
+Now, let's generate some sample data in insert it into the
+`conditions` table:
 
 ```sql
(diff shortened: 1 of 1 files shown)


tutorials/clustering.md
```diff
@@ -63,37 +63,54 @@ using our informational views:
 SELECT * FROM timescaledb_information.data_node;
 ```
 
-Now that we've created a database and connected a couple data nodes, let's go ahead
-and create a distributed hypertable. With the following commands, we'll create
-a hypertable and then distribute it:
+Now that we've created a database and connected a couple data nodes,
```
mkindahl: Suggestion: s/connected/added/ This is consistent with the name of `add_data_node`.

tutorials/clustering.md
```diff
@@ -63,37 +63,54 @@ using our informational views:
...
+let's go ahead and create a distributed hypertable. With the following
+commands, we'll create a distributed hypertable for collecting
```
mkindahl: When working with an editor, I was constantly reminded that I should use "you" instead of "we" and "us". :) It would be a lot to change if you did it in this chapter though, so feel free to ignore it.

tutorials/clustering.md
```diff
@@ -63,37 +63,54 @@ using our informational views:
...
+
+The data node view can also show how chunks are distributed across the
+nodes:
+
+```sql
+SELECT * FROM timescaledb_information.data_node;
 ```
 
-You can also check how you data is actually being distributed by running the following:
+You can, of course, query the distributed hypertable as normal to see
```
mkindahl: Remove "of course". It might not be obvious to the reader.

tutorials/clustering.md
erimatnor: Leaving as is for now. But, fwiw, in my experience "we" is more common, at least when you want to give the impression that you (as an author) are going through the steps together with the reader. At least in academic papers this seems to be the norm (have never encountered "you" in that context). Maybe our American friends can chime in here? @WireBaron

tutorials/clustering.md
erimatnor: I'll remove, but would be very surprised if it wasn't obvious to the reader that they could query the data in the table.

tutorials/clustering.md
mkindahl: Yes, it is what I was taught when writing research papers as well, but my editor claimed my writing style was too academic and only what he saw in academic circles.

tutorials/clustering.md
mkindahl: It's more about how you are perceived rather than what the reader knows. :)

tutorials/clustering.md
erimatnor: I guess that fact that we're a bunch of academics should be reflected somewhere :)

PR #242, opened 2019-10-25
Update prometheus tutorial after breaking changes

blagojts:
The changes to the prometheus_postgres_adapter and pg_prometheus
in config flags and the docker tags break this tutorial. This
commit makes it up to date.

### DIFF
--- a/tutorials/prometheus-adapter.md
+++ b/tutorials/prometheus-adapter.md
@@ -201,34 +201,17 @@ use a PostgreSQL docker image that has both Pg_Prometheus and TimescaleDB
 extensions installed (found [here][docker-pg-prom-timescale]).
 
 ```bash
-docker run --network prometheus_timescale_network  --name pg_prometheus -d -p 5432:5432 timescale/pg_prometheus:latest postgres \
+docker run --network prometheus_timescale_network  --name pg_prometheus \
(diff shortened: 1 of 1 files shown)


tutorials/prometheus-adapter.md
```diff
@@ -201,34 +201,17 @@ use a PostgreSQL docker image that has both Pg_Prometheus and TimescaleDB
...
+     -e POSTGRES_PASSWORD=secret -d -p 5432:5432 timescale/pg_prometheus:latest-pg11 postgres \
      -csynchronous_commit=off
 ```
 
 Note that this image inherits from the official PostgreSQL image and so all
 options documented there are applicable to this image as well. This is
 especially important for users that wish to persist data outside of docker
-volumes is the PGDATA environment variable and accompanying volume mount (visit https://hub.docker.com/_/postgres/ for available configuration options)
-
-#### Set the `postgres` user's password
```
blagojts: I wanted to simplify the tut a bit. So I removed this whole 'Set the `postgres` user's password' section with one env var used in docker run

blagojts:
I added the label for 1.0 as the since pg_prometheus works since then. But in the tutorial text the docker images we pull are for version 1.4.2

PR #252, opened 2019-11-01
[fix]Correct additions for v1.5.0

solugebefola:


solugebefola:
Corrections for the 1.5.0 content additions (as seen on docs-staging.timescale.com)

PR #259, opened 2019-11-12
Don't include aggregate columns in group by

davidkohn88:


davidkohn88:
gah. I missed this, it's all versions where contaggs exist, so looked it's 1.3

PR #266, opened 2019-11-26
Add description of data node bootstrapping

### LINKED ISSUE #269: Add description of data node bootstrapping
Add a description for how data node bootstrapping works to the documentation.

mkindahl:
Adding a detailed description of the data node bootstrapping.

Fixes #269

### DIFF
--- a/multinode/bootstrapping.md
+++ b/multinode/bootstrapping.md
@@ -1,0 +1,191 @@
+# Data Node Bootstrapping [](distdb-bootstrap)
+
+Scaling up TimescaleDB is achieved by adding data nodes to an access
+node using the [`add_data_node`][add_data_node] and
+[`attach_data_node`][attach_data_node] functions. In order for a data
+node to be available to the access node it is necessary to ensure that
+it is properly setup to work as a data node. This process is referred
+to as *bootstrapping the data node*, which is done automatically each
+time a data node is added to an access node (unless `bootstrap` is
+false, in which case it just verifies that the data node is properly
+set up).
+
+The steps performed when boostrapping a data node are:
+
+1. Connect to the data node and authenticate the user.
+1. Check that the database is on the data node and create it
+   otherwise.
+2. Check that the TimescaleDB extension is available in the database or
+   create it otherwise.
+3. Set metadata on the data node.
+
+## Authentication during bootstrapping
+
+All connections to the data node are made as the current user on the
+access node. Which requires that:
+
+1. The current user exists on both the data node and access node.
+2. The data node is set up to handle logins using either certificates
+   or password files, as described in the [Data Node Authentication
+   Chapter][distdb-auth].
+
+## Adding data nodes to the access node
+
+A new data node is added to the access node using the
+[`add_data_node`][add_data_node] command.
+
+```sql
+SELECT * FROM add_data_node('data_node_1', 'data1.example.com');
+```
+
+The access node has to ensure that the database on the data node is
+properly configured by either bootstrapping the database or verifying
+that the database exists and is correctly configured. This is
+controlled with the `bootstrap` option, which can be either `true`
+(indicating that the data node should be bootstrapped) or `false`
+(indicating that the data node will only be validated, but no
+bootstrapping will take place).
+
+```sql
+SELECT * FROM add_data_node('data_node_1', 'data1.example.com',
+                            boostrap => false);
+```
+
+The default value for the `bootstrap` option is `true`, which means
+that the data node is bootstrapped when added.
+
+## Setting up the database on the data node
+
+The database to bootstrap on the data node will have the name of the
+current database on the access node by default. For example, if you
+are connected to the `postgres` database on the access node, the
+database name will be `postgres` on the data node as well.
+
+To provision the data node in a different database on the instance use
+the `database` option to the [`add_data_node`][add_data_node] command,
+for example:
+
+```sql
+SELECT * FROM add_data_node('data_node_1', 'data1.example.com',
+                            database => 'my_database');
+```
+
+This is particularly useful if you want to experiment with having data
+nodes and access nodes in the same Postgres instance.
+
+### Bootstrapping the database
+
+If `bootstrap` is `true` (which is the default), the access node will
+first connect to either the `postgres` database or, if the `postgres`
+database does not exist, to the `template0` database on the data node,
+and create a new database:
+
+```sql
+CREATE DATABASE <database>
+   TEMPLATE template0 OWNER <owner>
+   ENCODING <encoding> LC_COLLATE <collation>
+   LC_CTYPE <character type>
+```
+
+Here, the `owner` is the user connected on the access node and the
+`encoding`, `collation`, and `character type` parameters are taken
+from the current database on the access node.
+
+Once the database is created, the access node disconnects and
+continues by connecting to the database and proceeds with
+bootstrapping the extension. If the database already exists on the
+data node, the procedure validates the database and proceeds with
+setting up the extension.
+
+### Validating the database
+
+The database will be validated by checking that the encoding,
+character type, and collation are the same as the current database on
+the access node.
+
+## Setting up the extension
+
+The `timescaledb` extension needs to be installed in the database on
+the data node, so the next step is ensuring that there is a valid
+extension installed in the database by either:
+
+* Creating the extension.
+* Verifying that the extension is compatible with the extension
+  installed on the access node.
+
+### Creating the schema for the extension
+
+To install the extension, a schema is created to store objects that
+the `timescaledb` extension needs. The schema name on the data node is
+set to match the extension schema name on the access node. If you have
+created the `timescaledb` extension in the default manner, `public`
+will be used for the schema on the access node, hence `public` will be
+used on the data node as well.
+
+If the schema name on the access node is `public`, no new schema is
+created on the data node, but otherwise it is expected that the schema
+name does not exist. In the event that the schema name exists (but is
+not `public`), an error will be raised and the bootstrapping will
+abort.
+
+To change the schema used on the access node and data node, you can
+create a new schema on the access node, then create the extension with
+the schema that you want:
+
+```sql
+CREATE SCHEMA magic;
+CREATE EXTENSION timescaledb SCHEMA magic;
+```
+
+When validating the schema, it is necessary that the schema is owned
+by the current user, or is `public`.
+
+>:TIP: Creating the extension in a non-default schema will
+> automatically place all TimescaleDB functions into that
+> schema. Therefore, all function references must be qualified within
+> that schema. For example, to call `create_hypertable` from the
+> previous magic schema:
+> 
+> ```sql
+> SELECT * FROM magic.create_hypertable('conditions', 'time', 3);
+> ```
+
+### Installing the extension
+
+After the schema is created, the extension is installed using the
+`CREATE EXTENSION` command with the `SCHEMA` option to refer to the
+schema from the section above. The schema name for the extension will
+be either `public` or a dedicated schema name.
+
+If the extension already exists, it is validated to ensure that it can
+be used with the access node, then the procedure sets the metadata.
+
+### Validating the extension
+
+Validation of the extension checks that the extension on the data node
+is of a version that is compatible with the version on the access node
+and that the access node has permissions to use the extension.
+
+The extension on the data node is compatible with the extension on the
+access node if the major versions are the same, and the minor version
+on the data node is before or the same as the minor version on the
+access node. The difference in the patch version doesn't affect compatibility. For example:
+
+- Compatible: data node v1.3.1, access node v1.5.1
+- Compatible: data node v1.3.5, access node v1.3.1
+- Incompatible: data node v1.4.1, access node v1.3.1
+- Incompatible: data node v2.0.0, access node v1.3.1
+
+In addition to validating the version, it is necessary to ensure that
+the access node has sufficient privileges on the data node. This
+requires that the connected user is the owner of the extension.
+
+## Set the metadata
+
+Once the extension is set up, the data node is marked as belonging to
+the access node by setting the `dist_id` of the data node to the UUID
+of the access node.
+
+[distdb-auth]: /getting-started/setup/data-node-authentication
+[add_data_node]: /api#add_data_node
+[attach_data_node]: /api#attach_data_node
--- a/page-index/page-index.js
+++ b/page-index/page-index.js
@@ -61,8 +61,8 @@ const pageIndex = [
                         href: "migrate-from-postgresql",
                         to: "/getting-started/migrating-data"
                     }, {
-                  			Title: "Data Node Authentication",
-                  			type: PAGE,
+                  	Title: "Data Node Authentication",
+                  	type: PAGE,
                         href: "data-node-authentication",
                     }
                 ]
@@ -274,6 +274,18 @@ const pageIndex = [
                 type: PAGE,
                 href: "limitations",
                 children: []
+            }
+        ]
+    }, {
+	Title: "Distributed Hypertable",
+        type: REDIRECT,
+        href: "multinode",
+        to: "/multinode/bootstrapping",
+        children: [
+            {
+                Title: "Data Node Bootstrapping",
+                type: PAGE,
+                href: "bootstrapping",
             }
         ]
     }, {


reference/data-node-bootstrapping.md
```diff
@@ -0,0 +1,130 @@
...
+
+When an access node adds a new data node to the distributed database
+it either bootstraps the data node or verifies that the data node is
+properly bootstrapped.
+
+A new data node is added to the access node using the `add_data_node`
+command. The default database to bootstrap on the data node is then
+given by the same name, that is if you are connected to the `my_data`
+database on the access node, it will be `my_data` on the data node as
+well. If you do not want to use the same database on the data node as
```
k-rus: I was confused by the formation, since it is not clear `the same name` as what, and it is unclear that `my_data` is an example.
So here is my suggestion:

> The default database to bootstrap has the same name as the access node database, i.e., the connected database from where `add_data_node` is called. For example, the name of the database is `my_test_db` on the access node and `add_data_node` is called, then a database will be bootstrap with name `my_test_db`. In the case of the database `postgres`, which is common starting point, `postgres` database will be bootstrap on the data node.

reference/data-node-bootstrapping.md
```diff
@@ -0,0 +1,130 @@
...
+1. Ensure that there is a database on the data node.
+2. Ensure that the extension is available in the database.
+3. Set metadata on the data node.
+
+## Setting up the database on the data node
+
+First, the access node has to ensure that there is a properly
+configured database on the data node by either bootstrapping a new
+database, or verifying that it is correctly set up, depending on the
+value of the `bootstrap` option to `add_data_node`.
```
k-rus: Will to be correct to be specific that it's a flag?
```suggestion
value of the `bootstrap` flag to `add_data_node`.
```

reference/data-node-bootstrapping.md
```diff
@@ -0,0 +1,130 @@
...
+
+The database name on the data node is either the name of the database
+on the access node, or the value of the `database` option to
+`add_data_node`.
+
+### Bootstrapping the database
+
+If `bootstrap` is true (which is the default), the access node will
+first connect to either the `postgres` database or, if the `postgres`
+database does not exist, to the `template0` database and create the
```
k-rus:
```suggestion
database does not exist, to the `template0` database on the data node and create new
```

reference/data-node-bootstrapping.md
```diff
@@ -0,0 +1,130 @@
...
+   LC_CTYPE <character type>
+```
+
+Here, the *owner* is the user connected on the access node, the
+*encoding*, *collation*, and *character type* is taken from the
+current database on the access node.
+
+Once the database is created, the access node disconnect and continue
+with bootstrapping the extension. If the database already exists on
+the data node, this step is skipped.
```
k-rus: Which step is skipped? I guess you mean the step of creating the database. Am I correct it means that there will be no error if the database exists and `bootstrap` is `true`?

reference/data-node-bootstrapping.md
```diff
@@ -0,0 +1,130 @@
...
+or verifying that the extension is compatible with the extension
+installed on the access node.
+
+### Bootstrapping the extension
+
+#### Creating the schema for the extension
+
+To install the extension, a schema is created to store the objects
+that the `timescaledb` extension needs. The schema name is decided by
+picking the same schema name as on the access node and using that, so
```
k-rus: The same schema as what? Is it the same as the current schema name during the call `add_data_node`?

reference/data-node-bootstrapping.md
```diff
@@ -0,0 +1,130 @@
...
+if you have created the `timescaledb` extension in the default manner,
+`public` will be used for the schema on the access node, hence
+`public` will be used on the data node as well.
+
+If the schema name for the extension on the access node is `public`,
+no new schema is created, but otherwise it is expected that the schema
```
k-rus: What does it mean `expected`? Will it return error if the schema exists?

reference/data-node-bootstrapping.md
```diff
@@ -0,0 +1,130 @@
...
+> ```
+> SELECT * FROM magic.create_hypertable('conditions', 'time', 3);
+> ```
+
+#### Installing the extension
+
+Once there is a valid schema for the extension, the extension will be
+installed.
+
+[If the extension already exists, it is validated to ensure that it
```
k-rus: Should it be a comment that the correct version of the extension is verified?

reference/data-node-bootstrapping.md
mkindahl: Correct, there will be no error.

reference/data-node-bootstrapping.md
mkindahl: It is not currently verified since the bootstrapping fails if the extension already exists, but I think we should add it if we allow the extension to be pre-installed.

k-rus: You are right. It is checked when a connection is created, if I remember correctly.

reference/data-node-bootstrapping.md
```diff
@@ -0,0 +1,130 @@
+# Data Node Bootstrapping [](distdb-bootstrap)
+
+In order for an access node to be able to use a data node it is
+necessary to ensure that it is properly setup to work as a data
+node. This is called *bootstrapping* the data node.
+
+When an access node adds a new data node to the distributed database
```
solugebefola:
```suggestion
When an access node is adding a new data node to the distributed database
```

reference/data-node-bootstrapping.md
```diff
@@ -0,0 +1,130 @@
...
+command. The default database to bootstrap on the data node is then
+given by the same name, that is if you are connected to the `my_data`
+database on the access node, it will be `my_data` on the data node as
+well. If you do not want to use the same database on the data node as
+on the access node, this can be overridden using the `database` option
+to the `add_data_node` command.
+
+All connections to the data node are made as the current user on the
+access node and it is necessary that the user exists on the data node
+and that it is properly set up to handle password-less logins from the
```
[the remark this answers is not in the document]
solugebefola: Also, this link needs a footer reference

reference/data-node-bootstrapping.md
k-rus: @solugebefola Thanks! I didn't think about compliance with our convention. Just thought it might be good idea to clarify that the values can be only `true` or `false`, which I've seen in docs outside Timescale.

reference/data-node-bootstrapping.md
k-rus: I think it will be good which step is skipped.

reference/data-node-bootstrapping.md
```diff
@@ -0,0 +1,130 @@
+# Data Node Bootstrapping [](distdb-bootstrap)
+
+In order for an access node to be able to use a data node it is
+necessary to ensure that it is properly setup to work as a data
```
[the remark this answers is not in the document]
mkindahl: Done.

reference/data-node-bootstrapping.md
```diff
@@ -0,0 +1,130 @@
...
+
+```
+CREATE SCHEMA magic;
+CREATE EXTENSION timescaledb SCHEMA magic;
+```
+
+When validating the schema, it is necessary that the schema is owned
+by the current user.
+
+> NOTE: This will place all the Timescale functions (for example,
```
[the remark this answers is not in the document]
mkindahl: Done.

reference/data-node-bootstrapping.md
```diff
@@ -0,0 +1,130 @@
...
+
+Here, the *owner* is the user connected on the access node, the
+*encoding*, *collation*, and *character type* is taken from the
+current database on the access node.
+
+Once the database is created, the access node disconnect and continue
+with bootstrapping the extension. If the database already exists on
+the data node, this step is skipped.
+
+[Currently, there is no validation of the database as there is if
```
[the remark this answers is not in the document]
mkindahl: Yes, these comments will be removed before committing once the actual work is done.

reference/data-node-bootstrapping.md
```diff
@@ -0,0 +1,130 @@
...
+
+In order for an access node to be able to use a data node it is
+necessary to ensure that it is properly setup to work as a data
+node. This is called *bootstrapping* the data node.
+
+When an access node adds a new data node to the distributed database
+it either bootstraps the data node or verifies that the data node is
+properly bootstrapped.
+
+A new data node is added to the access node using the `add_data_node`
```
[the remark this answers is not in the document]
mkindahl: Done.

reference/data-node-bootstrapping.md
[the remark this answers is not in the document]
mkindahl: Done.

reference/data-node-bootstrapping.md
mkindahl: Updated.

reference/data-node-bootstrapping.md
mkindahl: Elaborated.

reference/data-node-bootstrapping.md
mkindahl: Elaborated.

reference/data-node-bootstrapping.md
mkindahl: Reworded.

mkindahl:
> You need to modify the `page-index.js` file to add this page

Did that by adding a new chapter/section about multinode.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
+# Data Node Bootstrapping [](distdb-bootstrap)
+
+Scaling up multi-node Timescale is achieved by adding data nodes to
```
k-rus: The product is called TimescaleDB and `multi-node` seems to be tautology to `scaling up`:
```suggestion
Scaling up TimescaleDB is achieved by adding data nodes to
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+the access nodes using the `add_data_node` and `attach_data_node`
```
k-rus:
```suggestion
an access node using the `add_data_node` and `attach_data_node`
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+
+1. The current user exists on both the data node and access node.
+2. The data node is set up to handle logins using either certificates
+   or password files, as described in the [Data Node Authentication
+   Chapter][distdb-auth].
+
+## Adding data nodes to the access node
+
+When an access node is adding a new data node to the distributed
+database it either bootstraps the data node itself or verifies that
```
k-rus: `itself` is meaningless here:
```suggestion
database it either bootstraps the data node or verifies that
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+the data node is properly bootstrapped.
+
+A new data node is added to the access node using the `add_data_node`
+command.
+
+```
+SELECT * FROM add_data_node('data_node_1', 'data1.example.com');
+```
+
+The database to bootstrap on the data node is given by the name of the
```
k-rus: Providing a summary of the bootstrap steps will make this sentence meaningful, since bootstrapping a database on the data node was never mentioned.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+current database on the access node. For example, if you are connected
+to the `my_data` database on the access node, the database name will
+be `my_data` on the data node as well.
+
+If you do not want to use the same database on the data node as on the
```
k-rus: It can be simplified, e.g.,:
> If you want to provide different database name then the access node, use `database` option to the `add_data_node`:

Also I suggest to add that it is specifically useful if the data node is created in the same PostgreSQL cluster.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+access node, this can be overridden using the `database` option to the
+`add_data_node` command.
+
+```
+SELECT * FROM add_data_node('data_node_1', 'data1.example.com',
+                            database => 'other');
+```
+
+The access node has to ensure that there is a properly configured
```
k-rus: As this paragraph is directly after discussing creation of the database, this seems to me confusing:
> there is a properly configured database
May be to formulate as:
> that the database is properly configured

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+database on the data node by either bootstrapping a new database or
```
k-rus: IMHO it is important to mention that the above examples with omitted `bootstrap` option means `true`.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+`bootstrap` option to `add_data_node`. A data node can be added
+without bootstrapping by using the `bootstrap => false` option.
+
+```
+SELECT * FROM add_data_node('data_node_1', 'data1.example.com',
+                            boostrap => false);
+```
+
+If a data node is not bootstrapped, only the validation steps below
+are executed. If `bootstrap` is `true` (which is the default) the
```
k-rus: How are the steps different from `bootstrap` is `false`?

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+verifying that it is correctly set up, depending on the value of the
+`bootstrap` option to `add_data_node`. A data node can be added
+without bootstrapping by using the `bootstrap => false` option.
+
+```
+SELECT * FROM add_data_node('data_node_1', 'data1.example.com',
+                            boostrap => false);
+```
+
+If a data node is not bootstrapped, only the validation steps below
```
k-rus: If a data node is not bootstrapped, it is nothing to validate. Do you mean the case `bootstrap => false`?

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+
+If `bootstrap` is false, the database will be validated by checking
+that the encoding, character type, and collation are the same as the
+current database on the access node.
+
+## Setting up the extension
+
+The `timescaledb` extension needs to be installed in the database on
+the data node, so the next step is ensuring that there is a valid
+extension installed in the database by either: Creating the extension,
```
k-rus:
```suggestion
extension installed in the database by either: creating the extension,
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+data node as well.
+
+If the schema name for the extension on the access node is `public`,
+no new schema is created, but otherwise it is expected that the schema
+name does *not* exist. In the event that the schema name exists (but
+is not `public`), an error will be raised the bootstrapping aborted.
+
+Alternative description:
+> If the schema already exists on the data node the access node will
+> verify that it is correctly configured and proceed with installing
```
k-rus: What do you mean by correctly configured schema?

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+the data node, so the next step is ensuring that there is a valid
+extension installed in the database by either: Creating the extension,
+or verifying that the extension is compatible with the extension
+installed on the access node.
+
+### Creating the schema for the extension
+
+To install the extension, a schema is created to store the objects
+that the `timescaledb` extension needs. The schema name is decided
+looking what schema is used for the extension on the access node and
```
k-rus: If the extension is installed in several schemas on the access node, which one is it going to be used? Is it the same schema where the called `add_data_node` is located? May be provide an example.

E.g.:
```sql
CREATE SCHEMA distributed_db;
CREATE EXTENSION timescaledb SCHEMA distributed_db;
SELECT distributed_db.add_data_node(...);
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+functions. In order for a data node to be available to the access node
+it is necessary to ensure that it is properly setup to work as a data
+node. This process is refereed to as *bootstrapping the data node*.
+
+## Connections to the data node
+
+All connections to the data node are made as the current user on the
+access node. Which requires that:
+
+1. The current user exists on both the data node and access node.
```
k-rus: I believe it is important to mention which privileges the current user should have on the data nodes.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+Proposed changes:
+> If the extension already exists, it is validated to ensure that it
+> can be used with the access node and the procedure proceeds to set
+> the metadata.
+
+### Validating the extension
+
+Validation of the extension checks that the extension on the data node
+is of a version that is compatible with the version on the access node
+and that the access node have permissions to use the extension.
```
k-rus: Do you mean that the user has permission to use the extension?

k-rus: I suggest to remove this introduction, since the explanation is close enough.

multinode/bootstrapping.md
mkindahl: Handled. I had to reorder some paragraphs.

multinode/bootstrapping.md
mkindahl: See the new text.

multinode/bootstrapping.md
mkindahl: See the new text.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,193 @@
...
+and create new database:
+
+```
+CREATE DATABASE <database>
+   TEMPLATE template0 OWNER <owner>
+   ENCODING <encoding> LC_COLLATE <collation>
+   LC_CTYPE <character type>
+```
+
+Here, the *owner* is the user connected on the access node, the
```
solugebefola:
```suggestion
Here, the `owner` is the user connected on the access node, the
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,193 @@
...
+*encoding*, *collation*, and *character type* parameters are taken
```
solugebefola:
```suggestion
`encoding`, `collation`, and `character type` parameters are taken
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,193 @@
...
+from the current database on the access node.
+
+Once the database is created, the access node disconnect and continue
```
solugebefola:
```suggestion
Once the database is created, the access node disconnects and continues
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,193 @@
...
+
+The database will be validated by checking that the encoding,
+character type, and collation are the same as the current database on
+the access node.
+
+## Setting up the extension
+
+The `timescaledb` extension needs to be installed in the database on
+the data node, so the next step is ensuring that there is a valid
+extension installed in the database by either: creating the extension,
```
solugebefola: Split choices into a list?

multinode/bootstrapping.md
```diff
@@ -0,0 +1,193 @@
...
+or verifying that the extension is compatible with the extension
+installed on the access node.
+
+### Creating the schema for the extension
+
+To install the extension, a schema is created to store the objects
+that the `timescaledb` extension needs. The schema name is decided
```
solugebefola:
```suggestion
that the `timescaledb` extension needs. The schema name is decided by
```

multinode/bootstrapping.md
solugebefola:
```suggestion
looking at what schema name is used for the extension on the access node and
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,193 @@
...
+looking what schema is used for the extension on the access node and
+using the same name on the data node. If you have created the
+`timescaledb` extension in the default manner, `public` will be used
+for the schema on the access node, hence `public` will be used on the
+data node as well.
+
+If the schema name for the extension on the access node is `public`,
+no new schema is created, but otherwise it is expected that the schema
+name does *not* exist. In the event that the schema name exists (but
+is not `public`), an error will be raised the bootstrapping aborted.
```
solugebefola:
```suggestion
is not `public`), an error will be raised and the bootstrapping will abort.
```

multinode/bootstrapping.md
solugebefola:
```suggestion
and that the access node has permissions to use the extension.
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Validating the extension
+
+Validation of the extension checks that the extension on the data node
+is of a version that is compatible with the version on the access node
+and that the access node have permissions to use the extension.
+
+The extension on the data node is compatible with the extension on the
+access node if the major versions are the same and the minor version
+on the data node is before or same as the minor version on the access
```
solugebefola:
```suggestion
on the data node is before or the same as the minor version on the access
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,193 @@
...
+node. The patch version is ignored for the purpose of this comparison.
+
+In addition to validating the version, it is necessary to ensure that
+the access node have sufficient privileges on the data node. This is
```
solugebefola:
```suggestion
the access node has sufficient privileges on the data node. This is
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,193 @@
+# Data Node Bootstrapping [](distdb-bootstrap)
+
+Scaling up TimescaleDB is achieved by adding data nodes to an access
+nodes using the [`add_data_node`][add_data_node] and
+[`attach_data_node`][attach_data_node] functions. In order for a data
+node to be available to the access node it is necessary to ensure that
+it is properly setup to work as a data node. This process is refereed
```
WireBaron:
```suggestion
it is properly setup to work as a data node. This process is referred
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,193 @@
...
+A new data node is added to the access node using the
+[`add_data_node`][add_data_node] command.
+
+```
+SELECT * FROM add_data_node('data_node_1', 'data1.example.com');
+```
+
+The access node has to ensure that the database on the data node is
+properly configured by either bootstrapping the database or verifying
+that the database is correctly configured. This is controlled with the
```
WireBaron:
```suggestion
that the database exists and is correctly configured. This is controlled with the
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,193 @@
...
+SELECT * FROM add_data_node('data_node_1', 'data1.example.com',
+                            boostrap => false);
+```
+
+The default value for the `bootstrap` option is `true`, which means
+that the data node is bootstrapped when added.
+
+The steps to boostrap the data node are:
+
+1. Ensure that there is a database on the data node.
```
WireBaron:
```suggestion
1. Create the database on the data node.
```
Same suggestion for the next line.

page-index/page-index.js
```diff
@@ -276,6 +276,17 @@ const pageIndex = [
                 children: []
             }
         ]
+    }, {
+        Title: "Distributed Hypertable",
+        type: PAGE,
+        href: "multinode",
```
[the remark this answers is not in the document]
mkindahl: I think that is what I added? I tried to mimic how the other nodes were structured, but I am a little unsure of what the "href" is relative to and if it refers to a file or a directory.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,193 @@
...
+3. Set metadata on the data node.
+
+## Setting up the database on the data node
+
+The database to bootstrap on the data node is given by the name of the
+current database on the access node. For example, if you are connected
+to the `postgres` database on the access node, the database name will
+be `postgres` on the data node as well.
+
+To place the data node in a different database on the data node use
```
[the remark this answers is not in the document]
mkindahl: Not sure I follow. A data node is represented as a database in a PostgreSQL cluster, so not sure if you can place a database "on" a data node.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,193 @@
...
+the `database` option to the [`add_data_node`][add_data_node] command,
+for example:
+
+```
+SELECT * FROM add_data_node('data_node_1', 'data1.example.com',
+                            database => 'my_database');
+```
+
+This is particularly useful if you want to experiment with having data
+nodes and access nodes in the same Postgres cluster.
```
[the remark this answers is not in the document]
mkindahl: Instance is better.

multinode/bootstrapping.md
mkindahl: These are the privileges required for connecting to the data node, but more are needed to actually bootstrap the data node.

multinode/bootstrapping.md
k-rus: @mkindahl BUMP

multinode/bootstrapping.md
WireBaron: No, that was a bad suggestion on my part.  I was thinking of the `if_not_exists` functionality.

page-index/page-index.js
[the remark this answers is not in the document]
mkindahl: Ah. Fixed that.

multinode/bootstrapping.md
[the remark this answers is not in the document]
mkindahl: Maybe "provision", but I do not think that "convert" is the proper name. A data node is a database with the extension loaded, a schema, and some metadata that is set for the data node, so it is still a database but also a data node.

Rewrote it, PTAL.

multinode/bootstrapping.md
mkindahl: It depends on the implementation I am not sure we need to describe it in detail here.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,196 @@
...
+Proposed changes:
+> If the extension already exists, it is validated to ensure that it
+> can be used with the access node and the procedure proceeds to set
+> the metadata.
+
+### Validating the extension
+
+Validation of the extension checks that the extension on the data node
+is of a version that is compatible with the version on the access node
+and that the access node has permissions to use the extension.
```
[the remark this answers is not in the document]
mkindahl: Done.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,196 @@
...
+
+The extension on the data node is compatible with the extension on the
+access node if the major versions are the same and the minor version
+on the data node is before or the same as the minor version on the access
+node. The patch version is ignored for the purpose of this comparison.
```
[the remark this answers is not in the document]
mkindahl: Added a few examples.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,196 @@
...
+checked by ensuring that the connected user is the owner of the
+extension.
+
+## Set the metadata
+
+[ Need to mention that we do some verification here if there is no
+bootstrapping and also checks that are done on the access node to make
+sure that it does not have the role of a data node. /Mats]
+
+Once the extension is set up, the data node is marked as belonging to
```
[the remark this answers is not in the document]
mkindahl: I wonder if "flagged" is the right word. The =dist_uuid= field in the =metadata= table is set to the =uuid= of the access node. Wouldn't that be more similar to "marked", "imbued", "colored", or something along those lines?

multinode/bootstrapping.md
```diff
@@ -0,0 +1,201 @@
...
+access node. Which requires that:
+
+1. The current user exists on both the data node and access node.
+2. The data node is set up to handle logins using either certificates
+   or password files, as described in the [Data Node Authentication
+   Chapter][distdb-auth].
+
+## Adding data nodes to the access node
+
+While an access node is adding a new data node to the distributed
```
erimatnor:
```suggestion
When an access node adds a new data node to the distributed
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,201 @@
...
+A new data node is added to the access node using the
+[`add_data_node`][add_data_node] command.
+
+```
+SELECT * FROM add_data_node('data_node_1', 'data1.example.com');
+```
+
+The access node has to ensure that the database on the data node is
+properly configured by either bootstrapping the database or verifying
+that the database exists and is correctly configured. This is controlled with the
```
erimatnor:
```suggestion
that the database exists and is correctly configured. The behavior is controlled with the
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,201 @@
+# Data Node Bootstrapping [](distdb-bootstrap)
+
+Scaling up TimescaleDB is achieved by adding data nodes to an access
+node using the [`add_data_node`][add_data_node] and
+[`attach_data_node`][attach_data_node] functions. In order for a data
+node to be available to the access node it is necessary to ensure that
+it is properly setup to work as a data node. This process is referred
+to as *bootstrapping the data node*.
+
+## Connections to the data node
```
erimatnor:
```suggestion
## Authentication during bootstrapping
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,201 @@
...
+SELECT * FROM add_data_node('data_node_1', 'data1.example.com',
+                            boostrap => false);
+```
+
+The default value for the `bootstrap` option is `true`, which means
+that the data node is bootstrapped when added.
+
+The steps to boostrap the data node are:
+
+1. Ensure that there is a database on the data node.
```
erimatnor: I think it would be better if this list is upfront instead of last in this section (I think this might also be what @k-rus  asked for). It is generally easier to first have a big picture overview of the process and then drill down into the details. Then the reader knows better what to expect.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,201 @@
...
+Validation of the extension checks that the extension on the data node
+is of a version that is compatible with the version on the access node
+and that the access node has permissions to use the extension.
+
+The extension on the data node is compatible with the extension on the
+access node if the major versions are the same and the minor version
+on the data node is before or the same as the minor version on the access
+node. For example:
+
+- Compatible: data node v1.3.1, access node v1.5.1
```
erimatnor: The policy for extension compatibility seems a lot more important than the process of bootstrapping in general. Can we try to emphasize the policy and reduce the process stuff?  Another question is if this validation is specific to bootstrapping? Shouldn't it happen on every connection or at least also during upgrades (i.e., when the data node already exists)? I am asking because the information here seems important outside the context of bootstrapping (e.g., as part of a section on node compatibility or upgrading). Now this important information is buried in a lot of bootstrapping implementation detail.

Generally, I think it is best to focus on documenting _what_ we do and _why_, instead of _how_ we do it.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Bootstrapping the database
+
+If `bootstrap` is true (which is the default), the access node will
+first connect to either the `postgres` database or, if the `postgres`
+database does not exist, to the `template0` database on the data node,
+and create a new database:
+
+```
+CREATE DATABASE <database>
```
erimatnor: While I think it is great to document things, I am wondering what purpose these detailed bootstrapping docs serve (thinking mostly about stuff from this section and onward)? Do we intend for people to run through these steps manually?

I am worried that this will confuse people more than it helps given that this is automated by the `add_data_node` command. Further, these docs will easily get outdated as soon as we make a small change to the bootstrapping processes. 

If this is about making people better understand what is going on internally, then I think we should rather improve error messages and notices.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,201 @@
...
+- Compatible: data node v1.3.5, access node v1.3.1
+- Incompatible: data node 1.4.1, access node v1.3.1
+- Incompatible: data node 2.0.0, access node v1.3.1
+
+In addition to validating the version, it is necessary to ensure that
+the access node has sufficient privileges on the data node. This is
+checked by ensuring that the connected user is the owner of the
+extension.
+
+## Set the metadata
```
erimatnor: This seems very much like implementation detail. Do we need this?

multinode/bootstrapping.md
```diff
@@ -0,0 +1,201 @@
...
+
+[ Need to mention that we do some verification here if there is no
```
erimatnor: Why do we need to mention verification? Ideally, this is not something a user should have to worry about unless we've done something wrong. And, in case of failures, we should focus on raising insightful errors.

multinode/bootstrapping.md
mkindahl: I've gotten questions about how it works several times when on support. Also, if this procedure changes frequently, then we have done a bad job at the design, so that do not worry me too much. Once we release this, it will be stable only because it has to be stable.

multinode/bootstrapping.md
mkindahl: Yes, the verification happens each time a connection is established, so it is not restricted to only the bootstrapping.

I think this is a description of what we do and why? Not sure what you think should be written differently.

multinode/bootstrapping.md
mkindahl: It is a very small detail so I do not see it as a big problem to add. Also, we cannot really change this between versions, so it is very stable.

For need: there are DBAs that need to repair or understand a problem. This is really where the documentation is really needed.

multinode/bootstrapping.md
mkindahl: I do not think that it is a matter of either documentation or error messages. IMHO, providing documentation that is sufficient enough for debugging it is part of the solution.

multinode/bootstrapping.md
erimatnor: Sounds great that we have such interest and it would be interesting to deep-dive into that conversation. I've been following at least the `multinode-beta` channel but haven't seen any questions on the bootstrapping process so far (apart from the Docker extension issue, which is since fixed). Do you have a reference to those questions about the bootstrapping process on support Slack? 

Taking a step back though, the reason I am pushing back a bit is that I generally have the opinion that any need for detailed documentation is a signal that you've failed to provide a good interface. Most people never read the documentation and it easily gets out-of-date, so I am simply trying to manage trade-offs here. I'd rather have engineers spending time on code than docs, instead making sure that the interfaces are clear and that the code works in such a way that it makes documentation almost redundant.

Anyway, if you think this is important I am fine with merging this. I am just trying to look at the bigger docs picture than nitpicking spelling errors, grammar, etc.

multinode/bootstrapping.md
erimatnor: Just saying that extension compatibility seems more important than other minute details here which warrants lifting it up from being buried in these bootstrapping details. And, further, node validation doesn't seem to be specific to bootstrapping. So, my comment is about structure rather than content.

In other words, these two things (extension version compatibility and validation) are connected. I could imagine, e.g., a separate (more front-and-center) doc/chapter on requirements/prerequisites/policy related to node interactions that are relevant to both upgrades and bootstrapping that doesn't require digging into the details of the bootstrapping process. If I imagine myself wearing the hat of a cluster administrator I would personally prefer a concise chapter specifically on node requirements and compatibility. This is crucial for me to conceptualize how to organize and maintain a cluster, without having to care about the details of the bootstrapping process.

So, what I am proposing is organizing this differently to aid readers by separating key information from minute details. Concretely we'd separate requirements/prerequisites/policy from bootstrapping details. For example, if you write "Validation of the extension checks that the extension on the data node
is of a version that is compatible with the version on the access node" you need not list the exact policy for compatibility, including examples here. Just reference another chapter that clearly and concisely deals with that, and which we can easily index in the TOC and reference from other places in the doc (e.g., upgrade docs).

But maybe this is part of a larger docs conversation that is better suited in another forum? I just worry that this will never happen if we just let the docs grow organically without thinking about the larger organization and structure.

multinode/bootstrapping.md
erimatnor: Ok, I buy that.

multinode/bootstrapping.md
erimatnor: Just worried that there is no clear point where we draw the line and this is certainly an increased maintenance burden. I'd rather have basic docs that cover 80-90% (or whatever a reasonable number is) and than add details as needed when we know what things trip people up. Not saying this particular detail will never be useful, but we need to put our resources on things that gives us most bang-for-the-buck and covering all bases now is going to be a massive undertaking.

multinode/bootstrapping.md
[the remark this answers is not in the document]
solugebefola:
```suggestion
Once the extension is set up, the data node is assigned to
```
Maybe this is clearer?

multinode/bootstrapping.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Bootstrapping the database
+
+If `bootstrap` is true (which is the default), the access node will
+first connect to either the `postgres` database or, if the `postgres`
+database does not exist, to the `template0` database on the data node,
+and create a new database:
+
+```
+CREATE DATABASE <database>
```
solugebefola:
```suggestion
CREATE DATABASE < database >
```
This is necessary due to a quirk of our markdown parser that treats `<database>` as a potential React JSX element, even within code blocks.  Alternatively, you could use symbols other than `<>`

multinode/bootstrapping.md
```diff
@@ -0,0 +1,189 @@
...
+   TEMPLATE template0 OWNER <owner>
```
solugebefola:
```suggestion
   TEMPLATE template0 OWNER < owner >
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,189 @@
...
+   ENCODING <encoding> LC_COLLATE <collation>
```
solugebefola:
```suggestion
   ENCODING <encoding> LC_COLLATE < collation >
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,189 @@
...
+   LC_CTYPE <character type>
```
solugebefola:
```suggestion
   LC_CTYPE < character type >
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,189 @@
...
+the access node.
+
+## Setting up the extension
+
+The `timescaledb` extension needs to be installed in the database on
+the data node, so the next step is ensuring that there is a valid
+extension installed in the database by either:
+
+* creating the extension, or
+
```
solugebefola: Why this empty line?  Looks weird when rendered:

Screen Shot 2019-12-19 at 11 39 36 AM

multinode/bootstrapping.md
```diff
@@ -0,0 +1,189 @@
...
+CREATE SCHEMA magic;
+CREATE EXTENSION timescaledb SCHEMA magic;
+```
+
+When validating the schema, it is necessary that the schema is owned
+by the current user, or is `public`.
+
+>:TIP: This will place all the Timescale functions (for example,
+> `create_hypertable` in the `magic` schema), so you need to qualify
+> all function references with `magic`. For example:
```
solugebefola: > Creating the extension in a non-default schema will automatically place all TimescaleDB functions into that schema.  Therefore, all function references must be qualified within that schema.  For example, to call `create_hypertable` from the previous `magic` schema:

multinode/bootstrapping.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Installing the extension
+
+After the schema is created, the extension is installed using the
+`CREATE EXTENSION` command with the `SCHEMA` option to refer to the
+schema from the section above. The schema name for the extension will
+be either `public` or a dedicated schema name.
+
+If the extension already exists, it is validated to ensure that it
+can be used with the access node and the procedure proceeds to set
```
solugebefola:
```suggestion
can be used with the access node and the procedure then sets
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,189 @@
...
+the metadata.
+
+### Validating the extension
+
+Validation of the extension checks that the extension on the data node
+is of a version that is compatible with the version on the access node
```
solugebefola:
```suggestion
is of a version that is compatible with the version on the access node,
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,189 @@
...
+and that the access node has permissions to use the extension.
+
+The extension on the data node is compatible with the extension on the
+access node if the major versions are the same and the minor version
```
solugebefola:
```suggestion
access node if the major versions are the same, and the minor version
```

solugebefola:
I forgot to add: the page-index needs to be modified:  The entry bootstrapping needs to be set to this:
```
Title: "Distributed Hypertable",
        type: REDIRECT,
        href: "multinode",
        to: "/multinode/bootstrapping",
        children: [
            {
                Title: "Data Node Bootstrapping",
                type: PAGE,
                href: "bootstrapping",
            }
        ]
```

multinode/bootstrapping.md
mkindahl: What is used elsewhere? I have no strong opinion on what to use, but this is intended to be a [replacable](https://tdg.docbook.org/tdg/4.5/replaceable.html)

multinode/bootstrapping.md
mkindahl: Tried to reformulate it. PTAL.

multinode/bootstrapping.md
mkindahl: I think it is good to make the information available to the user, but I agree that we do need to look over the high-level structure. I think we can move the information as a separate PR for creating a description about validation.

multinode/bootstrapping.md
mkindahl: Yes, I agree that there is a possibility that the documentation will become out-of-date as we move forward, but:

- I do not think it is an argument against providing it both for ourselves and for the users.
- Documenting our intentions will allow users to provide feedback, which will make the product better.
- As we move forward, the documentation will be stable because changes cannot be arbitrarily made to the code and we need to keep these things stable to allow upgrades.

multinode/bootstrapping.md
mkindahl: Switched to use `<i>` tags, not sure if that is supported, but I see it in other parts of the document.

multinode/bootstrapping.md
mkindahl: Done.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+node to be available to the access node it is necessary to ensure that
+it is properly setup to work as a data node. This process is referred
+to as *bootstrapping the data node*, which is done automatically each
+time a data node is added to an access node (unless `bootstrap` is
+false, in which case it just verifies that the data node is properly
+set up).
+
+The steps performed when boostrapping a data node are:
+
+1. Connect to the data node and authenticating the user.
```
k-rus:
```suggestion
1. Connect to the data node and authenticate the user.
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+1. Check that there is a database on the data node and create it
```
k-rus:
```suggestion
1. Check that there is the database on the data node and create it
```
If I don't mistake, the database should have a specific name, not an any database, right?

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+   otherwise.
+2. Check that the Timescale extension is available in the database or
```
k-rus:
```suggestion
2. Check that the TimescaleDB extension is available in the database or
```
There is no extension with name `Timescale`.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+                            boostrap => false);
+```
+
+The default value for the `bootstrap` option is `true`, which means
+that the data node is bootstrapped when added.
+
+## Setting up the database on the data node
+
+The database to bootstrap on the data node is given by the name of the
+current database on the access node. For example, if you are connected
```
k-rus: I don't understand this sentence. Is the database given or is the name fo the current database? I think it is either given or it uses the current database by default. Suggestion:
```suggestion
The database to bootstrap on the data node will have the name of the
  current database on the access node by default. For example, if you are connected
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+nodes and access nodes in the same Postgres instance.
+
+### Bootstrapping the database
+
+If `bootstrap` is true (which is the default), the access node will
+first connect to either the `postgres` database or, if the `postgres`
+database does not exist, to the `template0` database on the data node,
+and create a new database:
+
+```
```
k-rus:
```suggestion
```sql
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+   ENCODING <encoding> LC_COLLATE <i>collation</i>
+   LC_CTYPE <i>character type</i>
+```
+
+Here, the `owner` is the user connected on the access node and the
+`encoding`, `collation`, and `character type` parameters are taken
+from the current database on the access node.
+
+Once the database is created, the access node disconnects and continues
+with bootstrapping the extension. If the database already exists on
```
k-rus: To bootstrap extension the access node needs to connect. Since the formulation is detailed to disconnect, it should include connect too.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+for the schema on the access node, hence `public` will be used on the
+data node as well.
+
+If the schema name for the extension on the access node is `public`,
+no new schema is created, but otherwise it is expected that the schema
+name does *not* exist. In the event that the schema name exists (but
+is not `public`), an error will be raised and the bootstrapping will abort.
+
+To change the schema used on the access node and data node, you can
+create a new schema, then create the extension with the schema that
```
k-rus:
```suggestion
create a new schema on the access node, then create the extension with the schema that
```
It is easy to confuse otherwise with the data node.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+you want:
+
+```
```
k-rus:
```suggestion
```sql
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+When validating the schema, it is necessary that the schema is owned
+by the current user, or is `public`.
+
+>:TIP: Creating the extension in a non-default schema will
+> automatically place all TimescaleDB functions into that
+> schema. Therefore, all function references must be qualified within
+> that schema. For example, to call `create_hypertable` from the
+> previous magic schema:
+> 
+> ```
```
k-rus:
```suggestion
> ```sql
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+### Validating the extension
+
+Validation of the extension checks that the extension on the data node
+is of a version that is compatible with the version on the access node
+and that the access node has permissions to use the extension.
+
+The extension on the data node is compatible with the extension on the
+access node if the major versions are the same, and the minor version
+on the data node is before or the same as the minor version on the
+access node. For example:
```
k-rus:
```suggestion
access node. The difference in the patch version doesn't affect compatibility. For example:
```
The example demonstrates that patch version can be any, so it is good to mention.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+the data node, so the next step is ensuring that there is a valid
+extension installed in the database by either:
+
+* Creating the extension.
+* Verifying that the extension is compatible with the extension
+  installed on the access node.
+
+### Creating the schema for the extension
+
+To install the extension, a schema is created to store the objects
```
solugebefola:
```suggestion
To install the extension, a schema is created to store objects
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+that the `timescaledb` extension needs. The schema name is decided by
+looking at what schema name is used for the extension on the access node and
+using the same name on the data node. If you have created the
```
solugebefola: Change to:
> The schema name on the data node is set to match the extension schema name on the access node

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+`timescaledb` extension in the default manner, `public` will be used
+for the schema on the access node, hence `public` will be used on the
+data node as well.
+
+If the schema name for the extension on the access node is `public`,
```
solugebefola:
```suggestion
If the schema name on the access node is `public`,
```
'for the extension' is implied, since it was discussed in the previous paragraph

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+no new schema is created, but otherwise it is expected that the schema
+name does *not* exist. In the event that the schema name exists (but
```
solugebefola: Clarification...do you mean:
> no new schema is created _on the data node_, but otherwise it is expected that the schema name does *not* exist

It's not clear, since you are discussing the access node earlier in the sentence.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Installing the extension
+
+After the schema is created, the extension is installed using the
+`CREATE EXTENSION` command with the `SCHEMA` option to refer to the
+schema from the section above. The schema name for the extension will
+be either `public` or a dedicated schema name.
+
+If the extension already exists, it is validated to ensure that it can
+be used with the access node and the procedure then sets the metadata.
```
solugebefola:
```suggestion
be used with the access node, then the procedure sets the metadata.
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+and that the access node has permissions to use the extension.
+
+The extension on the data node is compatible with the extension on the
+access node if the major versions are the same, and the minor version
+on the data node is before or the same as the minor version on the
+access node. For example:
+
+- Compatible: data node v1.3.1, access node v1.5.1
+- Compatible: data node v1.3.5, access node v1.3.1
+- Incompatible: data node 1.4.1, access node v1.3.1
```
solugebefola:
```suggestion
- Incompatible: data node v1.4.1, access node v1.3.1
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+- Incompatible: data node 2.0.0, access node v1.3.1
```
solugebefola:
```suggestion
- Incompatible: data node v2.0.0, access node v1.3.1
```

solugebefola: It's actually clearer if you remove the `v`s from all of the version numbers

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+the access node has sufficient privileges on the data node. This
+requires that the connected user is the owner of the extension.
+
+## Set the metadata
+
+Once the extension is set up, the data node is marked as belonging to
+the access node by setting the `dist_id` of the data node to the UUID
+of the access node.
+
+[distdb-auth]: /getting-started/data-node-authentication
```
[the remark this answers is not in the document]
solugebefola: I realized that your PR from a while back added this page, but didn't modify the `page-index.js` file to point to it.  You need to add this location to the page index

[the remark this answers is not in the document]
mkindahl: Not sure how to add it. I see it already in `page-index.js`: it is under "Setting Up TimescaleDB"

multinode/bootstrapping.md
mkindahl: Fixed.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,190 @@
...
+3. Set metadata on the data node.
+
+## Authentication during bootstrapping
+
+All connections to the data node are made as the current user on the
+access node. Which requires that:
+
+1. The current user exists on both the data node and access node.
+2. The data node is set up to handle logins using either certificates
+   or password files, as described in the [Data Node Authentication
```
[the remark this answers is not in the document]
mkindahl: It exists.

multinode/bootstrapping.md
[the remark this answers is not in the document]
solugebefola: Ah, I see the problem.  The link needs to be `/getting-started/setup/data-node-authentication`

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+   ENCODING <encoding> LC_COLLATE <i>collation</i>
+   LC_CTYPE <i>character type</i>
+```
+
+Here, the `owner` is the user connected on the access node and the
+`encoding`, `collation`, and `character type` parameters are taken
+from the current database on the access node.
+
+Once the database is created, the access node disconnects and
+continues by connecting to the database and proceed with bootstrapping
```
solugebefola:
```suggestion
continues by connecting to the database and proceeds with bootstrapping
```
Sorry, missed this before

multinode/bootstrapping.md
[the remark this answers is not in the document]
mkindahl: Fixed.

multinode/bootstrapping.md
mkindahl: Fixed.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+2. The data node is set up to handle logins using either certificates
+   or password files, as described in the [Data Node Authentication
+   Chapter][distdb-auth].
+
+## Adding data nodes to the access node
+
+A new data node is added to the access node using the
+[`add_data_node`][add_data_node] command.
+
+```
```
k-rus:
```suggestion
```sql
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+
+The access node has to ensure that the database on the data node is
+properly configured by either bootstrapping the database or verifying
+that the database exists and is correctly configured. This is
+controlled with the `bootstrap` option, which can be either `true`
+(indicating that the data node should be bootstrapped) or `false`
+(indicating that the data node will only be validated, but no
+bootstrapping will take place).
+
+```
```
k-rus:
```suggestion
```sql
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+The database to bootstrap on the data node will have the name of the
+current database on the access node by default. For example, if you
+are connected to the `postgres` database on the access node, the
+database name will be `postgres` on the data node as well.
+
+To provision the data node in a different database on the instance use
+the `database` option to the [`add_data_node`][add_data_node] command,
+for example:
+
+```
```
k-rus:
```suggestion
```sql
```

page-index/page-index.js
```diff
@@ -61,9 +61,9 @@ const pageIndex = [
                         href: "migrate-from-postgresql",
                         to: "/getting-started/migrating-data"
                     }, {
-                  			Title: "Data Node Authentication",
-                  			type: PAGE,
-                        href: "data-node-authentication",
+                  	Title: "Data Node Authentication",
+                  	type: PAGE,
+                        href: "/getting-started/setup/data-node-authentication",
```
solugebefola: Sorry, there must have been a miscommunication.  This entry didn't need to change, the link in the `bootstrapping.md` file needed to change to `/getting-started/setup/data-node-authentication`

solugebefola:
```suggestion
                        href: "data-node-authentication",
```

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+the access node has sufficient privileges on the data node. This
+requires that the connected user is the owner of the extension.
+
+## Set the metadata
+
+Once the extension is set up, the data node is marked as belonging to
+the access node by setting the `dist_id` of the data node to the UUID
+of the access node.
+
+[distdb-auth]: /getting-started/data-node-authentication
```
solugebefola:
```suggestion
[distdb-auth]: /getting-started/setup/data-node-authentication
```
This is the link that needed to change.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+
+### Bootstrapping the database
+
+If `bootstrap` is `true` (which is the default), the access node will
+first connect to either the `postgres` database or, if the `postgres`
+database does not exist, to the `template0` database on the data node,
+and create a new database:
+
+```sql
+CREATE DATABASE <i>database</i>
```
solugebefola: These \<i\> element tags are not functioning how you anticipated?
Screen Shot 2020-01-28 at 2 24 16 PM

Better to just put a space between '\<' and the text, i.e. '< database >'

mkindahl: Ok. Updated.

multinode/bootstrapping.md
```diff
@@ -0,0 +1,191 @@
...
+
+### Bootstrapping the database
+
+If `bootstrap` is `true` (which is the default), the access node will
+first connect to either the `postgres` database or, if the `postgres`
+database does not exist, to the `template0` database on the data node,
+and create a new database:
+
+```sql
+CREATE DATABASE < database >
```
solugebefola: OK, this is going to drive you crazy, but the brackets with the spaces still looked odd, so I did some experimentation.  As long as the code language is set as `sql` then the brackets with no spaces _does_ work.  I assume that when I was reviewing it before, the `sql` hadn't been added yet (when I took it out, the same problem cropped up when there were no spaces).  Some quirk with the code highlighting that I'll have to track down.  Anyway, if you want to keep as is, that's fine.  Otherwise you can remove the spaces as long as you keep the code block with a language attached.

mkindahl: :D

Ok. I'll remove the spaces then and submit. If anything pops up, we can fix it separately. :)

PR #274, opened 2020-01-14
Update Telegraf version to 1.13.0

blagojts:
Update the Telegraf tutorial text to use the new builds


svenklemm:
You only need to add label for the earliest version this applies to

Issue #275, opened 2020-01-14
Document alter replication factor functionality in multinode


k-rus:
Document alter replication factor after it is implemented.

k-rus:
Implemented definitions:
```sql
CREATE OR REPLACE FUNCTION  set_replication_factor(
    main_table              REGCLASS,
    replication_factor      INTEGER
) RETURNS VOID 
AS '@MODULE_PATHNAME@', 'ts_hypertable_set_replication_factor' LANGUAGE C VOLATILE;
```
Test with examples are in `dist_hypertable.out`.

PR #281, opened 2020-01-22
Document set_replication_factor

k-rus:
Adds documentation of multinode function set_replication_factor.

Closes https://github.com/timescale/docs.timescale.com-content/issues/275

### DIFF
--- a/api.md
+++ b/api.md
@@ -1235,11 +1235,11 @@ SELECT set_number_partitions('conditions', 2, 'device_id');
 
 ## set_replication_factor() [](set_replication_factor)
 Sets the replication factor of a distributed hypertable to the given value. 
-The setting will affect only new chunks. 
-Newly created chunks of the hypertable will be replicated to different data nodes,
-so the amount of replicas is the same as the replication factor.
+Changing the replication factor does not affect the number of replicas for existing chunks.
+Chunks created after changing the replication factor will be replicated 
+in accordance with new value of the replication factor.
 
-If existing chunks have less replicas than the given replication factor,
+If existing chunks have less replicas than new value of the replication factor,
 the function will print a warning.
 
 #### Required Arguments [](set_replication_factor-required-arguments)
--- a/api.md
+++ b/api.md
@@ -1246,8 +1246,8 @@ the function will print a warning.
 
 |Name|Description|
 |---|---|
-| `main_table` | Identifier of the distributed hypertable to update the replication factor for.|
-| `replication_factor` | The new value of the replication factor. Must be greater than 0.|
+| `main_table` | (REGCLASS) Identifier of the distributed hypertable to update the replication factor for.|
+| `replication_factor` | (INTEGER) The new value of the replication factor. Must be greater than 0.|
 
 #### Sample Usage [](set_replication_factor-examples)
 
--- a/api.md
+++ b/api.md
@@ -1237,7 +1237,9 @@ SELECT set_number_partitions('conditions', 2, 'device_id');
 Sets the replication factor of a distributed hypertable to the given value. 
 Changing the replication factor does not affect the number of replicas for existing chunks.
 Chunks created after changing the replication factor will be replicated 
-in accordance with new value of the replication factor.
+in accordance with new value of the replication factor. If the replication factor cannot be
+satisfied, since the amount of attached data nodes is less than new replication factor, 
+the command aborts with an error.
 
 If existing chunks have less replicas than new value of the replication factor,
 the function will print a warning.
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -1229,6 +1230,39 @@ For a table with more than one space dimension:
...
 ```
 
+
+---
+
+## set_replication_factor() [](set_replication_factor)
+Sets the replication factor of a distributed hypertable to the given value. 
+The setting will affect only new chunks. 
+Newly created chunks of the hypertable will be replicated to different data nodes,
+so the amount of replicas is the same as the replication factor.
```
mkindahl: Slightly unclear. Suggestion:

> Changing the replication factor will not change the number of replicas for existing chunks, but any chunks created after changing the replication factor will use the new value of the replication factor.

api.md
```diff
@@ -1229,6 +1230,39 @@ For a table with more than one space dimension:
...
+
+If existing chunks have less replicas than the given replication factor,
+the function will print a warning.
+
+#### Required Arguments [](set_replication_factor-required-arguments)
+
+|Name|Description|
+|---|---|
+| `main_table` | Identifier of the distributed hypertable to update the replication factor for.|
```
mkindahl: Why "main" table? How does it differ from a non-"main" table? I think you should just say "hypertable". I do not think you need to talk about the "identifier". Suggestion:

> Hypertable to change replication factor for.

k-rus: > Why "main" table? How does it differ from a non-"main" table? I think you should just say "hypertable". I do not think you need to talk about the "identifier". Suggestion:
> 
> > Hypertable to change replication factor for.

This is the actual name of the argument in the code. I agree that it is not meaningful enough name, however it is not an uncommon case (see [set_number_partitions](https://docs.timescale.com/clustering/api#set_number_partitions), which I used for inspiration). IMHO this name should consistent with the name in the code and it is not worth to change the code just for the argument name.

erimatnor: Legacy name is a bit unfortunate, but I believe it comes from `create_hypertable()` where naming the parameter `hypertable` doesn't make sense since it is not yet a hypertable at that point. Would probably be better to just name it `table` in all functions but should probably keep legacy name for consistency.

Fortunately, users will probably never reference this parameter explicitly since it is the first one.

api.md
k-rus: Thanks for noting and suggestion. I made an improvement, which I hope helps.

api.md
k-rus: If the implementation will be changed, it will be good opportunity to address this comment. It will be different PR the most likely.

api.md
```diff
@@ -1229,6 +1230,39 @@ For a table with more than one space dimension:
...
+
+If existing chunks have less replicas than new value of the replication factor,
+the function will print a warning.
+
+#### Required Arguments [](set_replication_factor-required-arguments)
+
+|Name|Description|
+|---|---|
+| `main_table` | Identifier of the distributed hypertable to update the replication factor for.|
+| `replication_factor` | The new value of the replication factor. Must be greater than 0.|
```
solugebefola: Put the data type in parentheses before the description. I.e. 
> (INTEGER) The new value...

solugebefola: Also, note any default values

k-rus: > Put the data type in parentheses before the description. I.e.
> 
> > (INTEGER) The new value...

Done

k-rus:
> Looks good. Quick question, have we decided that it should be an WARNING message instead of error? Because apparently the other PR which checks replication factor for exceeding data nodes emits error in a similar case

@pmwkaa This PR was covering implementation, which didn't include this check. The warning is created for the case if the current chunks are underreplicated.

Now I updated this PR with the proposed error implementation, so it includes the error description, example and error section. cc @mkindahl @erimatnor

api.md
```diff
@@ -1229,6 +1230,59 @@ For a table with more than one space dimension:
...
+
+If existing chunks have less replicas than new value of the replication factor,
+the function will print a warning.
+
+#### Required Arguments [](set_replication_factor-required-arguments)
+
+|Name|Description|
+|---|---|
+| `hypertable` | (REGCLASS) Identifier of the distributed hypertable to update the replication factor for.|
+| `replication_factor` | (INTEGER) The new value of the replication factor. Must be greater than 0,
```
solugebefola: Normally adding a newline here would be good, to keep the file properly readable.  Here though, it disrupts the table formatting:
Screen Shot 2020-01-23 at 9 49 54 AM

api.md
```diff
@@ -1229,6 +1230,59 @@ For a table with more than one space dimension:
...
+and smaller than or equal to the number of attached data nodes.|
+
+#### Errors
+
+An error will be given if:
+- `hypertable` is not a distributed hypertable.
+- `replication_factor` is less than `1`, which cannot be set on a distributed hypertable.
+- `replication_factor` is bigger than the number of attached data ndoes.
+
+If the bigger replication factor is desired, it is necessary to attach more data nodes
```
```suggestion
If a bigger replication factor is desired, it is necessary to attach more data nodes 
```

api.md
```diff
@@ -1229,6 +1230,59 @@ For a table with more than one space dimension:
...
+SELECT set_replication_factor('conditions', 2);
+```
+
+Example of the warning if any existing chunk of the distributed hypertable has less than 2 replicas:
+```
+WARNING:  hypertable "conditions" is under-replicated
+DETAIL:  Some chunks have less than 2 replicas.
+```
+
+Example of providing too big replication factor for a hypertable with 2 attached data nodes:
```
```suggestion
Example of providing too big of a replication factor for a hypertable with 2 attached data nodes:
```

k-rus: @solugebefola Will it be better to use `too high replication factor`?

solugebefola: still needs 'of a' , as in
> too high of a replication factor

Either 'high' or 'big' works, it's up to you which you prefer

api.md
k-rus: Will remember for future. Fixed.

PR #282, opened 2020-01-22
Combine (de)compress_chunk with show_chunks for time range lookup

mfreed:


solugebefola:
@mike please label with version

PR #286, opened 2020-02-05
Neater command to add Debian/Ubuntu repo

hongquan:
The old command contains too many quotes.
Also stop user from adding pgdg repo for Ubuntu 19.04+, because that
repo will install Postgres 12, which is not supported by TimescaleDB
yet.


solugebefola:
@RobAtticus how far back would this change apply version-wise?

RobAtticus:
Let's just do it for 1.5

PR #290, opened 2020-02-07
Add note about deadlock during create_hypertable

k-rus:
Adds warning of deadlocking during `create_hypertable` for tables with foreign key constraints.

Closes https://github.com/timescale/timescaledb/pull/1677

### DIFF
--- a/api.md
+++ b/api.md
@@ -295,7 +295,8 @@ converting table itself.
 The deadlock might abort `create_hypertable`. 
 >
 >The deadlock can be prevented by manually obtaining `SHARE ROW EXCLUSIVE` lock 
-on the referenced tables before calling `create_hypertable` in the same transaction.
+on the referenced tables before calling `create_hypertable` in the same transaction,
+see [PostgreSQL documentation][postgres-lock] for the syntax.
 <!-- -->
 
 #### Units [](create_hypertable-units)
@@ -2803,6 +2804,7 @@ and then inspect `dump_file.txt` before sending it together with a bug report or
 [postgres-createtablespace]: https://www.postgresql.org/docs/current/sql-createtablespace.html
 [postgres-cluster]: https://www.postgresql.org/docs/current/sql-cluster.html
 [postgres-altertable]: https://www.postgresql.org/docs/current/sql-altertable.html
+[postgres-lock]: https://www.postgresql.org/docs/current/sql-lock.html
 [migrate-from-postgresql]: /getting-started/migrating-data
 [memory-units]: https://www.postgresql.org/docs/current/static/config-setting.html#CONFIG-SETTING-NAMES-VALUES
 [telemetry]: /using-timescaledb/telemetry
--- a/api.md
+++ b/api.md
@@ -289,7 +289,7 @@ columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must inc
 partitioning key.
 
 >:WARNING: Running `create_hypertable` with data migration on a table with foreign key contraints to other tables 
-might result in a deadlock if parallel transactions simulteniously try to insert data into the table and 
+might result in a deadlock if parallel transactions simultaneously try to insert data into the table and 
 the referenced tables. This can result in aborting `create_hypertable`. The deadlock can be prevented by 
 manually obtaining `SHARE ROW EXCLUSIVE` lock on the referenced tables  before calling `create_hypertable` 
 in the same transaction.
--- a/api.md
+++ b/api.md
@@ -288,10 +288,11 @@ to how you handle constraints. A hypertable can contain foreign keys to normal S
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
 
->:WARNING: Running `create_hypertable` with the data migration on a table with foreign key contraints 
-might result in a deadlock. This is likely to happen if parallel transactions simultaneously try 
-to insert data into tables referenced from the converting table and into the converting table itself. 
-The deadlock can abort `create_hypertable`. 
+>:TIP: Running `create_hypertable` with the data migration on a table with foreign key contraints 
+might result in a deadlock. This is likely to happen when concurrent transactions simultaneously try 
+to insert data into tables, which are referenced in the foreign key constraints, and into the
+converting table itself. 
+The deadlock might abort `create_hypertable`. 
 >
 >The deadlock can be prevented by manually obtaining `SHARE ROW EXCLUSIVE` lock 
 on the referenced tables before calling `create_hypertable` in the same transaction.
--- a/api.md
+++ b/api.md
@@ -278,7 +278,8 @@ as a table with column headings.
 
 >:WARNING: The use of the `migrate_data` argument to convert a non-empty table can
 lock the table for a significant amount of time, depending on how much data is
-in the table.
+in the table. It can also run into deadlock if foreign key constraints exist to other
+tables.
 >
 >If you would like finer control over index formation and other aspects
     of your hypertable, [follow these migration instructions instead][migrate-from-postgresql].
@@ -287,14 +288,10 @@ in the table.
 to how you handle constraints. A hypertable can contain foreign keys to normal SQL table
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
-
->:TIP: Running `create_hypertable` with the data migration on a table with foreign key contraints 
-might result in a deadlock. This is likely to happen when concurrent transactions simultaneously try 
-to insert data into tables, which are referenced in the foreign key constraints, and into the
-converting table itself. 
-The deadlock might abort `create_hypertable`. 
 >
->The deadlock can be prevented by manually obtaining `SHARE ROW EXCLUSIVE` lock 
+>The deadlock is likely to happen when concurrent transactions simultaneously try 
+to insert data into tables that are referenced in the foreign key constraints, and into the
+converting table itself. The deadlock can be prevented by manually obtaining `SHARE ROW EXCLUSIVE` lock 
 on the referenced tables before calling `create_hypertable` in the same transaction,
 see [PostgreSQL documentation][postgres-lock] for the syntax.
 <!-- -->
--- a/api.md
+++ b/api.md
@@ -290,7 +290,7 @@ columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must inc
 partitioning key.
 >
 >The deadlock is likely to happen when concurrent transactions simultaneously try 
-to insert data into tables that are referenced in the foreign key constraints, and into the
+to insert data into tables that are referenced in the foreign key constraints and into the
 converting table itself. The deadlock can be prevented by manually obtaining `SHARE ROW EXCLUSIVE` lock 
 on the referenced tables before calling `create_hypertable` in the same transaction,
 see [PostgreSQL documentation][postgres-lock] for the syntax.
--- a/api.md
+++ b/api.md
@@ -287,6 +287,12 @@ in the table.
 to how you handle constraints. A hypertable can contain foreign keys to normal SQL table
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
+
+>:WARNING: Running `create_hypertable` with data migration on a table with foreign key contraints to other tables 
+might result in a deadlock if parallel transactions simulteniously try to insert data into the table and 
+the referenced tables. This can result in aborting `create_hypertable`. The deadlock can be prevented by 
+manually obtaining `SHARE ROW EXCLUSIVE` lock on the referenced tables  before calling `create_hypertable` 
+in the same transaction.
 <!-- -->
 
 #### Units [](create_hypertable-units)
--- a/api.md
+++ b/api.md
@@ -288,11 +288,13 @@ to how you handle constraints. A hypertable can contain foreign keys to normal S
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
 
->:WARNING: Running `create_hypertable` with data migration on a table with foreign key contraints to other tables 
-might result in a deadlock if parallel transactions simultaneously try to insert data into the table and 
-the referenced tables. This can result in aborting `create_hypertable`. The deadlock can be prevented by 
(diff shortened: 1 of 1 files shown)


k-rus:
@solugebefola I guess this PR targets the closest docs release. Let me know or assign correct version label.

api.md
```diff
@@ -287,6 +287,12 @@ in the table.
 to how you handle constraints. A hypertable can contain foreign keys to normal SQL table
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
+
+>:WARNING: Running `create_hypertable` with data migration on a table with foreign key contraints to other tables 
+might result in a deadlock if parallel transactions simulteniously try to insert data into the table and
```
```suggestion
might result in a deadlock if parallel transactions simultaneously try to insert data into the table and 
```

api.md
```diff
@@ -287,6 +287,12 @@ in the table.
 to how you handle constraints. A hypertable can contain foreign keys to normal SQL table
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
+
+>:WARNING: Running `create_hypertable` with data migration on a table with foreign key contraints to other tables
```
erimatnor: Maybe this should be a `NOTE` and not `WARNING` (if such a distinction exists).

k-rus: I don't have strong opinion and troubled to decide, which one to use. It would be nice to hear reasoning why one over another.

api.md
```diff
@@ -287,6 +287,12 @@ in the table.
 to how you handle constraints. A hypertable can contain foreign keys to normal SQL table
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
+
+>:WARNING: Running `create_hypertable` with data migration on a table with foreign key contraints to other tables 
+might result in a deadlock if parallel transactions simulteniously try to insert data into the table and 
+the referenced tables. This can result in aborting `create_hypertable`. The deadlock can be prevented by 
+manually obtaining `SHARE ROW EXCLUSIVE` lock on the referenced tables  before calling `create_hypertable` 
+in the same transaction.
```
mkindahl: Hmm... quite hard to read. Too many indirect references and digressions. Here is a slightly improved version, but I think it need some more work.

> >:WARNING: If you run `create_hypertable` on a table with foreign key constraints 
> you might have a deadlock. This can happen if other transactions try to insert data into the foreign table at the same time as the `create_hypertable` is running. As a result, the `create_hypertable` will be aborted.
> >
> > You can prevent the deadlock by executing `SHARE ROW EXCLUSIVE` in the same transaction as `create_hypertable`.

k-rus: It is difficult to disagree that my formulation is too complicated :)
I made some improvements.

api.md
```diff
@@ -287,6 +287,14 @@ in the table.
 to how you handle constraints. A hypertable can contain foreign keys to normal SQL table
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
+
+>:WARNING: Running `create_hypertable` with the data migration on a table with foreign key contraints
```
gayyappan: I would change this from a WARNING to a NOTE (or some other keyword like FAQ . any suggestions @solugebefola ).

api.md
```diff
@@ -287,6 +287,14 @@ in the table.
 to how you handle constraints. A hypertable can contain foreign keys to normal SQL table
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
+
+>:WARNING: Running `create_hypertable` with the data migration on a table with foreign key contraints 
+might result in a deadlock. This is likely to happen if parallel transactions simultaneously try
```
gayyappan: reword "if parallel transactions" to "when concurrent transactions"

api.md
gayyappan: I strongly second using "Note" instead of warning. These are potential issues that we want developers to be aware of rather than problems with the db. It would  be appropriate for an FAQ  section

api.md
```diff
@@ -288,11 +288,13 @@ to how you handle constraints. A hypertable can contain foreign keys to normal S
...
 partitioning key.
 
->:WARNING: Running `create_hypertable` with data migration on a table with foreign key contraints to other tables 
-might result in a deadlock if parallel transactions simultaneously try to insert data into the table and 
-the referenced tables. This can result in aborting `create_hypertable`. The deadlock can be prevented by 
-manually obtaining `SHARE ROW EXCLUSIVE` lock on the referenced tables  before calling `create_hypertable` 
-in the same transaction.
+>:WARNING: Running `create_hypertable` with the data migration on a table with foreign key contraints 
+might result in a deadlock. This is likely to happen if parallel transactions simultaneously try 
+to insert data into tables referenced from the converting table and into the converting table itself.
```
gayyappan: this is confusing.  How about `insert data into the foreign tables referenced by the hypertable or into the hypertable itself.`

api.md
k-rus: @gayyappan @erimatnor There is not such thing as note, only either `TIP` or `WARNING`. I can switch to tip.

api.md
k-rus: > foreign tables

Unfortunately others are confused with term `foreign table`, which means remote table accessible through foreign data wrapper.

gayyappan: How about "insert data into the foreign tables (i.e tables referenced by the hypertable's foreign key constraints)".  This clarifies what is meant by foreign tables here.

k-rus: I improved the sentence. [It](https://github.com/timescale/docs.timescale.com-content/pull/290/files#diff-00e71df22262087fd8ad820708997657R292-R294) should be better, but I am not sure if it is enough.

api.md
```diff
@@ -287,6 +287,15 @@ in the table.
...
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
+
+>:TIP: Running `create_hypertable` with the data migration on a table with foreign key contraints 
+might result in a deadlock. This is likely to happen when concurrent transactions simultaneously try 
+to insert data into tables, which are referenced in the foreign key constraints, and into the
+converting table itself. 
+The deadlock might abort `create_hypertable`. 
+>
+>The deadlock can be prevented by manually obtaining `SHARE ROW EXCLUSIVE` lock
```
mfreed: I'm not sure this will be shown in the same TIP?  Also, do you want to give the syntax on how to do so, or pointer to postgres docs?

mfreed: cc: @solugebefola re: multi-paragraph :TIP: notation.

k-rus: > multi-paragraph :TIP: notation.

There is an example of Warning with multi-paragraph. I copied the usage from there

k-rus: > Also, do you want to give the syntax on how to do so, or pointer to postgres docs?

Fixed.

api.md
solugebefola: Only special tags are :WARNING: and :TIP: right now.

api.md
```diff
@@ -287,6 +287,15 @@ in the table.
 to how you handle constraints. A hypertable can contain foreign keys to normal SQL table
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
+
+>:TIP: Running `create_hypertable` with the data migration on a table with foreign key contraints 
+might result in a deadlock. This is likely to happen when concurrent transactions simultaneously try 
+to insert data into tables, which are referenced in the foreign key constraints, and into the
+converting table itself.
```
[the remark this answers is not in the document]
k-rus: @gayyappan To get a deadlock it is necessary to insert into FK table first and then into the converting table. So the parallel transaction will try to obtain the lock on the converting table, which is locked at the beginning of `create_hypertable`, while keeping lock on the FK table. `create_hypertable` will try to create FK at some point and will request lock on the FK table. Thus deadlock will happen.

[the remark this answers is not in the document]
gayyappan: so we have 2 concurrent  txns : 
txn1 : create_hypertable( regtable, ...)
txn2: insert into foreign table
Do we need an insert in txn1 (i.e. migrate_date ==  true) for the deadlock? That's what I am trying to clarify. Or do we have another txn i.e. txn3 that is inserting into regtable?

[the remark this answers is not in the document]
k-rus: @gayyappan Thank you for clarifying your question. As it said in the first sentence, it is with the data migration. If no data migration then no need to create chunks, which create FKs in them.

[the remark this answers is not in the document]
gayyappan: Should have re-read the sentence. Thanks!

api.md
```diff
@@ -287,6 +287,15 @@ in the table.
 to how you handle constraints. A hypertable can contain foreign keys to normal SQL table
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
+
+>:TIP: Running `create_hypertable` with the data migration on a table with foreign key contraints
```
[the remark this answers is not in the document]
k-rus: Fixed

api.md
```diff
@@ -287,6 +287,15 @@ in the table.
 to how you handle constraints. A hypertable can contain foreign keys to normal SQL table
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
+
+>:TIP: Running `create_hypertable` with the data migration on a table with foreign key contraints 
+might result in a deadlock. This is likely to happen when concurrent transactions simultaneously try 
+to insert data into tables, which are referenced in the foreign key constraints, and into the
```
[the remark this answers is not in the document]
k-rus: Fixed

k-rus:
I moved the proposed text into the existing warning. Minor fixes.

api.md
```diff
@@ -278,7 +278,8 @@ as a table with column headings.
 
 >:WARNING: The use of the `migrate_data` argument to convert a non-empty table can
 lock the table for a significant amount of time, depending on how much data is
-in the table.
+in the table. It can also run into deadlock if foreign key constraints exist to other
+tables.
```
mkindahl: Suggestion, reverse the order to make it easier to read:

> Depending on the amount of data in the table, use of `migrate_table` can cause long lock times. Using `migrate_data` can also cause deadlocks if the table have foreign key constraints to other tables.

api.md
```diff
@@ -287,6 +288,12 @@ in the table.
 to how you handle constraints. A hypertable can contain foreign keys to normal SQL table
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
+>
+>The deadlock is likely to happen when concurrent transactions simultaneously try 
+to insert data into tables that are referenced in the foreign key constraints, and into the
+converting table itself. The deadlock can be prevented by manually obtaining `SHARE ROW EXCLUSIVE` lock
```
mkindahl: Many nested and interleaved concepts making it hard to read, suggest something like this:

> If the table have foreign key constraints to other tables, a deadlock can happen if other transactions insert into the referenced table at the same time as data is migrated into the hypertable. The deadlock can be prevented ....

k-rus: IMHO starting sentence with `if` is not a good writing style. And your suggestion unfortunately is rather hard to read due to 2 ifs in it.
I will see if I can simplify the sentence.

api.md
k-rus: I don't want to modify the text, which was before me. It will just increase amount of discussions.
IMHO the sentence must start with `migrate_data` to make it clear that it is related to usage of this argument.

solugebefola:
What's the status of this PR @k-rus ?  I see changes requested, so I'm waiting to review until those are addressed.

k-rus:
> What's the status of this PR @k-rus ? I see changes requested, so I'm waiting to review until those are addressed.

@solugebefola I wasn't able to reach a version, which satisfy reviewers.

solugebefola:
@k-rus I'm going to close this PR until you have time to add changes?

k-rus:
> @k-rus I'm going to close this PR until you have time to add changes?

@solugebefola Have you looked to the text? Do you have an opinion?

api.md
erimatnor: I guess tip makes sense, but don't feel strongly. Warning seems too alarming for this thing, IMO.

k-rus:
@mkindahl @gayyappan You are currently blocking this PR. I spent quite some time trying to figure out how to improve the text further, but I don't see any good way, which will really improve the text. Is there text incorrect or unreadable, so it cannot be accepted?

api.md
```diff
@@ -287,6 +288,12 @@ in the table.
 to how you handle constraints. A hypertable can contain foreign keys to normal SQL table
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
+>
+>The deadlock is likely to happen when concurrent transactions simultaneously try 
+to insert data into tables that are referenced in the foreign key constraints and into the
+converting table itself. The deadlock can be prevented by manually obtaining `SHARE ROW EXCLUSIVE` lock 
+on the referenced tables before calling `create_hypertable` in the same transaction,
+see [PostgreSQL documentation][postgres-lock] for the syntax.
```
mkindahl:
```suggestion
>If you have a foreign key constraint from the hypertable to a normal SQL table, a deadlock can occur if concurrent transactions try to insert data into both the referenced table and the hypertable. The deadlock can be prevented by manually locking the referenced table using `SHARE ROW EXCLUSIVE` in the transaction before calling `create_hypertable`. See [PostgreSQL documentation][postgres-lock] for the syntax.
```

k-rus: @mkindahl Thank you for providing suggestion. I see several issues with it, which make it more complicated and not improving over the current text:
1. There is no hypertable at this moment, so I don't think it is correct to say `hypertable`.
2. The beginning of the suggestion repeats what is already said in the beginning of the warning. I think it is important to avoid repeating text. The beginning will be suitable if the paragraph is moved into a separate warning as it was originally. The issue is that it will be too many warnings and many reviewers think it should not be a warning, however I don't think it is a tip either.
3. I tried to fix your last sentence to be correct, but then it becomes more ambiguous.

I am not fully satisfied with my text, but I wasn't able to find a way to really improve it. I think I already spent too much time on this text.

erimatnor:
Nit: typo in commit/PR title.

k-rus:
> Nit: typo in commit/PR title.

Good catch. I fixed it in the PR title. Commit messages in docs are full mess with the current approach. I don't think it is worth to fix there.

PR #304, opened 2020-02-21
Organize tutorials into categories

CoolAssPuppy:
Setup for future expansion of tutorials.

### DIFF
--- a/tutorials.md
+++ b/tutorials.md
@@ -1,24 +1,38 @@
 # Tutorials
+We've created a host of code-focused tutorials that will help you get
+started with *TimescaleDB*. 
+
+For most of these tutorials, you will also need a working instance of *TimescaleDB*. The fastest way
+to get started with *TimescaleDB* is [setting up *Timescale Cloud*][timescale-cloud-install].
+*Timescale Cloud* is a hosted version of *TimescaleDB*. If you prefer,
+you can download the [TimescaleDB Community Edition][timescale-community-install]
+and run it on your desktop or your own cloud infrastructure.
+
+### Common scenarios for using TimescaleDB
 
 - **[Start Here - Hello Timescale][Hello Timescale]**: If you are new to TimescaleDB
-or even SQL, check
-out our tutorial with NYC taxicab data to get an idea of the capabilities our
-database has to offer.
-- **[Geospatial data][postGIS]**: If you want to delve into the further potential of
-TimescaleDB, we have guidance for more advanced topics
-like PostGIS integration.
+or even SQL, check out our tutorial with NYC taxicab data to get an idea of the 
+capabilities our database has to offer.
 - **[Time-series Forecasting][Forecasting]**: Use R, Apache MADlib and Python to perform
 data analysis and make forecasts on your data.
+
+### How to use specific TimescaleDB features
+
 - **[Replication][]**: TimescaleDB takes advantage of well established PostgreSQL methods for replication.  Here we provide a detailed guide along with additional resources for setting up streaming replicas.
 - **[Continuous Aggregates][]**: Getting started with continuous aggregates.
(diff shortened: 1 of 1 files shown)


tutorials.md
```diff
@@ -1,19 +1,36 @@
...
 - **[Start Here - Hello Timescale][Hello Timescale]**: If you are new to TimescaleDB
-or even SQL, check
-out our tutorial with NYC taxicab data to get an idea of the capabilities our
-database has to offer.
+or even SQL, check out our tutorial with NYC taxicab data to get an idea of the 
+capabilities our database has to offer.
 - **[Geospatial data][postGIS]**: If you want to delve into the further potential of
-TimescaleDB, we have guidance for more advanced topics
-like PostGIS integration.
+TimescaleDB, we have guidance for more advanced topics like PostGIS integration.
```
solugebefola: Just a note: With the redesign of the Hello Timescale tutorial, the place where this links to is not really a good place to jump the reader to, as it's not really a standalone section anymore (and it was always a little weird to jump the reader into the middle of another tutorial).  If possible, it would be a good idea to make a separate PostGIS tutorial, even something short; or make the PostGIS section of Hello Timescale into it's own clearly distinct section.

solugebefola: I might suggest in the interim to remove this link altogether.

mfreed:
Sorry, I should have been clearer:  I think we should also add this top-level organization to the left-hand menu in the docs, so that under the Tutorial heading we now have 4 sections, something like:

* Common Scenarios
* Specific TimescaleDB Features
* Product Integrations
* Additional Resources

Each of which then has a set of sub-sections corresponding to the actual tutorials.  I believe this would involve editting `page-index.js`.

CoolAssPuppy:
I think I got this right, but didn't know if an ANCHOR could have "children". Let me know if this isn't possible.

mfreed:
@solugebefola Can you review for correctness?

page-index/page-index.js
```diff
@@ -439,53 +439,81 @@ const pageIndex = [
...
-                href: "outflux",
-                children: []
+                Title: "Common Scenarios",
+                type: ANCHOR,
+                href: "common-scenarios",
+                children: [
+                    {
+                        Title: "Get started: Hello, Timescale!",
+                        type: PAGE,
+                        href: "tutorial-hello-timescale",
```
solugebefola: So the directory structure uses 'parent' items to determine the page path.  For example, for this entry, the path would be `/tutorials/common-scenarios/tutorial-hello-timescale`.  Unfortunately, this doesn't work for anchors, since you need to add a `#` to the `href` value and adding additional path elements to the end of that doesn't work functionally (or semantically).

The only option I could see would be to add a hidden `DIRECTORY` type items that would have the same href as the anchors, but without the `#` and put the tutorial `PAGE` items as children of those.  The links to those tutorials would have to be updated to reflect that (but they need to be updated anyway)

solugebefola: Let me know if this doesn't make sense

CoolAssPuppy:
@mfreed this is a fairly involved change. are you sure you want to go down this path vs. the simple cleaned up tutorials page? With this approach, tutorial names wouldn't show up in the sidebar, and that makes me a bit squeamish when it comes to discovery of new content.

page-index/page-index.js
mfreed: @solugebefola Maybe we should just accept this as is, then you could do another PR with the above fixes?

mfreed:
Yeah, I was wrong.  Let's just keep this PR simple, and if we want to address the above, use a separate PR.

Sorry!  So it otherwise then LGTM.

CoolAssPuppy:
Oh boy, I tried doing a git reset and, well, let's see if that worked! :-)

PR #314, opened 2020-02-28
Info about data retention for continous aggregates

mkindahl:
Adding information to the data retention section about how to handle
data retention for tables that have continous aggregates.

Subtask of: timescale/timescaledb-private#622

### DIFF
--- a/using-timescaledb/data-retention.md
+++ b/using-timescaledb/data-retention.md
@@ -1,14 +1,41 @@
 # Data Retention [](data-retention)
 
-TimescaleDB allows efficient deletion of old data at the chunk level
-using the `drop_chunks` function.
+An intrinsic part of time-series data is that new data is accumulated
+and old data is rarely, if ever, updated and the relevance of the data
+diminishes over time.  It is therefore often desirable to delete old
+data to save disk space.
+
+As an example, if you have a hypertable definition of `conditions`
+where you collect raw data into chunks of one day:
+
+```sql
+CREATE TABLE conditions(
+    time TIMESTAMPTZ NOT NULL,
+    device INTEGER,
+    temperature FLOAT
+);
+
+SELECT * FROM create_hypertable('conditions', 'time',
+       chunk_time_interval => '1 day'::interval);
+```
+
+If you collect a lot of data and realize that you never actually use
+raw data older than 30 days, you might want to delete data older than
+30 days from `conditions`.
+
+However, deleting large swaths of data from tables can be costly and
+slow if done row-by-row using the standard `DELETE` command. Instead,
+TimescaleDB provides a function `drop_chunks` that quickly drop data
+at the granularity of chunks without incurring the same overhead.
+
+For example:
 
 ```sql
 SELECT drop_chunks(interval '24 hours', 'conditions');
 ```
 
-This will drop all chunks from the hypertable 'conditions' that _only_ include
-data older than this duration, and will _not_ delete any
+This will drop all chunks from the hypertable `conditions` that _only_
+include data older than this duration, and will _not_ delete any
 individual rows of data in chunks.
 
 For example, if one chunk has data more than 36 hours old, a second
@@ -19,6 +46,70 @@ the `conditions` hypertable will still have data stretching back 36 hours.
 
 For more information on the `drop_chunks` function and related
 parameters, please review the [API documentation][drop_chunks].
+
+### Data Retention with Continuous Aggregates
+
+Continuing on the example above, you have discovered that you need
+daily summary of the average, minimum, and maximum temperature for
+each day, so you have created a continuous aggregate
+`conditions_summary_daily` to collect this data:
+
+```sql
+CREATE VIEW conditions_summary_daily
+WITH (timescaledb.continuous) AS
+SELECT device,
+       time_bucket('1 day'::interval, "time") AS bucket,
+       AVG(temperature),
+       MAX(temperature),
+       MIN(temperature)
+FROM conditions
+GROUP BY device, bucket;
+```
+
+When you now try to drop chunks from `conditions` you get an error:
+
+```
+postgres=# SELECT drop_chunks('30 days'::interval, 'conditions');
+ERROR:  cascade_to_materializations options must be set explicitly
+HINT:  Hypertables with continuous aggs must have the cascade_to_materializations option set to either true or false explicitly.
+```
+
+Since the data in `conditions_summary_daily` is now dependent on the
+data in `conditions` you have to explicitly say if you want to remove
+this data from the continuous aggregate using the
+`cascade_to_materializations` option. Not giving a value when there is
+a continuous aggregate defined on a hypertable will generate an error,
+as you can see above. To drop the chunks from `condition` and cascade
+it to drop the corresponding chunks in the continuous aggregate you
+set `cascade_to_materializations` to `TRUE`:
+
+```
+postgres=# SELECT COUNT(*)
+postgres-#   FROM drop_chunks('30 days'::INTERVAL, 'conditions',
+postgres-#                    cascade_to_materializations => TRUE);
+ count 
+-------
+    61
+(1 row)
+```
+
+To only remove chunks from the hypertable `conditions` and not cascade
+to dropping chunks on the continuous aggregate
+`conditions_summary_daily` you can provide the value `FALSE` to
+`cascade_to_materialization`, but only after you have removed the
+dependency on the corresponding chunks in the continuous aggregate. To
+do that, use `ALTER VIEW` to change the
+`ignore_invalidation_older_than` parameter in the continuous aggregate
+to the same range that you intend to remove from the hypertable.
+
+```sql
+ALTER VIEW conditions_summary_daily SET (
+   timescaledb.ignore_invalidation_older_than = '29 days'
+);
+
+SELECT drop_chunks('30 days'::interval, 'conditions',
+                   cascade_to_materialization => FALSE);
+```
 
 ### Automatic Data Retention Policies
 


using-timescaledb/data-retention.md
```diff
@@ -20,6 +20,46 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
+CREATE VIEW conditions_summary WITH (timescaledb.continous) AS
+SELECT location, time_bucket('1d', "time") AS day,
+       AVG(temperature) AS avg_temperature,
+       MAX(temperature) AS max_temperature,
+       MIN(temperature) AS min_temperature
+FROM conditions
+GROUP BY location, time_bucket('1d', "time")
+```
+
+To remove chunks older than, say 30 days, you would have to first set
```
gayyappan:
```suggestion
To remove chunks older than, say 30 days, from conditions you would have to first set
```

using-timescaledb/data-retention.md
```diff
@@ -20,6 +20,46 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
+the `ignore_invalidation_older_than`:
+
+```sql
+ALTER VIEW conditions_summary SET (
+  timescaledb.ignore_invalidation_older_than = '30 days'
+)
+```
+
+After that, you can drop the old chunks from the raw table
+`conditions` without triggering an update of the conditional
```
gayyappan:
```suggestion
`conditions` without triggering an update of the continuous
```

using-timescaledb/data-retention.md
```diff
@@ -20,6 +20,46 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
+aggregate:
+
+```sql
+SELECT drop_chunks(interval '30 days', 'conditions');
+```
+
 ### Automatic Data Retention Policies :enterprise_function:
```
gayyappan:
```suggestion
### Automatic Data Retention Policies 
```

using-timescaledb/data-retention.md
mkindahl: Will rewrite slightly.

using-timescaledb/data-retention.md
mkindahl: This might seem like a not, but I think I should handle this as a separate PR, since that require a different discussion.

gayyappan:
> Not sure I understand. This is about data retention for continuous aggregates. What do you think is missing?

This section addresses data retention policies for hypertables that have continuous aggregates that are dependent on then. We also need to document how the data from the "materialized hypertable" i.e. continuous aggregate can be dropped,.

using-timescaledb/data-retention.md
```diff
@@ -20,6 +20,46 @@ the `conditions` hypertable will still have data stretching back 36 hours.
 For more information on the `drop_chunks` function and related
 parameters, please review the [API documentation][drop_chunks].
 
+### Data Retention and Continuous Aggregates
+
+If you have a continous aggregate set up with a hypertable, it can be
+useful to keep the aggregated data but clean up old hypertable chunks
+to save memory. Since a drop of the old chunks will trigger an update
```
gayyappan: __
```suggestion
to reduce storage costs. Since a drop of the old chunks will trigger an update
```

using-timescaledb/data-retention.md
mkindahl: Uploaded as #326

mkindahl:
> > Not sure I understand. This is about data retention for continuous aggregates. What do you think is missing?
> 
> This section addresses data retention policies for hypertables that have continuous aggregates that are dependent on then. We also need to document how the data from the "materialized hypertable" i.e. continuous aggregate can be dropped,.

Added some descriptions about all the situations and pushed a new version. PTAL.

using-timescaledb/data-retention.md
```diff
@@ -1,15 +1,21 @@
 # Data Retention [](data-retention)
 
-TimescaleDB allows efficient deletion of old data at the chunk level
-using the `drop_chunks` function.
+As your tables grow, it can become cumbersome to keep all that data
+available in the hypertable, so you will need to remove data that are
+no longer relevant. Since data in hypertables are chunked, a normal
+`DELETE` will be very inefficient and TimescaleDB instead provide the
+`drop_chunks` function that is able to quickly drop chunks that are
+not needed any more.
```
erimatnor: Not sure "cumbersome" is the appropriate/accurate term here. Also, DELETEs aren't inefficient because hypertables are chunked; it's simply inefficient the way it works in PostgreSQL (chunking actually helps here even for DELETEs).  So this is not an accurate description.

I suggest something like:

```suggestion
The relevance of time-series data often diminishes over time and therefore it is often desirable to delete old data to save disk space. However, deleting large swaths of data from tables can be costly and slow if done row-by-row using the standard `DELETE` command. The recommended way to delete old data on a hypertable is therefore to use `drop_chunks`, which quickly drops data at the granularity of chunks without incurring the same overhead.
```

using-timescaledb/data-retention.md
```diff
@@ -1,15 +1,21 @@
...
+For example:
 
 ```sql
 SELECT drop_chunks(interval '24 hours', 'conditions');
 ```
 
-This will drop all chunks from the hypertable 'conditions' that _only_ include
-data older than this duration, and will _not_ delete any
-individual rows of data in chunks.
+This will drop all chunks from the hypertable `conditions` that
```
erimatnor: This is wrong because it changes the meaning of what was there before and is no longer reflecting how `drop_chunks` works.

Note that `drop all chunks ... that include data older than 24 hours` is not correct, because if a chunk includes some data that is older than 24 hours but some that is younger than 24 hours, then that chunk will _not_ be dropped. In other words, if the drop-threshold hits somewhere within a chunk, that chunk will remain in its entirety.

using-timescaledb/data-retention.md
```diff
@@ -20,6 +26,68 @@ the `conditions` hypertable will still have data stretching back 36 hours.
 For more information on the `drop_chunks` function and related
 parameters, please review the [API documentation][drop_chunks].
 
+### Data Retention and Continuous Aggregates
```
erimatnor:
```suggestion
### Data Retention on Continuous Aggregates
```

using-timescaledb/data-retention.md
```diff
@@ -20,6 +26,68 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
+
+The function `drop_chunks` can be used to control data retention for
+the hypertable as well as the continuous aggregate. Given a hypertable
```
erimatnor:
```suggestion
The function `drop_chunks` can also be used to control data retention on continuous aggregates. Given a hypertable
```

using-timescaledb/data-retention.md
```diff
@@ -20,6 +26,68 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
+SELECT location,
+       time_bucket('1d', "time") AS day,
+       AVG(temperature) AS avg_temperature,
+       MAX(temperature) AS max_temperature,
+       MIN(temperature) AS min_temperature
+FROM conditions
+GROUP BY location, time_bucket('1d', "time")
+```
+
+Data retention on a continous aggregate will normally not propagate to
```
erimatnor: This doesn't continue the sentence started before the view example as one would expect.

using-timescaledb/data-retention.md
```diff
@@ -20,6 +26,68 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
+the hypertable, so the call below will only remove chunks from
+`conditions_summary`.
+
+```sql
+SELECT drop_chunks(INTERVAL '30 days', 'conditions_summary');
```
erimatnor: Is `INTERVAL` really required here?

using-timescaledb/data-retention.md
```diff
@@ -20,6 +26,68 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
+```
+
+If you want to remove chunks from `conditions` as well, you need to
+cascade the drop chunks to the hypertable using the
+`cascade_to_materialization` option:
+
+```sql
+SELECT drop_chunks(INTERVAL '30 days', 'conditions_summary',
```
erimatnor: This doesn't look right. I thought the `cascade_to_materialization` is only when you are dropping on the hypertable (`conditions`) and want it to cascade to the aggregate. Not the other way around. I guess the `materialization` here is the aggregate, so it doesn't make sense to cascade to it if you are already dropping on the aggregate (i.e., there's nothing to cascade to). AFAIK, the cascade option in this example does nothing.

But I understand the confusion because this parameter tripped me up several times myself and I don't yet fully understand it. So, I might be wrong here.

using-timescaledb/data-retention.md
```diff
@@ -20,6 +26,68 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
+                   cascade_to_materialization => TRUE);
+```
+
+It can also be useful to keep the aggregated data only but clean up
+old hypertable chunks to reduce storage cost. Since a drop of the old
+chunks will trigger an update of the continuous aggregate and remove
```
erimatnor: Is it really accurate that drop_chunks on a hypertable will also remove the downsampled data (aggregate)? This doesn't sound right to me. AFAIK, it would only cascade to the aggregate if `cascade_to_materialization` is set to TRUE. (Again, this confuses me too so I am not 100% sure about this).

using-timescaledb/data-retention.md
mkindahl: Right, an "only" is missing. The original sentence was long and a little hard to read so tried to improve it, but restored the original.

using-timescaledb/data-retention.md
mkindahl: Not entirely accurate: this is also about data retention for the raw table that the continuous aggregate is using.

using-timescaledb/data-retention.md
mkindahl: Nope, but I think it is clearer.

using-timescaledb/data-retention.md
mkindahl: This was my understanding, but maybe @gayyappan can clarify if it is accurate.

using-timescaledb/data-retention.md
mkindahl: I tested just to make sure: the `INTERVAL` is needed here, a string will not do.

using-timescaledb/data-retention.md
erimatnor: I am confused. With raw table, do you mean the "internal" hypertable which provides storage for the view or the non-aggregated table which you define the aggregate on (i.e., `conditions`)? I assume the latter, and in that case I don't think this is about retention on that non-aggregated table because dropping chunks on the aggregate shouldn't affect the non-aggregated table.

This relates to the confusion about `cascade_to_materialization`

using-timescaledb/data-retention.md
erimatnor: The non-aggregated table `conditions` is not a materialization so if the drop_chunks on the aggregate that cascades to `conditions` because `cascade_to_materialization` is set doesn't make any sense.

gayyappan: drop of the raw hypertable chunks cascades to aggregates ( IF cascade_to_materializations is set to true). Not vice-versa.

using-timescaledb/data-retention.md
mkindahl: No, but dropping chunks on the non-aggregated table affects the aggregated table (or can, at least, if you set it up correctly), so the section does not only cover data retention on continuous aggregates.

using-timescaledb/data-retention.md
```diff
@@ -1,14 +1,21 @@
 # Data Retention [](data-retention)
 
-TimescaleDB allows efficient deletion of old data at the chunk level
-using the `drop_chunks` function.
+The relevance of time-series data often diminishes over time and
+therefore it is often desirable to delete old data to save disk
+space. However, deleting large swaths of data from tables can be
+costly and slow if done row-by-row using the standard `DELETE`
+command. To alleviate this, TimescaleDB provide a function
```
erimatnor:
```suggestion
command. To alleviate this, TimescaleDB provides a function
```

using-timescaledb/data-retention.md
```diff
@@ -1,14 +1,21 @@
...
+`drop_chunks` that can be used to quickly drops data at the
```
erimatnor:
```suggestion
`drop_chunks` that can be used to quickly drop data at the
```

using-timescaledb/data-retention.md
```diff
@@ -20,6 +27,41 @@ the `conditions` hypertable will still have data stretching back 36 hours.
 For more information on the `drop_chunks` function and related
 parameters, please review the [API documentation][drop_chunks].
 
+### Data Retention and Continuous Aggregates
+
+The function `drop_chunks` can be used to control data retention for
+the continuous aggregate as well, but if you have a continuous
+aggregate defined on the table you need to provide a value for the
+`cascade_to_materialization`. Not giving a value when there is a
+continuous aggregate defined on a hypertable will generate an error.
```
erimatnor: > Not giving a value when there is a continuous aggregate defined on a hypertable will generate an error

This sounds super-odd to me. I would double-check that this is the case.

Just saying.

using-timescaledb/data-retention.md
```diff
@@ -20,6 +27,41 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
+
+Providing the value `FALSE` to `cascade_to_materialization` will only
+remove chunks from the hypertable and not cascade to dropping chunks
+on the continuous aggregate.
+
+It is not possible to remove chunks from the hypertable without
+allowing the corresponding chunks to be removed from the continuous
+aggregate, so it is necessary to alter the continuous aggregate and
+set `ignore_invalidation_older_than` to the range to remove.
```
erimatnor: This is also super-odd. Why isn't this set automatically when you do drop_chunks without cascade_to_materialization?

Again, just double-check that this is actually the case because UX here is very strange.

mkindahl: Here I am trying to drop chunks from `conditions` without having set `ignore_invalidations_older_than`, and then setting `ignore_invalidations_older_than` for the aggregate and repeating the operation.

```
postgres-#   FROM drop_chunks(INTERVAL '30 days', 'conditions',
postgres(#                    cascade_to_materializations => FALSE);
ERROR:  older_than must be greater than the timescaledb.ignore_invalidation_older_than parameter of public.conditions_summary
postgres=# ALTER VIEW conditions_summary SET (
postgres(#    timescaledb.ignore_invalidation_older_than = '30 days'
postgres(# );
ALTER VIEW
postgres=# SELECT COUNT(*) AS dropped_chunks
postgres-#   FROM drop_chunks(INTERVAL '30 days', 'conditions',
postgres(#                    cascade_to_materializations => FALSE);
NOTICE:  making sure all invalidations for public.conditions_summary have been processed prior to dropping chunks
 dropped_chunks 
----------------
           1080
(1 row)
```

using-timescaledb/data-retention.md
solugebefola: try 'with' then.

using-timescaledb/data-retention.md
```diff
@@ -20,6 +27,41 @@ the `conditions` hypertable will still have data stretching back 36 hours.
 For more information on the `drop_chunks` function and related
 parameters, please review the [API documentation][drop_chunks].
 
+### Data Retention and Continuous Aggregates
+
+The function `drop_chunks` can be used to control data retention for
+the continuous aggregate as well, but if you have a continuous
+aggregate defined on the table you need to provide a value for the
+`cascade_to_materialization`. Not giving a value when there is a
```
solugebefola:
```suggestion
`cascade_to_materialization` parameter. Not giving a value when there is a
```

using-timescaledb/data-retention.md
solugebefola: This statement is confusing.  Are you saying that you can't remove chunks from the hypertable without removing the dependency of the cagg on those dropped chunks?  I think you need to state that.  Your current statement implies that the user would be trying to remove the actual chunks from the cagg table (i.e. dropping chunks on `conditions_materialized_view` table, or whatever it's called).

using-timescaledb/data-retention.md
mkindahl: Done.

using-timescaledb/data-retention.md
mkindahl: Fixed wording.

using-timescaledb/data-retention.md
solugebefola: This still needs correction

using-timescaledb/data-retention.md
solugebefola: Also still needs correction

using-timescaledb/data-retention.md
```diff
@@ -20,6 +27,42 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
+continuous aggregate defined on a hypertable will generate an error.
+
+Providing the value `FALSE` to `cascade_to_materialization` will only
+remove chunks from the hypertable and not cascade to dropping chunks
+on the continuous aggregate.
+
+It is not possible to remove chunks from the hypertable without
+removing the dependency on the corresponding chunks in the continuous
+aggregate. To do that, use `ALTER VIEW` to change the
+`ignore_invalidation_older_than` parameter in continuous aggregate to
```
solugebefola:
```suggestion
`ignore_invalidation_older_than` parameter in the continuous aggregate to
```

using-timescaledb/data-retention.md
```diff
@@ -20,6 +27,42 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
+the range that you intend to remove from the hypertable.
```
solugebefola:
```suggestion
the same range that you intend to remove from the hypertable.
```

using-timescaledb/data-retention.md
```diff
@@ -20,6 +27,42 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
+the continuous aggregate as well, but if you have a continuous
+aggregate defined on the table you need to provide a value for the
+`cascade_to_materialization` parameter. Not giving a value when there is a
+continuous aggregate defined on a hypertable will generate an error.
+
+Providing the value `FALSE` to `cascade_to_materialization` will only
+remove chunks from the hypertable and not cascade to dropping chunks
+on the continuous aggregate.
+
+It is not possible to remove chunks from the hypertable without
```
solugebefola:
```suggestion
It is not possible to remove chunks from the hypertable without first
```

using-timescaledb/data-retention.md
```diff
@@ -20,6 +27,42 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
 parameters, please review the [API documentation][drop_chunks].
 
+### Data Retention with Continuous Aggregates
+
+The function `drop_chunks` can be used to control data retention for
+the continuous aggregate as well, but if you have a continuous
+aggregate defined on the table you need to provide a value for the
+`cascade_to_materialization` parameter. Not giving a value when there is a
+continuous aggregate defined on a hypertable will generate an error.
+
```
gayyappan: Could you please add the example and commands here?  i.e. create the view dependent on the condition table and illustrate the error using the example above  `` select drop_chunks(...); ``

using-timescaledb/data-retention.md
gayyappan:
```suggestion
It is possible to remove chunks from the hypertable after removing
```
use positive instead of double negative. I think that is easier to parse.

using-timescaledb/data-retention.md
```diff
@@ -20,6 +27,42 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
+```sql
+ALTER VIEW conditions_summary SET (
+   timescaledb.ignore_invalidation_older_than = '30 days'
+);
+
+SELECT drop_chunks(INTERVAL '30 days', 'conditions',
+                   cascade_to_materialization => FALSE);
+```
+
+If you want to remove chunks from `conditions_summary` as well, you
```
gayyappan:
```suggestion
On the other hand, if you want to remove chunks from `conditions_summary` and `conditions`,  you
```

using-timescaledb/data-retention.md
```diff
@@ -20,6 +27,42 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
+need to cascade the drop chunks to the continuous aggregate using the
```
gayyappan:
```suggestion
can cascade the drop chunks to the continuous aggregate using the
```

using-timescaledb/data-retention.md
mkindahl: The normal practice is to *not* show how failing examples, but rather show positive examples, which is shown below. Why do you want to show the negative example?

using-timescaledb/data-retention.md
mkindahl: Done.

using-timescaledb/data-retention.md
gayyappan: Since you called out the error condition, I asked to illustrate that.
1. Nevertheless, we should add the view definition. The example implicitly assumes that we have a view conditions_summary that depends on conditions. 
2. Also using the example to describe the ignore_invalidation_older_than parameter and the chunks that get dropped would help in clarifying the behavior here. I was wondering if a small table or figure would help show the timelines of what gets dropped and why it errors out. 
3. Would be good to add drop_chunks error for ignore_invalidation_older_than and describe the problem. (i.e. a negative example would be helpful in this case).

mkindahl: Ok, I did a significant rewrite of the start of the data retention chapter to provide a running example that progressively shows how to do data retention and also show the error you can run into. I think this is significantly clearer, but please have a look.

mkindahl: Note that I did not add a figure. I am creating a figure for timescale/docs.timescale.com-content#350 that outline the complete flow of data.

using-timescaledb/data-retention.md
```diff
@@ -20,6 +47,70 @@ the `conditions` hypertable will still have data stretching back 36 hours.
...
+SELECT device,
+       time_bucket('1 day'::interval, "time") AS bucket,
+       AVG(temperature),
+       MAX(temperature),
+       MIN(temperature)
+FROM conditions
+GROUP BY device, bucket;
+```
+
+When you now try to drop chunks from `conditions` you get strange error:
```
gayyappan:
```suggestion
When you now try to drop chunks from `conditions` you get an error:
```

Issue #309, opened 2020-02-26
Gapfill examples should infer range from WHERE


mfreed:
This uses an older syntax, which needs the START/STOP in the function declaration.  We should use the "easier/prettier" example of where this is inferred from a WHERE predicate:

https://docs.timescale.com/latest/using-timescaledb/reading-data#gap-filling

solugebefola:
Can

solugebefola:
Assuming closed with the PR merge

PR #321, opened 2020-03-01
Cleanup gapfill examples

mfreed:
- Remove start/finish from gapfill function given we infer it from WHERE clause
- Fix error in example performing mathematical manipulation on gapfill inputs


mfreed:
Fixes https://github.com/timescale/docs.timescale.com-content/issues/309

PR #325, opened 2020-03-03
Add section on integer-based continuous aggregates

mkindahl:
A section on integer-based continuous aggregates is added with some
examples on how to define a view using a custom-defined time interval.

Subtask of timescale/timescaledb-private#609

### DIFF
--- a/page-index/page-index.js
+++ b/page-index/page-index.js
@@ -383,6 +383,10 @@ const pageIndex = [
                         type: ANCHOR,
                         href: "#dropping-data"
                     }, {
+                        Title: "Advanced Topics",
+                        type: ANCHOR,
+                        href: "#advanced-usage"
+                    }, {
                         Title: "Best Practices",
                         type: ANCHOR,
                         href: "#best-practices"
--- a/tutorials/continuous-aggs-tutorial.md
+++ b/tutorials/continuous-aggs-tutorial.md
@@ -156,23 +156,14 @@ aggregates only for the data that satisfies the condition:
 `time_bucket('1h', pickup_datetime) < max(pickup_time) - '1h'` (if the
 `refresh_lag` is set to 1 hour)
 
-Both `refresh_lag` or `refresh_interval` can be set to an integer
-value, which is then interpreted as the number of microseconds for the
-lag or interval respectively.
-
-To keep the continuous aggregate up-to-date,
-you can set `timescaledb.refresh_lag` to a negative value. This will allow the continuous aggregate
-to update when new data comes in. However, do note that this comes with performance implications,
-since the aggregate query will be run more often. 
-
 The `timescaledb.max_interval_per_job` parameter is used when we want to limit
 the amount of data processed by an update of the continuous aggregate and use
 smaller or bigger batch sizes (the batching is done automatically by TimescaleDB).
 `timescaledb.max_interval_per_job` specifies the batch size.
 
-`refresh_lag` and `max_interval_per_job` are additional parameters that can be
-specified while creating or altering a continuous aggregate. Refer to the
-documentation for the syntax.
+The parameters `refresh_lag` and `max_interval_per_job` can be
+specified while creating or altering a continuous aggregate. Refer to
+the documentation for the syntax.
 
 The `timescaledb.ignore_invalidation_older_than` parameter is used
 when you want to avoid updating a continuous aggregate when modifying
@@ -249,13 +240,6 @@ refresh_lag                | 01:00:00
 refresh_interval           | 00:30:00
 max_interval_per_job       | 20:00:00
 materialization_hypertable | _timescaledb_internal._materialized_hypertable_2
-```
-
-We can also manually update the continuous aggregate query by using the
-`REFRESH` command.
-
-``` sql
-REFRESH MATERIALIZED VIEW cagg_rides_view;
 ```
 
 You will find more details about the API in the [documentation][tsdb_doc].
--- a/using-timescaledb/continuous-aggregates.md
+++ b/using-timescaledb/continuous-aggregates.md
@@ -159,10 +159,10 @@ aggregate and the aggregate has to be updated.
 
 By default, all modifications trigger a update of the continuous
 aggregate. If the `ignore_invalidation_older_than` parameter is set to
-a duration, modifications with a timestamp that is older will be
-ignored and not trigger an update of the continuous aggregate. As a
-result, refresh jobs will complete faster since they are only dealing
-with new data as part of the refresh.
+a duration, modifications with an older time will be ignored and not
+trigger an update of the continuous aggregate. As a result, refresh
+jobs will complete faster since they are only dealing with new data as
+part of the refresh.
 
 A common use case is to drop the raw data from the hypertable and just
 keep the downsampled data in the continuous aggregate. So, for
@@ -180,10 +180,6 @@ SELECT drop_chunks(INTERVAL '30 days', 'device_readings')
 
 You can read more about data retention with continuous aggregates in
 the [*Data retention*][sec-data-retention] section.
-
->:TIP: Most times the continuous aggregate view will be updated by the background job;
-  however, if you would like to run it yourself, you may use the
-  [`REFRESH MATERIALIZED VIEW` command][api-refresh-continuous-aggs].
 
 **Using `timescaledb.information` Views:**
 The various options used to create the continuous aggregate view, as well as its
@@ -245,6 +241,108 @@ dropped.
 The same argument must also be supplied to the [`add_drop_chunks_policy`
 function][api-add-drop-chunks] when creating a data retention policy for a
 hypertable with a continuous aggregate.  
+
+---
+
+### Advanced Topics [](advanced-usage)
+
+#### Continuous Aggregates using Integer-Based Time [](create-integer)
+
+Usually, continuous aggregates are defined on a [date/time-type](https://www.postgresql.org/docs/current/datatype-datetime.html) column, but it is
+also possible to create your own custom scheme for handling
+aggregation for tables that are using an integer time
+column. This can be useful if you have tables that use other measures
+of time that can be represented as integer values, such as nanosecond
+epochs, minutes since founding date, or whatever is suitable for your
+application.
+
+As an example, suppose that you have a table with CPU and disk usage
+for some devices where time is measured in
+[microfortnights][fff-system] (a microfortnight is a little more than
+a second). Since you are using an integer-valued column as time, you
+need to provide the chunk time interval when creating the
+hypertable. In this case, let each chunk consist of a millifortnight
+(a 1000 microfortnights, which is about 20 minutes).
+
+```sql
+CREATE TABLE devices(
+  time BIGINT,        -- Time in microfortnights since epoch
+  cpu_usage INTEGER,  -- Total CPU usage
+  disk_usage INTEGER, -- Total disk usage
+  PRIMARY KEY (time)
+);
+
+SELECT create_hypertable('devices', 'time',
+                         chunk_time_interval => 1000);
+```
+
+To define a continuous aggregate on a hypertable that is using an
+integer time dimension, it is necessary to have a function to get the
+current time in whatever representation that you are using and set it
+for the hypertable using
+[`set_integer_now_func`][api-set-integer-now-func]. The function can
+be defined as a normal PostgreSQL function, but needs to be
+[`STABLE`][pg-func-stable], take no arguments, and return an integer
+value of the same type as the time column in the table. In our case,
+this should suffice:
+
+```sql
+CREATE FUNCTION current_microfortnight() RETURNS BIGINT
+LANGUAGE SQL STABLE AS $$
+	SELECT CAST(1209600 * EXTRACT(EPOCH FROM CURRENT_TIME) / 1000000 AS BIGINT)
+$$;
+
+SELECT set_integer_now_func('devices', 'current_microfortnight');
+```
+
+Once the replacement for current time has been set up, you can define
+a continuous aggregate for the `devices` table.
+
+```sql
+CREATE VIEW devices_summary
+WITH (timescaledb.continuous) AS
+SELECT time_bucket('500', time) AS bucket,
+       avg(cpu_usage) AS avg_cpu,
+       avg(disk_usage) AS avg_disk
+     FROM devices
+     GROUP BY bucket;
+```
+
+You can now insert some rows to check if the aggregation works as
+expected.
+
+```sql
+CREATE EXTENSION tablefunc;
+
+INSERT INTO devices(time, cpu_usage, disk_usage)
+SELECT time,
+       normal_rand(1,70,10) AS cpu_usage,
+	   normal_rand(1,2,1) * (row_number() over()) AS disk_usage
+  FROM generate_series(1,10000) AS time;
+```
+
+>:TIP: You can use the `tablefunc` extension to generate a normal
+>distribution and use the `row_number` function to turn it into a
+>cumulative sequence.
+
+You can now check that the view contains the correct data. 
+
+```sql
+postgres=# SELECT * FROM devices_summary ORDER BY bucket LIMIT 10;
+ bucket |       avg_cpu       |       avg_disk
+--------+---------------------+----------------------
+      0 | 63.0000000000000000 |   6.0000000000000000
+      5 | 69.8000000000000000 |   9.6000000000000000
+     10 | 70.8000000000000000 |  24.0000000000000000
+     15 | 75.8000000000000000 |  37.6000000000000000
+     20 | 71.6000000000000000 |  26.8000000000000000
+     25 | 67.6000000000000000 |  56.0000000000000000
+     30 | 68.8000000000000000 |  90.2000000000000000
+     35 | 71.6000000000000000 |  88.8000000000000000
+     40 | 66.4000000000000000 |  81.2000000000000000
+     45 | 68.2000000000000000 | 106.0000000000000000
+(10 rows)
+```
 
 ---
 
@@ -365,15 +463,17 @@ compute the number of distinct items in a set, are not supported by continuous
 aggregates. However, many of these types of functions have parallelizable approximations
 that can provide quite accurate approximations of the actual result and would therefore
 be able to integrate seamlessly with the continuous aggregates project. We intend to
-implement a number of these aggregates.  
+implement a number of these aggregates.
 
 ---
 
 
+[fff-system]: https://en.wikipedia.org/wiki/FFF_system
 [sec-data-retention]: /using-timescaledb/data_retention#data-retention
 [postgres-materialized-views]: https://www.postgresql.org/docs/current/rules-materializedviews.html
 [api-continuous-aggs]:/api#continuous-aggregates
 [postgres-createview]: https://www.postgresql.org/docs/current/static/sql-createview.html
+[pg-func-stable]: https://www.postgresql.org/docs/current/static/sql-createfunction.html
 [time-bucket]: /api#time_bucket
 [api-continuous-aggs-create]: /api#continuous_aggregate-create_view
 [postgres-parallel-agg]:https://www.postgresql.org/docs/current/parallel-plans.html#PARALLEL-AGGREGATION
@@ -382,6 +482,7 @@ implement a number of these aggregates.
 [api-continuous-aggregate-stats]: /api#timescaledb_information-continuous_aggregate_stats
 [api-drop-chunks]: /api#drop_chunks
 [api-set-chunk-interval]: /api#set_chunk_time_interval
+[api-set-integer-now-func]: /api#set_integer_now_func
 [api-add-drop-chunks]: /api#add_drop_chunks_policy
 [timescale-github]: https://github.com/timescale/timescaledb
 [support-slack]: https://slack-login.timescale.com


using-timescaledb/continuous-aggregates.md
```diff
@@ -96,6 +96,120 @@ with PostgreSQL. This means that aggregates with `ORDER BY` clauses or `DISTINCT
 additionally we require that no `FILTER` clauses appear ([though we will allow them in future
 releases](#future-work)).
 
+### Integer-based Continuous Aggregates [](create-integer)
```
gayyappan:
```suggestion
###Continuous Aggregates for integer-based hypertables [](create-integer)
```

gayyappan:
```suggestion
### Continuous Aggregates for integer-based hypertables [](create-integer)
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -96,6 +96,120 @@ with PostgreSQL. This means that aggregates with `ORDER BY` clauses or `DISTINCT
...
+
+Usually, continuous aggregates are defined on a time column, but it is
```
gayyappan: Great example! The only suggestion is to move this under a new section of the tutorial like Advanced Topics/Advanced Users (or something similar).
This is fairly self-contained and is of interest only to certain users. Also, will not affect the current flow of the tutorial.  Maybe Best Practices and this section go under Advanced Topics?

I am not able to add a comment to unchanged parts of the doc. I see a section called Data retention for continuous aggregates. The paragraph about down sampling and ignore_invalidation_older_than (around line 281) could probably be moved to the Data retention section. I think it is a better fit there.

using-timescaledb/continuous-aggregates.md
```diff
@@ -96,6 +96,120 @@ with PostgreSQL. This means that aggregates with `ORDER BY` clauses or `DISTINCT
...
+Once the replacement for current timestamp has been set up, you can
+define the continuous aggregate for the `devices` table.
+
+```sql
+CREATE VIEW devices_summary
+WITH (
+  timescaledb.continuous,
+  timescaledb.refresh_interval='1 minutes',
+  timescaledb.refresh_lag = '-5')
+AS SELECT time_bucket('5', time) AS bucket,
```
mkindahl: @gayyappan One thing that is not entirely clear here is how `refresh_lag` is used together with the integer-now function, and I think this needs to be added. For inserted rows, are they added when

1. `row.time <= current_fortnight() + refresh_lag`
2. `row.time <= current_fortnight() - refresh_lag`.

The `refresh_lag` is negative here (it was copied from the test), which suggest 1, but usual use of refresh lag suggest 2.

using-timescaledb/continuous-aggregates.md
```diff
@@ -96,6 +96,120 @@ with PostgreSQL. This means that aggregates with `ORDER BY` clauses or `DISTINCT
...
+     95 | 68.6000000000000000 | 136.2000000000000000
+(20 rows)
+```
+
+If you want to refresh the view immediately, you can do it in the
+normal manner:
+
+```sql
+REFRESH MATERIALIZED VIEW devices_summary;
+```
```
mkindahl: @gayyappan When I do this, I get the following errors:

```
postgres=# refresh materialized view devices_summary;
INFO:  new materialization range for public.devices (time column time) larger than allowed in one run, truncating 1915095555 to 100
INFO:  materializing continuous aggregate public.devices_summary: nothing to invalidate, new range up to 100
WARNING:  REFRESH did not materialize the entire range since it was limited by the max_interval_per_job setting
REFRESH MATERIALIZED VIEW
```

Which seems mostly confusing and users are guaranteed to react to it and wonder what they have done wrong. Shall we add a notice to the documentation and/or change the error message?

using-timescaledb/continuous-aggregates.md
mkindahl: > I am not able to add a comment to unchanged parts of the doc. I see a section called Data retention for continuous aggregates. The paragraph about down sampling and ignore_invalidation_older_than (around line 281) could probably be moved to the Data retention section. I think it is a better fit there.

The PR for the additions to the data retention section is #314.

mkindahl: > Great example! The only suggestion is to move this under a new section of the tutorial like Advanced Topics/Advanced Users (or something similar).
> This is fairly self-contained and is of interest only to certain users. Also, will not affect the current flow of the tutorial. Maybe Best Practices and this section go under Advanced Topics?

Good idea. I'll add an "Advanced Topics" section. Note that this is not the tutorial, this is the "using" section of the documentation.

mkindahl: Moved section to new "Advanced Topics" section.

using-timescaledb/continuous-aggregates.md
mkindahl: Done.

using-timescaledb/continuous-aggregates.md
mkindahl: It appears to be 1, from `api.md` at least.

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +248,124 @@ hypertable with a continuous aggregate.
 
 ---
 
+### Advanced Topics [](using-cagg-advanced)
+
+#### Continuous Aggregates for Integer-Based Hypertables [](create-integer)
```
erimatnor:
```suggestion
#### Continuous Aggregates for Hypertables using Integer-Based Time [](create-integer)
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +248,124 @@ hypertable with a continuous aggregate.
...
+
+Usually, continuous aggregates are defined on a time column, but it is
+also possible to create your own custom scheme for handling
+aggregation for tables that are using an integer for the time
+column. This can be useful if you have tables that use other measures
```
erimatnor:
```suggestion
aggregation for tables that are using an integer time
column. This can be useful if you have tables that use other measures
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +248,124 @@ hypertable with a continuous aggregate.
 
 ---
 
+### Advanced Topics [](using-cagg-advanced)
+
+#### Continuous Aggregates for Integer-Based Hypertables [](create-integer)
+
+Usually, continuous aggregates are defined on a time column, but it is
```
erimatnor:
```suggestion
Usually, continuous aggregates are defined on a [date/time-type](https://www.postgresql.org/docs/current/datatype-datetime.html) column, but it is
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +248,124 @@ hypertable with a continuous aggregate.
...
+SELECT * FROM create_hypertable('devices', 'time',
+                                chunk_time_interval => 1000);
+```
+
+To define a continuous aggregate on a table that is using an integer
+dimension, it is necessary to have a function to get the current
+timestamp in whatever representation that you are using and set it for
+the hypertable using
+[`set_integer_now_func`][api-set-integer-now-func]. The function can
+be defined as a normal Postgres function, but need to be
```
erimatnor:
```suggestion
be defined as a normal PostgreSQL function, but need to be
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +248,124 @@ hypertable with a continuous aggregate.
...
+```
+
+Once the replacement for current timestamp has been set up, you can
+define the continuous aggregate for the `devices` table.
+
+```sql
+CREATE VIEW devices_summary
+WITH (timescaledb.continuous,
+      timescaledb.refresh_interval = '1 minute',
+      timescaledb.refresh_lag = '-500') AS
```
erimatnor: Unclear why we are using this `refresh_lag` setting. Doesn't this make the aggregate very inefficient (refresh always)?

Just worried that users are following this example because they don't understand these settings (I am myself confused by them) and then don't get the performance they expect.

It is also confusing that the `refresh_interval` and `refresh_lag` are using different time representations.

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +248,124 @@ hypertable with a continuous aggregate.
...
+CREATE EXTENSION tablefunc;
+
+INSERT INTO devices(time, cpu_usage, disk_usage)
+SELECT time,
+       normal_rand(1,70,10) AS cpu_usage,
+	   normal_rand(1,2,1) * (row_number() over()) AS disk_usage
+  FROM generate_series(1,10000) AS time;
+```
+
+You can now check that the view contain the correct data.
```
erimatnor:
```suggestion
You can now check that the view contains the correct data.
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +248,124 @@ hypertable with a continuous aggregate.
...
+
+```sql
+postgres=# SELECT * FROM devices_summary ORDER BY bucket;
```
erimatnor: Maybe trim the output, LIMIT to 10 rows or something.

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +248,124 @@ hypertable with a continuous aggregate.
...
+     90 | 71.6000000000000000 | 151.4000000000000000
+     95 | 68.6000000000000000 | 136.2000000000000000
+(20 rows)
+```
+
+If you want to refresh the view immediately, you can do it in the
+normal manner:
+
+```sql
+REFRESH MATERIALIZED VIEW devices_summary;
```
erimatnor: Note that this might be confusing because apparently this is designed to not be called directly. Also, it won't always refresh the view as one would expect (depends on your view settings). I guess it works in this case because you configured a `refresh_lag` of `-bucket`, which is not the recommended (or default) setting since it makes the agg inefficient.

I know this is not intuitive, but that's also why I am wondering if we shouldn't mention manual refresh since it currently doesn't do what you expect in all cases.

using-timescaledb/continuous-aggregates.md
mkindahl: If we do not set the `refresh_lag` the view will not be updated, which might give the appearance that the view is not working. A refresh lag set to this value just means that the view is updated for recent values, so AFAICT that would mean that the "most recent" value might not be accurate, but not affect performance.

Than the two parameters use different time representations does not seem strange to me. The refresh interval gives how often the refresh is executed and should be given in wall time, but the `refresh_lag` is referring to microfortnights in this example, so it makes sense that it is an integer.

using-timescaledb/continuous-aggregates.md
mkindahl: Done.

using-timescaledb/continuous-aggregates.md
mkindahl: It is already in the manual as a tip above, so we should not shy from mentioning it. If it does not behave as expected in all cases, we should document this, or fix the behavior, but I do not think that omitting mentioning it is a helpful for the users.

Also, I do not think that refresh lag affects performance, see comment above, but I might have misunderstood how it works.

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +248,114 @@ hypertable with a continuous aggregate.
 
 ---
 
+### Advanced Topics [](using-cagg-advanced)
+
+#### Continuous Aggregates for Hypertables using Integer-Based Time [](create-integer)
```
solugebefola:
```suggestion
#### Continuous Aggregates using Integer-Based Time [](create-integer)
```
'Hypertables' is implied here?

using-timescaledb/continuous-aggregates.md
```diff
@@ -370,10 +478,12 @@ implement a number of these aggregates.
 ---
 
 
+[fff-system]: https://dl.acm.org/doi/10.1145/1229899.1229919
```
solugebefola: Not sure this is the correct link?  Led me to:

Screen Shot 2020-03-09 at 3 35 06 PM

using-timescaledb/continuous-aggregates.md
```diff
@@ -382,6 +492,7 @@ implement a number of these aggregates.
 [api-continuous-aggregate-stats]: /api#timescaledb_information-continuous_aggregate_stats
 [api-drop-chunks]: /api#drop_chunks
 [api-set-chunk-interval]: /api#set_chunk_time_interval
+[api-set-integer-now]: /api#set_integer_now_func
```
solugebefola:
```suggestion
[api-set-integer-now-func]: /api#set_integer_now_func
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +248,114 @@ hypertable with a continuous aggregate.
...
+                                chunk_time_interval => 1000);
+```
+
+To define a continuous aggregate on a table that is using an integer
+dimension, it is necessary to have a function to get the current
+timestamp in whatever representation that you are using and set it for
+the hypertable using
+[`set_integer_now_func`][api-set-integer-now-func]. The function can
+be defined as a normal PostgreSQL function, but need to be
+[`STABLE`][pg-func-stable], take no arguments, and should return an
```
solugebefola:
```suggestion
be defined as a normal PostgreSQL function, but needs to: be
[`STABLE`][pg-func-stable], take no arguments, and return an
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +248,114 @@ hypertable with a continuous aggregate.
...
+```sql
+CREATE VIEW devices_summary
+WITH (timescaledb.continuous,
+      timescaledb.refresh_interval = '1 minute',
+      timescaledb.refresh_lag = '-500') AS
+SELECT time_bucket('500', time) AS bucket,
+       avg(cpu_usage) AS avg_cpu,
+       avg(disk_usage) AS avg_disk
+     FROM devices
+    GROUP BY bucket;
```
solugebefola:
```suggestion
     GROUP BY bucket;
```

solugebefola:
Also, you should put an anchor to this section in `page-index.js`

using-timescaledb/continuous-aggregates.md
mkindahl: It is actually the right link, but maybe I should pick another one instead that is not a PDF.

mkindahl: Picked the Wikipedia page instead.

using-timescaledb/continuous-aggregates.md
erimatnor: In other places in our docs, we state there are caveats with setting the refresh_lag like this and it is not the default setting. For instance, our docs says: `Tuning the refresh_lag parameter lower will mean that the aggregates will follow inserts more closely, but can cause some write amplification which may degrade insert performance. `. And, further, the example setting here is not only a "magic number" but also negative!, which I think is confusing for the reader. So, in this example we're basically not only ignoring our own best practices, but we're also confusing the user with very non-intuitive settings.

With default settings, isn't it only the most recent bucket that is not updated? Can't we handle that by explaining in the tutorial that this is how continuous aggs work?

I think it is better to explain the recommended behavior than trying to circumvent it ourselves.

using-timescaledb/continuous-aggregates.md
erimatnor: The docs says that an aggressive refresh_lag affects insert performance and increases write amplification. (See above).

If refresh_lag would not affect performance, then there is no reason for that parameter to exist at all since one would always prefer no lag.

using-timescaledb/continuous-aggregates.md
erimatnor: I am confused by this. What is the purpose of this link?

mkindahl:
> Also, you should put an anchor to this section in `page-index.js`

I did an attempt at that, but the "Advanced Topics" now contain two sub-sections and I am not sure how that works in `page-index.js` and if we should have it. Please advice.

using-timescaledb/continuous-aggregates.md
mkindahl: AFAICT, write amplification does not automatically lead to lower performance, but it does increase utilization. However, I still think that we should keep the `REFRESH MATERIALIZED VIEW` given that it is already discussed elsewhere in the documentation.

erimatnor: Ok, let's keep `REFRESH MATERIALIZED VIEW`. Fwiw, I think this command _should_ be a first-class command for users since it is for regular views. It's just that I've heard some people say it is not supposed to be used (and it  prints a lot of cryptic info) and semantics of it is currently confusing since it doesn't completely refresh the view like you would expect.

gayyappan: @mkindahl I think this is the REFRESH ... that you referring to:
_:TIP: Most times the continuous aggregate view will be updated by the background job; however, if you would like to run it yourself, you may use the REFRESH MATERIALIZED VIEW command._

should be rephrased as : 
:TIP: The continuous aggregate view will be updated by the background job; however, if you would like to verify if the view is working correctly, you may use the REFRESH MATERIALIZED VIEW command.

mkindahl: Fixed the tip, but if `alter_job_schedule` is going to be used, you need to dig out the `job_id`, which is not obvious how to do, so I would say that we either keep the `REFRESH` even though it is not perfect, or remove it entirely.

I think it is better to keep it since it does ensure that the view shows *something*, even it if is not perfect. The usage in the tutorial is more reminiscent of what you would do when debugging (just see that it works, don't really care about the exact figures).

gayyappan: If you are keeping REFRESH, please add a note that this is generally how we manually check if the view is working and not running into any errors.The REFRESH MATERIALIZED VIEW will behave like  a _single run_ of the (automatically run) background worker job. (I don't think we actually say this anywhere else).
Please modify the verbiage around this paragraph because it could be interpreted as the entire view is getting updated

_If you want to refresh the
continuous aggregate view immediately, you can use [`REFRESH
MATERIALIZED VIEW`][api-refresh-continuous-aggs]:_

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +246,117 @@ hypertable with a continuous aggregate.
 
 ---
 
+### Advanced Topics [](advanced-usage)
+
+#### Continuous Aggregates using Integer-Based Time [](create-integer)
+
+Usually, continuous aggregates are defined on a [date/time-type](https://www.postgresql.org/docs/current/datatype-datetime.html) column, but it is
```
solugebefola: Can you make this into a footnote-style reference please

using-timescaledb/continuous-aggregates.md
```diff
@@ -181,9 +181,7 @@ SELECT drop_chunks(INTERVAL '30 days', 'device_readings')
 You can read more about data retention with continuous aggregates in
 the [*Data retention*][sec-data-retention] section.
 
->:TIP: Most times the continuous aggregate view will be updated by the background job;
-  however, if you would like to run it yourself, you may use the
-  [`REFRESH MATERIALIZED VIEW` command][api-refresh-continuous-aggs].
+>:TIP: Most times the continuous aggregate view will be updated by the background job; however, if you would like verify that the view is working correctly, you may use the [`REFRESH MATERIALIZED VIEW` command][api-refresh-continuous-aggs].
```
erimatnor: This is a bit misleading IMO. The background job to refresh the view can be failing even if a manual refresh updates the view. So, I don't think a manual refresh is good  indication of the automation working.

using-timescaledb/continuous-aggregates.md
erimatnor: As per my other comment, I think saying that you can use a manual refresh as a method of checking if things are working is highly misleading. A background job could be failing or missing even if a manual refresh works.

gayyappan: @erminator If the manual refresh works, the background job that does the automatic refresh should work as well.i.e to say if manual refresh does not work => bgw job will fail. (Yes, there are other cases where bgw jobs  could fail but they are related to the job scheduling and not to the materialization itself). We have a materialization bug if the bgw job materialization fails but the refresh succeeds

using-timescaledb/continuous-aggregates.md
mkindahl: Done.

using-timescaledb/continuous-aggregates.md
mkindahl: Removed the tip.

using-timescaledb/continuous-aggregates.md
mkindahl: Removed all text mentioning the `REFRESH MATERIALIZED VIEW`.

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +246,117 @@ hypertable with a continuous aggregate.
...
+```
+
+Note that the continuous aggregate will lag behind the hypertable 500
+microfortnights (since the bucket size is 500 microfortnights) so you
+have to wait for it to be updated. If you want to refresh the
+continuous aggregate view immediately, you can use [`REFRESH
+MATERIALIZED VIEW`][api-refresh-continuous-aggs]:
+
+```sql
+REFRESH MATERIALIZED VIEW devices_summary;
```
gayyappan: There are still references to REFRESH MATERIALIZED VIEW on lines 350-355.

using-timescaledb/continuous-aggregates.md
k-rus: The rest of PR uses multiple lines if sentences or paragraphs are too long, which is nice. Can it be fixed to into multilines here too, so it will be as originally and possible to track changes?

k-rus: I don't understand how calling the refresh will verify that cagg bgw works correctly. Thus I feel the current formulation is misleading.

page-index/page-index.js
```diff
@@ -383,9 +383,9 @@ const pageIndex = [
                         type: ANCHOR,
                         href: "#dropping-data"
                     }, {
-                        Title: "Best Practices",
```
k-rus: Best practices section is still in the document. Should Advanced Topics be added to the index instead of replacing?

using-timescaledb/continuous-aggregates.md
mkindahl: I removed this paragraph for the time being since we consider `REFRESH MATERIALIZED VIEW` a debug command, but not sure what alternative we have for testing the view right now, so it might re-surface.

mkindahl: Because running `REFRESH MATERIALIZED VIEW` will dispatch a job to update the view, so it will indirectly check that the view works. Several users have stumbled over that they set up the view and then don't see it populated, so we have to direct them to use `REFRESH MATERIALIZED VIEW`.

using-timescaledb/continuous-aggregates.md
mkindahl: Still not sure how to handle this. We are considering `REFRESH MATERIALIZED VIEW` as a debug command, but users will run into the issue that the materialization has not executed, so how should we direct them to having it updated?

mkindahl: I removed this for now, as a separate commit, so that I can restore it if necessary.

page-index/page-index.js
mkindahl: Good point. Added.

tutorials/continuous-aggs-tutorial.md
```diff
@@ -156,10 +156,6 @@ aggregates only for the data that satisfies the condition:
 `time_bucket('1h', pickup_datetime) < max(pickup_time) - '1h'` (if the
 `refresh_lag` is set to 1 hour)
 
-Both `refresh_lag` or `refresh_interval` can be set to an integer
-value, which is then interpreted as the number of microseconds for the
-lag or interval respectively.
-
 To keep the continuous aggregate up-to-date,
 you can set `timescaledb.refresh_lag` to a negative value. This will allow the continuous aggregate
```
gayyappan, line 160: This part needs to be revisited.
1) I believe we no longer need a negative value to keep this up to date. Using a value of 0 should suffice. But needs a test before updating the docs. (Since 1.6 the lag is calculated against the  current time as returned by now() )
2) Also, materialized_only setting will help with this. 
Perhaps the best approach is to update this section along with the materialized_only changes.

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +244,111 @@ hypertable with a continuous aggregate.
...
+column. This can be useful if you have tables that use other measures
+of time that can be represented as integer values, such as nanosecond
+epochs, minutes since founding date, or whatever is suitable for your
+application.
+
+As an example, suppose that you have a table with CPU and disk usage
+for some devices where time is measured in
+[microfortnights][fff-system]. Since you are using an integer-valued
+column as time, you need to provide the chunk time interval when
+creating the hypertable. In this case, let each chunk consist of a
```
k-rus, line 263: I guess you meant 1000 millifortnights, otherwise the sentence is repeating the first sentence:

```suggestion
creating the hypertable. In this case, let each chunk consist of 1000
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +244,111 @@ hypertable with a continuous aggregate.
...
+millifortnight (which is about 20 minutes).
```
k-rus, line 264: Then it will be:
```suggestion
millifortnights and each millifortnight is about 20 minutes.
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +244,111 @@ hypertable with a continuous aggregate.
...
+  cpu_usage INTEGER,  -- Total CPU usage
+  disk_usage INTEGER, -- Total disk usage
+  PRIMARY KEY (time)
+);
+
+SELECT create_hypertable('devices', 'time',
+                         chunk_time_interval => 1000);
+```
+
+To define a continuous aggregate on a table that is using an integer
```
erimatnor, line 278:
```suggestion
To define a continuous aggregate on a hypertable that is using an integer
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +244,111 @@ hypertable with a continuous aggregate.
...
+dimension, it is necessary to have a function to get the current
```
k-rus: I think it is good to mention that it is `time dimension` and `integer` is a type of it.
```suggestion
To define a continuous aggregate on a table that is using an integer
time dimension, it is necessary to have a function to get the current
```
or 
```suggestion
To define a continuous aggregate on a table that is using an integer for
the time dimension, it is necessary to have a function to get the current
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +244,111 @@ hypertable with a continuous aggregate.
...
+  PRIMARY KEY (time)
+);
+
+SELECT create_hypertable('devices', 'time',
+                         chunk_time_interval => 1000);
+```
+
+To define a continuous aggregate on a hypertable that is using an integer
+dimension, it is necessary to have a function to get the current
+timestamp in whatever representation that you are using and set it for
```
k-rus, line 280: It might be just personal opinion:
```suggestion
dimension, it is necessary to have a function to get the current
time in whatever representation that you are using and set it for
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +244,111 @@ hypertable with a continuous aggregate.
...
+the hypertable using
+[`set_integer_now_func`][api-set-integer-now-func]. The function can
+be defined as a normal PostgreSQL function, but needs to: be
+[`STABLE`][pg-func-stable], take no arguments, and return an integer
+of the same type as the time column in the table. In our case, this
```
k-rus, line 285:
```suggestion
[`STABLE`][pg-func-stable], take no arguments, and return an integer value
of the same type as the time column in the table. In our case, this
```
Or just value.

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +244,111 @@ hypertable with a continuous aggregate.
...
+should suffice:
```
k-rus, line 286: Will it be better to move the fact that millifortnight is about 20 mins here, since this fact is used in calculating current time?

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +244,111 @@ hypertable with a continuous aggregate.
...
+CREATE OR REPLACE FUNCTION current_microfortnight() RETURNS BIGINT
+LANGUAGE SQL STABLE AS $$
+	SELECT CAST(1209600 * EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) / 1000000 AS BIGINT)
+$$;
+
+SELECT set_integer_now_func('devices', 'current_microfortnight');
+```
+
+Once the replacement for current timestamp has been set up, you can
+define the continuous aggregate for the `devices` table.
```
k-rus, line 298: I guess it should be `a` continuous aggregate, since it (its shape) wasn't described before.
```suggestion
define a continuous aggregate for the `devices` table.
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +244,111 @@ hypertable with a continuous aggregate.
...
+
+INSERT INTO devices(time, cpu_usage, disk_usage)
+SELECT time,
+       normal_rand(1,70,10) AS cpu_usage,
+	   normal_rand(1,2,1) * (row_number() over()) AS disk_usage
+  FROM generate_series(1,10000) AS time;
+```
+
+>:TIP: You can use the `tablefunc` extension to generate a Gaussian
+>distribution and use the `row_number` function to turn it into a
```
k-rus, line 324: I believe `normal distribution` is more commonly used and it matches the name of SQL function, which will be useful for those users, who are not familiar with statistics.
```suggestion
>:TIP: You can use the `tablefunc` extension to generate a normal
>distribution and use the `row_number` function to turn it into a
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +244,111 @@ hypertable with a continuous aggregate.
...
+     30 | 68.8000000000000000 |  90.2000000000000000
+     35 | 71.6000000000000000 |  88.8000000000000000
+     40 | 66.4000000000000000 |  81.2000000000000000
+     45 | 68.2000000000000000 | 106.0000000000000000
+(10 rows)
+```
+
+Note that the continuous aggregate will lag behind the hypertable 500
+microfortnights (since the bucket size is 500 microfortnights) so you
+have to wait for it to be updated.
```
k-rus, line 348: Why? It is the doc for 1.7, where real time aggregates are added. Should the query produce the complete result?

tutorials/continuous-aggs-tutorial.md
mkindahl: I removed the paragraph. We are describing the effect of setting negative values in `refresh_lag` in the users part of the manual. Since the continuous aggregate is now update to date I do not think we need to describe this "trick" in the tutorial.

using-timescaledb/continuous-aggregates.md
mkindahl: I'm talking about timestamp everywhere else and am using the `CURRENT_TIMESTAMP` to get the epoch, so I think that "timestamp" is more consistent.

using-timescaledb/continuous-aggregates.md
mkindahl: The chunks are one millifortnight, not 1000.

mkindahl:
> the section on keeping the view up to date needs some changes. I leave it you to either fix it as part of this PR or in another PR. Will approve based on what you decide.

I just removed the paragraph. It is described in detail in the "using" chapter, so I do not think we need to discuss it here.

using-timescaledb/continuous-aggregates.md
k-rus: Section `Using Continuous Aggregates` in this file uses `current time`, so `current timestamp` is not consistent with the rest of the document. I don't see that you use `timestamp` in the text, except here. `current time` is easier to follow and will be consistent with the rest of the document.

using-timescaledb/continuous-aggregates.md
```diff
@@ -248,6 +244,108 @@ hypertable with a continuous aggregate.
...
+SELECT create_hypertable('devices', 'time',
+                         chunk_time_interval => 1000);
+```
+
+To define a continuous aggregate on a hypertable that is using an integer
+dimension, it is necessary to have a function to get the current
+timestamp in whatever representation that you are using and set it for
+the hypertable using
+[`set_integer_now_func`][api-set-integer-now-func]. The function can
+be defined as a normal PostgreSQL function, but needs to: be
```
k-rus, line 284: Typo:
```suggestion
be defined as a normal PostgreSQL function, but needs to be
```

using-timescaledb/continuous-aggregates.md
mkindahl: Good point. Switched to use `current_time` and removing uses of "timestamp" in this commit. There are still some mention of timestamps in the chapter, but that is for a separate PR.

using-timescaledb/continuous-aggregates.md
mkindahl: Done.

PR #326, opened 2020-03-04
Remove enterprise tag

mkindahl:
The `add_drop_chunks_policy` is no longer an enterprise feature, so the
enterprise feature tag is removed from the section that describes
automatic data retention policies.

### DIFF
--- a/api.md
+++ b/api.md
@@ -1341,7 +1341,7 @@ aggregate, such as views that are built on top of the continuous aggregate view.
 
 
 ---
-## Automation policies :enterprise_function: [](automation-policies)
+## Automation policies [](automation-policies)
 
 TimescaleDB includes an automation framework for allowing background tasks to
 run inside the database, controllable by user-supplied policies. These tasks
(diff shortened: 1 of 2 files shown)


mkindahl:
> we also have references to enterprise edition in the API docs. I think move_chunks is enterprise but everything else is non-enterprise. @erimatnor could you confirm?

This is what he told me before. Removed all enterprise tags except on move chunks. PTAL.

mkindahl:
> Looks good to me.
> should check with someone who knows whether changes to timescaledb_information.license (api.md) are needed?

Will do. Would like to have approval from @erimatnor before pushing as well.

mkindahl:
> > Looks good to me.
> > should check with someone who knows whether changes to timescaledb_information.license (api.md) are needed?
> 
> Will do. Would like to have approval from @erimatnor before pushing as well.

We need a separate discussion about any changes regarding the `license` view, so no changes in this PR.

mkindahl:
> I guess you did a search in the docs for this tag? The only feature that should have it is move chunks.

Yes, I searched all of the docs:

```bash
$ find . -name '*.md' -exec grep ':enterprise_function:' {} +
./api.md:## move_chunk() :enterprise_function: [](move_chunk)
./README.md:    + Adding `:enterprise_function:` to a header adds decorator text "enterprise function".
./using-timescaledb/move_chunk.md:### Move Chunks :enterprise_function: [](move)
```

Issue #334, opened 2020-03-11
Documentation changes for 1.6.1


mkindahl:
The following changes need to be documented for 1.6.1

- [x] Changes to `drop_chunks` (see timescale/timescaledb#1666)
- [x] Changes to `add_drop_chunks_policy` (see timescale/timescaledb#1666)

mkindahl:
Closing issue since 1.6.1 is released.

PR #335, opened 2020-03-11
Updating documentation for drop_chunks

mkindahl:
Updating documentation for `drop_chunks` to match what we will release
in 1.6.1.

Subtask of #334

### DIFF
--- a/api.md
+++ b/api.md
@@ -574,16 +574,18 @@ the same semantics as the `show_chunks` [function](#show_chunks).
 |Name|Description|
 |---|---|
 | `older_than` | Specification of cut-off point where any full chunks older than this timestamp should be removed. |
+| `table_name` | Hypertable or continuous aggregate from which to drop chunks.
 | `newer_than` | Specification of cut-off point where any full chunks newer than this timestamp should be removed. |
 
 #### Optional Arguments [](drop_chunks-optional-arguments)
 
 |Name|Description|
 |---|---|
-| `table_name` | Hypertable name from which to drop chunks. If not supplied, all hypertables are affected.
 | `schema_name` | Schema name of the hypertable from which to drop chunks. Defaults to `public`.
 | `cascade` | Boolean on whether to `CASCADE` the drop on chunks, therefore removing dependent objects on chunks to be removed. Defaults to `FALSE`.
 | `cascade_to_materializations` | Set to `TRUE` to also remove chunk data from any associated continuous aggregates. Set to `FALSE` to only drop raw chunks (while keeping data in the continuous aggregates). Defaults to `NULL`, which errors if continuous aggregates exist.|
+
+>:TIP: The `table_name` argument was optional in previous versions, but this is now deprecated: `table_name` should always be given.
 
 The `older_than` and `newer_than` parameters can be specified in two ways:
 
@@ -611,15 +613,15 @@ intersection between two ranges will result in an error.
 
 #### Sample Usage [](drop_chunks-examples)
 
-Drop all chunks older than 3 months ago:
+Drop all chunks older than 3 months ago from hypertable `conditions`:
 ```sql
-SELECT drop_chunks(interval '3 months');
+SELECT drop_chunks(interval '3 months', 'conditions');
 ```
 
 Example output:
 
 ```sql
-              drop_chunks               
+              drop_chunks
 ----------------------------------------
  _timescaledb_internal._hyper_3_5_chunk
  _timescaledb_internal._hyper_3_6_chunk
@@ -629,14 +631,9 @@ Example output:
 (5 rows)
 ```
 
-Drop all chunks more than 3 months in the future. This is useful for correcting data ingested with incorrect clocks:
+Drop all chunks more than 3 months in the future from hypertable `conditions`. This is useful for correcting data ingested with incorrect clocks:
 ```sql
-SELECT drop_chunks(newer_than => now() + interval '3 months');
-```
-
-Drop all chunks from hypertable `conditions` older than 3 months:
-```sql
-SELECT drop_chunks(interval '3 months', 'conditions');
+SELECT drop_chunks(newer_than => now() + interval '3 months', table_name => 'conditions');
 ```
 
 Drop all chunks from hypertable `conditions` before 2017:
@@ -654,12 +651,7 @@ Drop all chunks from hypertable `conditions` older than 3 months, including depe
 SELECT drop_chunks(interval '3 months', 'conditions', cascade => TRUE);
 ```
 
-Drop all chunks newer than 3 months ago:
-```sql
-SELECT drop_chunks(newer_than => interval '3 months');
-```
-
-Drop all chunks older than 3 months ago and newer than 4 months ago:
+Drop all chunks older than 3 months ago and newer than 4 months ago from hypertable `conditions`:
 ```sql
 SELECT drop_chunks(older_than => interval '3 months', newer_than => interval '4 months', table_name => 'conditions')
 ```
@@ -960,7 +952,7 @@ SELECT move_chunk(
   index_destination_tablespace => 'tablespace_3',
   reorder_index => 'conditions_device_id_time_idx',
   verbose => TRUE
-);  
+);
 ```
 
 ---
@@ -1118,7 +1110,7 @@ and the system will recompress your chucks.
 Decompress a single chunk
 
 ``` sql
-SELECT decompress_chunk('_timescaledb_internal._hyper_2_2_chunk');  
+SELECT decompress_chunk('_timescaledb_internal._hyper_2_2_chunk');
 ```
 
 ---


api.md
```diff
@@ -580,11 +580,33 @@ the same semantics as the `show_chunks` [function](#show_chunks).
 
 |Name|Description|
 |---|---|
-| `table_name` | Hypertable name from which to drop chunks. If not supplied, all hypertables are affected.
+| `table_name` | Hypertable name or continuous aggregate view from which to drop chunks. This is optional, but see below.
 | `schema_name` | Schema name of the hypertable from which to drop chunks. Defaults to `public`.
 | `cascade` | Boolean on whether to `CASCADE` the drop on chunks, therefore removing dependent objects on chunks to be removed. Defaults to `FALSE`.
 | `cascade_to_materializations` | Set to `TRUE` to also remove chunk data from any associated continuous aggregates. Set to `FALSE` to only drop raw chunks (while keeping data in the continuous aggregates). Defaults to `NULL`, which errors if continuous aggregates exist.|
 
+The `table_name` is optional and can be used to remove chunks from all
```
solugebefola: Not quite sure what you are trying to say here?  Does adding `table_name` remove chunks from _all_ hypertables in the database?  The argument description suggests it's used to drop chunks on a specific hypertable.  Or do you mean that `table_name` can be used for _any_ hypertable in the database?

solugebefola: Or are you saying that leaving out `table_name` will drop chunks, but only under the following circumstances?

mkindahl: Ah, no, sorry my fault. You can remove all chunks in a database if you *don't* provide a table name. Will fix the wording.

api.md
```diff
@@ -580,11 +580,33 @@ the same semantics as the `show_chunks` [function](#show_chunks).
 
 |Name|Description|
 |---|---|
-| `table_name` | Hypertable name from which to drop chunks. If not supplied, all hypertables are affected.
+| `table_name` | Hypertable name or continuous aggregate view from which to drop chunks. This is optional, but see below.
```
gayyappan: @mkindahl Could you remove this line. Lets not draw attention to this. Even though it often works , it causes weird unintelligible errors when there are tables with different datatypes for time column.

gayyappan: __
```suggestion
| `table_name` | Hypertable name or continuous aggregate view from which to drop chunks.
```

api.md
gayyappan:
```suggestion

```
Please remove the line for reasons stated above. Would rather not have users trying this out without table_name.

api.md
```diff
@@ -580,11 +580,33 @@ the same semantics as the `show_chunks` [function](#show_chunks).
...
 | `schema_name` | Schema name of the hypertable from which to drop chunks. Defaults to `public`.
 | `cascade` | Boolean on whether to `CASCADE` the drop on chunks, therefore removing dependent objects on chunks to be removed. Defaults to `FALSE`.
 | `cascade_to_materializations` | Set to `TRUE` to also remove chunk data from any associated continuous aggregates. Set to `FALSE` to only drop raw chunks (while keeping data in the continuous aggregates). Defaults to `NULL`, which errors if continuous aggregates exist.|
 
+The `table_name` is optional and can be used to remove chunks from all
+hypertables in the database under the following circumstances:
+
+- All hypertables in the database use either a timestamp-based or
```
gayyappan: sorry, but lets remove this as well from the API. Just shows we have a bad API.

api.md
mkindahl: Yes, and users will run into them. Isn't it better to explain what is going on?

gayyappan: I strongly think it is better to point them to correct API usage. We don't document it as optional even though the API might work correctly in some cases. But in any non-trivial setup, users will run into issues. (Don't think we actually have many people using this with optional table_name, otherwise would have had bugs reported).

mkindahl: I think it is better to give users something to search for and find a solution, but I agree that we cannot make the option required (because of backwards compatibility) and it does not make sense to support it going forwards, so we should softly deprecate it.

I added a section under `drop_chunks` with known issues and outlined what does not work and what the solution is.

api.md
mkindahl: Fixed. The text was moved to a "known issues" section instead, since it is a slightly cumbersome interface.

api.md
mkindahl: I think it is not a big concern, but I removed the section anyway.

api.md
mkindahl: I added a tip that omitting the table name is deprecated and that the table name should be given.

mkindahl: Removed the section.

api.md
```diff
@@ -580,11 +580,13 @@ the same semantics as the `show_chunks` [function](#show_chunks).
 
 |Name|Description|
 |---|---|
-| `table_name` | Hypertable name from which to drop chunks. If not supplied, all hypertables are affected.
+| `table_name` | Hypertable name or continuous aggregate view from which to drop chunks.
 | `schema_name` | Schema name of the hypertable from which to drop chunks. Defaults to `public`.
 | `cascade` | Boolean on whether to `CASCADE` the drop on chunks, therefore removing dependent objects on chunks to be removed. Defaults to `FALSE`.
 | `cascade_to_materializations` | Set to `TRUE` to also remove chunk data from any associated continuous aggregates. Set to `FALSE` to only drop raw chunks (while keeping data in the continuous aggregates). Defaults to `NULL`, which errors if continuous aggregates exist.|
 
+>:TIP: The `table_name` argument was optional in previous versions, but this is now deprecated: `table_name` should always be given.
```
k-rus: Since hypertable name should be always given, all examples should use it. Otherwise, this statement is incorrect.

api.md
```diff
@@ -1376,12 +1378,7 @@ one drop-chunks policy may exist per hypertable.
 | `if_not_exists` | (BOOLEAN) Set to true to avoid throwing an error if the policy already exists. A notice is issued instead. Defaults to false. |
 | `cascade_to_materializations` | (BOOLEAN) Set to `TRUE` to also remove chunk data from any associated continuous aggregates. Set to `FALSE` to only drop raw chunks (while keeping data in the continuous aggregates). Defaults to `NULL`, which errors if continuous aggregates exist.|
 
->:TIP: When dropping data from the raw hypertable while retaining data
-on a continuous aggregate, the `older_than` parameter to `drop_chunks`
-has to be longer than the `timescaledb.ignore_invalidation_older_than`
-parameter on the continuous aggregate. That is because we cannot
-process invalidations on data regions where the raw data has been
-dropped.
+>:TIP: When dropping data from the raw hypertable while retaining data on a continuous aggregate, the `older_than` parameter to `drop_chunks` has to be longer than the `timescaledb.ignore_invalidation_older_than` parameter on the continuous aggregate. That is because we cannot process invalidations on data regions where the raw data has been dropped.
```
k-rus: Can you split the paragraph in multiple lines as it was in origin? I think it is preferable approach except content placed in tables. Or was any specific reason to remove new lines for the paragraph?

api.md
mkindahl: Good point. Fixed.

api.md
svenklemm: yes please split lines here

mkindahl: It shows up very strange in my editor. Reverted the hunk.

mkindahl: Also note that there are several very long lines elsewhere in the text, so I do not think keeping lines short can be a requirement or even a guideline.

svenklemm: its not enforced anywhere but they should be kept short to make reviews easier

mkindahl:
> I doubt that a hypertable name can be put as the first argument in `drop_chunks` due to its SQL definition. @mkindahl Have you tested updated examples?

No, it doesn't work.Fixed.

api.md
```diff
@@ -629,14 +631,9 @@ Example output:
 (5 rows)
 ```
 
-Drop all chunks more than 3 months in the future. This is useful for correcting data ingested with incorrect clocks:
-```sql
-SELECT drop_chunks(newer_than => now() + interval '3 months');
-```
-
-Drop all chunks from hypertable `conditions` older than 3 months:
+Drop all chunks more than 3 months in the future in table `conditions`. This is useful for correcting data ingested with incorrect clocks:
```
k-rus:
```suggestion
Drop all chunks more than 3 months in the future in hypertable `conditions`. This is useful for correcting data ingested with incorrect clocks:
```

api.md
```diff
@@ -654,12 +651,7 @@ Drop all chunks from hypertable `conditions` older than 3 months, including depe
 SELECT drop_chunks(interval '3 months', 'conditions', cascade => TRUE);
 ```
 
-Drop all chunks newer than 3 months ago:
-```sql
-SELECT drop_chunks(newer_than => interval '3 months');
-```
-
-Drop all chunks older than 3 months ago and newer than 4 months ago:
+Drop all chunks older than 3 months ago and newer than 4 months ago in table `conditions`:
```
k-rus:
```suggestion
Drop all chunks older than 3 months ago and newer than 4 months ago in hypertable `conditions`:
```

mkindahl:
> Updated text says:
> 
> > Drop chunks **in** hypertable
> 
> While old text uses **from**.
> 
> Should it be consistent? E.g.:
> 
> > drop chunks **from** hypertable

It is actually not very consistent in the document either. "Chunks in" and "chunks from" are used interchangeably. I'm reluctant to incorporate such changes in this PR since it will just generate long discussions. Let's take it separately.

api.md
```diff
@@ -580,11 +580,13 @@ the same semantics as the `show_chunks` [function](#show_chunks).
 
 |Name|Description|
 |---|---|
-| `table_name` | Hypertable name from which to drop chunks. If not supplied, all hypertables are affected.
+| `table_name` | Hypertable or continuous aggregate from which to drop chunks.
 | `schema_name` | Schema name of the hypertable from which to drop chunks. Defaults to `public`.
 | `cascade` | Boolean on whether to `CASCADE` the drop on chunks, therefore removing dependent objects on chunks to be removed. Defaults to `FALSE`.
 | `cascade_to_materializations` | Set to `TRUE` to also remove chunk data from any associated continuous aggregates. Set to `FALSE` to only drop raw chunks (while keeping data in the continuous aggregates). Defaults to `NULL`, which errors if continuous aggregates exist.|
 
+>:TIP: The `table_name` argument was optional in previous versions, but this is now deprecated: `table_name` should always be given.
```
solugebefola: Should `table_name` be in Required Arguments, then?

solugebefola:
One clarification, but up to you whether it needs to be corrected

api.md

mkindahl: > All original examples in `drop_chunks` section use **from**. Thus I suggest to be consistent and use it for now.

I replaced what I think are the one you mean.

PR #339, opened 2020-03-15
Expand examples with alter_job_schedule

mfreed:
Fix docs bug where query against reorder_policies must cast hypertable table by
::regclass.  Also provide additional examples of altering job schedule for
continuous aggregates, and expressly specifying the next_start time.


solugebefola:
Label with version please

PR #340, opened 2020-03-17
Moving updates for 1.6.1 to master branch

mkindahl:


api.md
```diff
@@ -1361,25 +1353,26 @@ are meant to implement data retention or perform tasks that will improve query
 performance on older chunks. Each policy is assigned a scheduled job
 which will be run in the background to enforce it.
 
-## add_drop_chunks_policy() [](add_drop_chunks_policy)
+## add_drop_chunks_policy() :enterprise_function: [](add_drop_chunks_policy)
```
solugebefola: I think this is no longer an enterprise function?  Confirmation @mfreed ?

mkindahl: It should not be an enterprise function. Only move chunks. Fixed.

PR #362, opened 2020-04-15
Describe how to drop chunks from aggregate

mkindahl:
The description in data retention is focused on dropping chunks from
the raw table, but it is also interesting to drop the chunks from the
materialized table but keep them in the raw data. This commit add text
about how to handle this case.

### DIFF
--- a/using-timescaledb/data-retention.md
+++ b/using-timescaledb/data-retention.md
@@ -111,6 +111,20 @@ SELECT drop_chunks('30 days'::interval, 'conditions',
                    cascade_to_materialization => FALSE);
 ```
 
+Dropping chunks from the materialized view is similar to dropping
+chunks from the tables with raw data. You can drop the chunks of the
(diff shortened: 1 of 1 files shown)


using-timescaledb/data-retention.md
```diff
@@ -111,6 +111,20 @@ SELECT drop_chunks('30 days'::interval, 'conditions',
...
+continuous aggregate using `drop_chunks`, but you need to set
+`ignore_invalidation_older_than` to ensure that new data outside the
+dropped region does not update the materialized table.
+
+```sql
+ALTER VIEW conditions_summary_daily SET (
+   timescaledb.ignore_invalidation_older_than = '29 days'
+);
+
+SELECT drop_chunks(INTERVAL '30 days', 'conditions_summary_daily);
```
svenklemm, line 125: missing `'` after the cagg name

mkindahl: FIxed.

PR #363, opened 2020-04-15
Add data node instance requirements

mkindahl:
The instance holding the data node need to have a compatible version of
the extension installed, so added a short text highlighting this.

Documentation for #timescale/timescaledb-private#723


mkindahl:
It is necessary to add a more complete description of how to configure the data node instance, but this commit just highlights that the extension need to be available on the instance holding the data node.

mkindahl:
Documentation for timescale/timescaledb-private#581 timescale/timescaledb#1662

PR #364, opened 2020-04-15
Add Quick Starts for Python and Ruby

CoolAssPuppy:

PR #358, opened 2020-04-09
Added more Grafana tutorials

CoolAssPuppy:

PR #371, opened 2020-04-20
[fix]Update mislabelled code blocks

solugebefola:
Code blocks had mislabelled languages, which caused the pages not
to build


solugebefola:
From #358 #364

PR #374, opened 2020-04-22
Add PostgreSQL 12 as installation option

### LINKED ISSUE #373: Missing link for PG12 in installation page?
Page: https://docs.timescale.com/latest/getting-started/installation/windows/installation-windows

Seems the default version is updated to 1.7, and https://timescalereleases.blob.core.windows.net/windows/timescaledb-postgresql-12_1.7.0-windows-amd64.zip is already available. But there is no option to choose `PostgreSQL version:12`.

Is it just forget to add the new version, or support for pg12 in 1.7.0 is still in beta stage?

k-rus:
1.7.0 includes binaries for PG 12, thus the option should be available to users.

Fixes #373

### DIFF
--- a/page-index/page-index.js
+++ b/page-index/page-index.js
@@ -47,7 +47,7 @@ const pageIndex = [
                 Title: "Installing",
                 type: REACT_PAGE,
                 href: "installation",
-                options: {pg_version: ["11", "10", "9.6"]},
+                options: {pg_version: ["12", "11", "10", "9.6"]},
                 component: "InstallationPage",
                 children: [
                     {


page-index/page-index.js
```diff
@@ -47,7 +47,7 @@ const pageIndex = [
                 Title: "Installing",
                 type: REACT_PAGE,
                 href: "installation",
-                options: {pg_version: ["11", "10", "9.6"]},
+                options: {pg_version: ["12", "11", "10", "9.6"]},
```
solugebefola, line 50: Aren't we deprecating 10, and 9.6?

k-rus: We deprecate them, but we still support them and binaries are generated and work. I don't know how the deprecation is actual supposed to work. There is already comment in the text, which mentions deprecation. See on [this page](https://docs.timescale.com/latest/getting-started/installation/windows/installation-windows):

>Note: TimescaleDB requires PostgreSQL 11.4 or 12.0+. Support for PostgreSQL 9.6.3+ and 10.9+ is deprecated and will be removed in a future release.

solugebefola: So I think that we should remove `10` and `9.6` for the `v1.7` branch of the docs.  @mfreed opinion?

solugebefola: There's no docker of 1.7 with PG10 or 9.6, right?

k-rus: @solugebefola How could the users get binaries for deprecated version in such case? Should we create a special page, which will be a work to do?

solugebefola: We have 1.7 binaries for PG10 and 9.6?

solugebefola: Regardless, we should merge this now so folks have a way to access the PG12 binary and remove the others in a separate PR if necessary.

k-rus: @solugebefola I built and published docker images for 9.6 and 10.
1.7 binaries for 9.6 and 10 are already available for download through our docs. Do you suggest to remove binaries, which are already available for users.

k-rus: I suggest to keep access to 9.6 and 10 even they are replicated and remove them when we remove support, i.e., from 2.0.

PR #375, opened 2020-04-22
Update continuous-aggregates.md

bboule:
Update to call out Real Time Aggs in docs making it searchable

### DIFF
--- a/using-timescaledb/continuous-aggregates.md
+++ b/using-timescaledb/continuous-aggregates.md
@@ -20,9 +20,17 @@ automatic, it doesn’t add any maintenance burden to your database.
 
 **How it Works:** A *materialization background job* regularly takes
 raw data from the hypertable and computes a partial aggregate that it
-stores (materializes) in the continuous aggregate.
+stores (materializes) in the continuous aggregate. When new data is 
+inserted, updated, or deleted in the hypertable, the continuous aggregate
+will automatically decide what data needs to be re-materialized and 
+schedule a re-materialization to happen the next time the materialization 
+job runs.
 
-Querying the *continuous aggregate view* will then compute an
+
+**Real-Time Aggregates** 
+
+Real-time aggregates are a capability (first introduced in TimescaleDB 1.7)
+whereby querying the *continuous aggregate view* will then compute an
 up-to-date final aggregate result by combining the materialized
 partial aggregate with recent data from the hypertable that has yet to
 be materialized by the continuous aggregate. By combining raw and
@@ -30,10 +38,39 @@ materialized data in this way, one gets accurate and up-to-date
 results while still enjoying the speedups of pre-computing a large
 portion of the result.
 
-When new data is inserted, updated, or deleted in the hypertable, the
-continuous aggregate will automatically decide what data needs to be
-re-materialized and schedule a re-materialization to happen the next
-time the materialization job runs.
+As an example, continuous aggregates _without_ this real-time capability
+make it really fast to get aggregate 
+answers by precomputing these values (such as the min/max/average value over
+each hour). This way, if you are collecting raw data every second, querying hourly
+data over the past week means reading 24 x 7 = 168 values from the database, as
+opposed to processing 60 x 60 x 24 x 7 = 604,800 values at query time.  
+
+But this type of continuous aggregate does not incorporate the very
+latest data, _i.e._, since the last time the asynchronous aggregation job ran
+inside the database. So if you are generating hourly rollups, you might only
+run this materialization job every hour.
+
+With real-time aggregates, a single, simple query will combine your
+pre-computed hourly rollups with the raw data from the last
+hour, to always give you an up-to-date answer.  Now, instead of touching
+604,800 rows of raw data, the query reads 167 pre-computed rows of
+hourly data and 3600 rows of raw secondly data, leading to significant
+performance improvements.
+
+Real-time aggregates are now the default behavior for any continuous aggregates.  
+To revert to previous behavior, in which the query touches materialized data only 
+and doesn't combine with the latest raw data, add the following parameter when 
+creating the continuous aggregate view:
+
+timescaledb.materialized_only=true
+
+You can also use this in conjunction with the ALTER VIEW to turn this feature
+on or off at any time.
+
+>:TIP: Upgrading continuous aggregates that were created in a version earlier than
+TimescaleDB 1.7 to use real-time aggregates should also ALTER the view to
+set ```timescaledb.materialized_only=false```.   Subsequent to this change, queries 
+to the view will immediately use the real-time aggregate feature at query time.
 
 ### Creating a Continuous Aggregate View [](create)
 [Continuous aggregates][api-continuous-aggs] are created by setting the
--- a/using-timescaledb/continuous-aggregates.md
+++ b/using-timescaledb/continuous-aggregates.md
@@ -62,14 +62,16 @@ To revert to previous behavior, in which the query touches materialized data onl
 and doesn't combine with the latest raw data, add the following parameter when 
 creating the continuous aggregate view:
 
+```
 timescaledb.materialized_only=true
+```
 
 You can also use this in conjunction with the ALTER VIEW to turn this feature
 on or off at any time.
 
->:TIP: Upgrading continuous aggregates that were created in a version earlier than
-TimescaleDB 1.7 to use real-time aggregates should also ALTER the view to
-set ```timescaledb.materialized_only=false```.   Subsequent to this change, queries 
+>:TIP: To upgrade continuous aggregates that were created in a version earlier than
+TimescaleDB 1.7 to use real-time aggregates, ALTER the view to
+set `timescaledb.materialized_only=false`.   All Subsequent queries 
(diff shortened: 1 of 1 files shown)


using-timescaledb/continuous-aggregates.md
```diff
@@ -22,7 +22,11 @@ automatic, it doesn’t add any maintenance burden to your database.
 raw data from the hypertable and computes a partial aggregate that it
 stores (materializes) in the continuous aggregate.
 
-Querying the *continuous aggregate view* will then compute an
+
+**Real-Time Aggregate** 
+
+Real-Time Aggregates are a concept that was introduced in TimescaleDB 1.7
+whereby Querying the *continuous aggregate view* will then compute an
```
mfreed, line 29: Querying -> querying

mfreed: Also, what about the following paragraph after this next paragraph (github interface not letting me add comment there:

```
As an example, continuous aggregates made it really fast to get aggregate 
answers by precomputing these values (such as the min/max/average value over
each hour). This way, if you are collecting raw data every second, querying hourly
data over the past week means reading 24 x 7 = 168 values from the database, as
opposed to processing 60 x 60 x 24 x 7 = 604,800 values at query time.  But one
limitation with continuous aggregates is that they don't incorporate the very
latest data, _i.e._, since the last time the asynchronous aggregation job ran
inside the database. So if you are generating hourly rollups, you might only
run this job every hour.

But now with real-time aggregates, a single, simple query will combine your
pre-computed hourly rollups with the raw data from the last
hour, to always give you an up-to-date answer.  Now, instead of touching
604,800 rows of raw data, the query reads 167 pre-computed rows of
hourly data and 3600 rows of raw secondly data, leading to significant
performance improvements.
```

mfreed: Also the next paragraph "When new data"...doesn't seem to fit under RTA, and instead should be more top-level?

using-timescaledb/continuous-aggregates.md
```diff
@@ -22,7 +22,11 @@ automatic, it doesn’t add any maintenance burden to your database.
 raw data from the hypertable and computes a partial aggregate that it
 stores (materializes) in the continuous aggregate.
 
-Querying the *continuous aggregate view* will then compute an
+
+**Real-Time Aggregate** 
+
+Real-Time Aggregates are a concept that was introduced in TimescaleDB 1.7
```
mfreed, line 28:
```suggestion
Real-time aggregates are a capability (first introduced in TimescaleDB 1.7)
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -35,6 +39,20 @@ continuous aggregate will automatically decide what data needs to be
 re-materialized and schedule a re-materialization to happen the next
 time the materialization job runs.
 
+This is now the default behavior for any new continuous aggregate built
+after the 1.7 upgrade or for any new instance built with version 1.7.  
+To revert the continuous aggregate to the pre 1.7 behavior (that is not 
+including data that has not been materialized in your query) you can simply 
+add the following parameter at creation time:
```
mfreed, line 46:
```suggestion
Real-time aggregates are now the default behavior for any continuous aggregates.  
To revert to previous behavior, in which the query touches materialized data only 
and doesn't combine with the latest raw data, add the following parameter when 
creating the continuous agggregate view
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -35,6 +39,20 @@ continuous aggregate will automatically decide what data needs to be
...
+
+timescaledb.materialized_only=true
+
+You can also use this in conjunction with the ALTER VIEW to turn this feature
+on or off at any time, this is also the process for upgrading any continuous
+aggregates that were created prior to the 1.7 upgrade (setting this flag to FALSE).  
+Once you issue the ALTER VIEW and set the above parameter your continuous aggregate 
+will then use the real-time feature at query time.
```
mfreed, line 54:
```suggestion
You can also use this in conjunction with the ALTER VIEW to turn this feature
on or off at any time.

>:TIP: Upgrading continuous aggregates that were created in a version earlier that
TimescaleDB 1.7 to use real-time aggregates should also ALTER the view to
set `timescaledb.materialized_only=false`.   Subsequent to this change, queries 
to the view will immediately use the real-time aggregate feature at query time.
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -22,7 +22,11 @@ automatic, it doesn’t add any maintenance burden to your database.
 raw data from the hypertable and computes a partial aggregate that it
 stores (materializes) in the continuous aggregate.
 
-Querying the *continuous aggregate view* will then compute an
+
+**Real-Time Aggregate**
```
mfreed, line 26:
```suggestion
**Real-Time Aggregates** 
```

svenklemm: I dont know maybe its too late for that but i really like to call it Real-Time Aggregation cause that works much better in conjunction with continuous aggregates

using-timescaledb/continuous-aggregates.md
```diff
@@ -20,20 +20,55 @@ automatic, it doesn’t add any maintenance burden to your database.
...
+latest data, _i.e._, since the last time the asynchronous aggregation job ran
+inside the database. So if you are generating hourly rollups, you might only
+run this job every hour.
+
+But now with real-time aggregates, a single, simple query will combine your
+pre-computed hourly rollups with the raw data from the last
+hour, to always give you an up-to-date answer.  Now, instead of touching
+604,800 rows of raw data, the query reads 167 pre-computed rows of
+hourly data and 3600 rows of raw secondly data, leading to significant
+performance improvements.
```
mfreed, line 56: Sorry, my bad. This was text from my previous release notes.  I should have read more carefully :)

```suggestion
As an example, continuous aggregates _without_ this real-time capability
make it really fast to get aggregate 
answers by precomputing these values (such as the min/max/average value over
each hour). This way, if you are collecting raw data every second, querying hourly
data over the past week means reading 24 x 7 = 168 values from the database, as
opposed to processing 60 x 60 x 24 x 7 = 604,800 values at query time.  

But this type of continuous aggregate does not incorporate the very
latest data, _i.e._, since the last time the asynchronous aggregation job ran
inside the database. So if you are generating hourly rollups, you might only
run this materialization job every hour.

With real-time aggregates, a single, simple query will combine your
pre-computed hourly rollups with the raw data from the last
hour, to always give you an up-to-date answer.  Now, instead of touching
604,800 rows of raw data, the query reads 167 pre-computed rows of
hourly data and 3600 rows of raw secondly data, leading to significant
performance improvements.
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -20,20 +20,55 @@ automatic, it doesn’t add any maintenance burden to your database.
...
+
+Real-time aggregates are now the default behavior for any continuous aggregates.  
+To revert to previous behavior, in which the query touches materialized data only 
+and doesn't combine with the latest raw data, add the following parameter when 
+creating the continuous agggregate view
+
+timescaledb.materialized_only=true
```
mfreed, line 63:
```suggestion
creating the continuous aggregate view:

timescaledb.materialized_only=true
```

--> **Also** wrap `timescaledb.materialized_only=true` in triple backticks --> Sorry I couldn't do this in example, as markdown was interpreting it differently.

using-timescaledb/continuous-aggregates.md
```diff
@@ -20,20 +20,55 @@ automatic, it doesn’t add any maintenance burden to your database.
...
+
+You can also use this in conjunction with the ALTER VIEW to turn this feature
+on or off at any time.
+
+>:TIP: Upgrading continuous aggregates that were created in a version earlier that
```
mfreed, line 68:
```suggestion
>:TIP: Upgrading continuous aggregates that were created in a version earlier than
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -69,7 +69,7 @@ on or off at any time.
 
 >:TIP: Upgrading continuous aggregates that were created in a version earlier than
 TimescaleDB 1.7 to use real-time aggregates should also ALTER the view to
-set `timescaledb.materialized_only=false`.   Subsequent to this change, queries 
+set ```timescaledb.materialized_only=false```.   Subsequent to this change, queries
```
solugebefola, line 72: Don't think this is necessary

solugebefola: It doesn't affect how it parses.  If you want it in a non-inline code block, you need the backticks on separate lines
> \`\`\`
code here
\`\`\`

solugebefola: Looks fine inline though:
Screen Shot 2020-04-22 at 12 14 39 PM

mfreed: Oops, yes, I meant this to be single ticks not three ticks.

mfreed:
Replaced by https://github.com/timescale/docs.timescale.com-content/pull/376

PR #392, opened 2020-05-14
Remove cascade option from drop_chunks

mkindahl:
The `cascade` option is removed from `drop_chunks`.

Documentation for timescale/timescaledb-private#786


solugebefola:
@svenklemm I'm not sure of the details here, so can you review?  Is cascade going away because it's the default in `2.0`?

PR #402, opened 2020-05-20
Remove getting-started/exploring-enterprise

alexthinath:
This section is no longer relevant with our current cloud model. I believe it can be removed altogether to avoid potential confusion.

### DIFF
--- a/getting-started/exploring-enterprise.md
+++ b/getting-started/exploring-enterprise.md
@@ -56,7 +56,7 @@ You'll be able to use the above command to check your license type and license e
 
 Continue exploring our enterprise features in our API docs.
 
-
+ 
 [products]: https://www.timescale.com/products
 [license]: https://www.timescale.com/enterprise-signup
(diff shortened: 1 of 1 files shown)


mfreed:
I don't think the commit should be removing the markdown (e.g., for on-premise customers), but instead that the commit should just removing the section from page-index.js (and any other links).

solugebefola:
Status update on this PR?

alexthinath:
I didnt make any content changes since I wasnt sure how the team wanted to address removing this section. The smallest change github would let me file an issue was adding a line. I'm fine to follow Mike's suggestion, that seems good.

solugebefola:
So is the plan to just remove the page-index entry?  It will be completely unreachable.  Or is the plan to hide the page except by direct link?  In that case, it requires changing the page-index type to `NON_MENU_PAGE`

solugebefola:
Either way, any other links on the site would need to be removed

alexthinath:
deleting this

PR #410, opened 2020-05-27
Add suggestions and limitations around gapfill

mfreed:
- Gapfill must be a top-level expression
- Recommendation to include a WHERE clause for gapfill for constraint exclusion

### DIFF
--- a/api.md
+++ b/api.md
@@ -289,9 +289,9 @@ to how you handle constraints. A hypertable can contain foreign keys to normal S
 columns, but the reverse is not allowed. UNIQUE and PRIMARY constraints must include the
 partitioning key.
 >
->The deadlock is likely to happen when concurrent transactions simultaneously try 
+>The deadlock is likely to happen when concurrent transactions simultaneously try
 to insert data into tables that are referenced in the foreign key constraints and into the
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -1990,6 +1990,19 @@ filling for the interval between `start` and `finish`. It can only be used with
...
+be inferred from the WHERE clause if not supplied as arguments.
+
+>:TIP: We recommend using a WHERE clause whenever possible (instead of just
+ `start` and `finish` arguments), as TimescaleDB will then properly perform
+ chunk exclusion having been supplied the WHERE predicate.  Chunk exclusion
+ will not occur if only `start` and `finish` arguments are supplied.
+
+The `time_bucket_gapfill` must be a top-level expression in a query or
+subquery, as shown in the above examples.  You cannot, for example, do
+something like `round(time_bucket_gapfill(...))` or cast the result the gapfill
```
solugebefola, line 2003:
```suggestion
something like `round(time_bucket_gapfill(...))` or cast the result of the gapfill
```

svenklemm: since chunk exclusion requires a filter this seems to suggest that start and stop will be applied as filter which is not the case, start and stop only controls the interval where gapfilled tuples will be injected

PR #421, opened 2020-06-20
Add PG12 supporting compression

k-rus:
Since 1.7 release, compression is supported in PG12 too.

### DIFF
--- a/api.md
+++ b/api.md
@@ -976,7 +976,8 @@ policy](#add_compress_chunks_policy) for when to compress chunks.
 Advanced usage of compression alows users to [compress chunks
 manually](#compress_chunk), instead of automatically as they age.
 
->:WARNING: Compression is only available for Postgres 10.2+ or 11.
+>:WARNING: Compression is not available when using TimescaleDB on 
+PostgreSQL 9.6.
 
  #### Restrictions
  Version 1.5 does not support altering or inserting data into compressed
--- a/using-timescaledb/compression.md
+++ b/using-timescaledb/compression.md
@@ -1,7 +1,7 @@
 # Compression
 
->:WARNING:Compression is disabled when using Timescale in conjunction with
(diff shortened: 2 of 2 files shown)


api.md
```diff
@@ -976,7 +976,7 @@ policy](#add_compress_chunks_policy) for when to compress chunks.
 Advanced usage of compression alows users to [compress chunks
 manually](#compress_chunk), instead of automatically as they age.
 
->:WARNING: Compression is only available for Postgres 10.2+ or 11.
+>:WARNING: Compression is only available for Postgres 10.2+, 11 or 12.
```
svenklemm, line 979: Maybe change to is not available on 9.6

k-rus: @svenklemm Does it really matter? This statement existed for long time and 9.6 wasn't mentioned. Now it is going to be gone. This change will affect only 1.7.0, 1.7.1 and 1.7.2.
I just want to be careful with making warning text longer.

k-rus: @svenklemm Another disadvantage of specifying 9.6. It might be confusing and require to write about versions before 9.6 and 13. Have you experienced that users were confused with current formulation? I haven't seen any and if there is no-one then I suggest to keep as currently proposed.

svenklemm: `Compression is not available for Postgres 9.6.`
The suggested change would actually be shorter and I would prefer it because its true even if people look in the documentation for the wrong version

k-rus: @svenklemm Thank you for the clarification. This suggested me to look around and I can see it is inconsistent in the documentation. [This page](https://docs.timescale.com/v1.6/using-timescaledb/compression) has formulation similar to your suggestion:
> Compression is disabled when using Timescale in conjunction with PostgreSQL 9.6. In order to use compression, you must be using PostgreSQL 10.2 or higher.

I will improve the PR.

k-rus: Addressed

PR #422, opened 2020-06-20
Support compression on all PG versions

k-rus:
2.0 release removes support for PostgreSQL 9.6 and 10, thus the 
compression is always supported and the warnings are not needed.

### DIFF
--- a/api.md
+++ b/api.md
@@ -976,14 +976,12 @@ policy](#add_compress_chunks_policy) for when to compress chunks.
 Advanced usage of compression alows users to [compress chunks
 manually](#compress_chunk), instead of automatically as they age.
 
->:WARNING: Compression is not available when using TimescaleDB on 
-PostgreSQL 9.6.
+#### Restrictions
 
- #### Restrictions
- Version 1.5 does not support altering or inserting data into compressed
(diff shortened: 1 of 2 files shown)


k-rus:
I removed another waring and removed useless indentation in the following paragraph.

api.md
```diff
@@ -976,14 +976,12 @@ policy](#add_compress_chunks_policy) for when to compress chunks.
...
-PostgreSQL 9.6.
-
- #### Restrictions
- Version 1.5 does not support altering or inserting data into compressed
- chunks. The data can be queried without any modifications, however if you
- need to backfill or update data in a compressed chunk you will need to
- decompress the chunk(s) first.
+#### Restrictions
+
+Version 1.5 does not support altering or inserting data into compressed
```
svenklemm, line 981: should remove the version here as docs are versioned and this is true even for versions > 1.5

k-rus: Will it be better to remove it completely in a separate PR, which will be deployed for 1.5 and higher?

svenklemm: oh yea good thought this change would need to be pulled into different than the pg version change

PR #423, opened 2020-06-22
Update documentation regarding pg_upgrade

mkindahl:
The variable `timescaledb.restoring` does not need to be set when
performing a `pg_upgrade` so that was removed from the instructions.

In addition, the `timescaledb_post_restore` function will now *reset*
the `timescaledb.restoring` flag rather than setting it to `off`.
Having `timescaledb.restoring` set to `off` on a database can cause
issues when upgrading PostgreSQL for older versions of TimescaleDB.

### DIFF
--- a/api.md
+++ b/api.md
@@ -2775,6 +2775,9 @@ Specifically this sets the `timescaledb.restoring` GUC to `on` and stops any
 background workers which may have been performing tasks until the [`timescaledb_post_restore`](#timescaledb_post_restore)
 fuction is run following the restore. See [backup/restore docs][backup-restore] for more information.
 
+>:WARNING: Using this function when doing an upgrade could cause
+>issues in TimescaleDB versions before 1.7.1.
+
 >:WARNING: After running `SELECT timescaledb_pre_restore()` you must run the
   [`timescaledb_post_restore`](#timescaledb_post_restore) function before using the database normally.
 
@@ -2788,7 +2791,7 @@ SELECT timescaledb_pre_restore();
 
 ## timescaledb_post_restore() [](timescaledb_post_restore)
 Perform the proper operations after restoring the database has completed.
-Specifically this sets the `timescaledb.restoring` GUC to `off` and restarts any
+Specifically this resets the `timescaledb.restoring` GUC and restarts any
 background workers. See [backup/restore docs][backup-restore] for more information.
 
 #### Sample Usage  [](timescaledb_post_restore-examples)
--- a/using-timescaledb/update-db.md
+++ b/using-timescaledb/update-db.md
@@ -7,7 +7,7 @@ automated migration scripts that convert any internal state if necessary.
 
 >:TIP: If you are looking to upgrade the version of the **PostgreSQL instance** (e.g. from 11 to 12) rather than the version of the TimescaleDB extension, you have two choices. Either use [`pg_upgrade`][pg_upgrade] with the command:
 > ```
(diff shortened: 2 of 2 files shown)


api.md
```diff
@@ -2767,6 +2767,10 @@ Specifically this sets the `timescaledb.restoring` GUC to `on` and stops any
 background workers which may have been performing tasks until the [`timescaledb_post_restore`](#timescaledb_post_restore)
 fuction is run following the restore. See [backup/restore docs][backup-restore] for more information.
 
+Note that you do not need so use this function when doing an upgrade
```
svenklemm, line 2770: should this be tip or warning?

mkindahl: I think a warning is more appropriate. Changed.

api.md
```diff
@@ -2767,6 +2767,10 @@ Specifically this sets the `timescaledb.restoring` GUC to `on` and stops any
 background workers which may have been performing tasks until the [`timescaledb_post_restore`](#timescaledb_post_restore)
 fuction is run following the restore. See [backup/restore docs][backup-restore] for more information.
 
+>:WARNING: Note that you do not need so use this function when doing an upgrade
```
svenklemm, line 2770: there is typo `need so` => `need to`

api.md
```diff
@@ -2767,6 +2767,10 @@ Specifically this sets the `timescaledb.restoring` GUC to `on` and stops any
 background workers which may have been performing tasks until the [`timescaledb_post_restore`](#timescaledb_post_restore)
 fuction is run following the restore. See [backup/restore docs][backup-restore] for more information.
 
+>:WARNING: Note that you do not need to use this function when doing an upgrade
+using `pg_upgrade` and using this function when doing an upgrade could
+cause issue in previous versions of TimescaleDB.
```
davidkohn88, line 2772: `using this function when doing an upgrade could cause issue in previous versions of TimescaleDB.` 

I would reword as using this function when doing an upgrade can cause problems in previous versions....

mkindahl: Reformulated.

api.md
```diff
@@ -2775,6 +2775,8 @@ Specifically this sets the `timescaledb.restoring` GUC to `on` and stops any
 background workers which may have been performing tasks until the [`timescaledb_post_restore`](#timescaledb_post_restore)
 fuction is run following the restore. See [backup/restore docs][backup-restore] for more information.
 
+>:WARNING: Using this function when doing an upgrade could cause issue in previous versions of TimescaleDB.
```
davidkohn88, line 2778: Nit: could cause an issue, or could cause issues, not could cause issue.

mkindahl: Used "could cause issues".

api.md
```diff
@@ -2775,6 +2775,8 @@ Specifically this sets the `timescaledb.restoring` GUC to `on` and stops any
 background workers which may have been performing tasks until the [`timescaledb_post_restore`](#timescaledb_post_restore)
 fuction is run following the restore. See [backup/restore docs][backup-restore] for more information.
 
+>:WARNING: Using this function when doing an upgrade could cause issues in previous versions of TimescaleDB.
```
mfreed, line 2778: What does  "previous versions" mean?  Should we specify anything like "versions earlier than 1.x"?

mkindahl: Clarified.

api.md

api.md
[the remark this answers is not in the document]
mkindahl: Nothing special is needed, strangely enough.

PR #432, opened 2020-06-24
Remove version of current compression restriction

k-rus:
The documentation is versioning per TimescaleDB release. Thus it is
clear for which versions the restrictions are applicable and there is
no need to mention specific version in the text.

### DIFF
--- a/api.md
+++ b/api.md
@@ -978,7 +978,7 @@ manually](#compress_chunk), instead of automatically as they age.
 
 #### Restrictions
 
-Version 1.5 does not support altering or inserting data into compressed
+The current version does not support altering or inserting data into compressed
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -978,7 +978,7 @@ manually](#compress_chunk), instead of automatically as they age.
 
 #### Restrictions
 
-Version 1.5 does not support altering or inserting data into compressed
+Current version does not support altering or inserting data into compressed
```
svenklemm, line 981: I think there should be an article here `The current version` but maybe have native english speaker chime in

k-rus: Good point. Missed it. Fixed.

PR #439, opened 2020-07-13
Document how to remove telemetry jobs

mkindahl:
If telemetry jobs are not removed from the jobs table, they will still
be scheduled and potentially generate errors or warnings. This commit
document how to remove and add telemetry jobs.

Fixes timescale/timescaledb#1788

### DIFF
--- a/using-timescaledb/telemetry.md
+++ b/using-timescaledb/telemetry.md
@@ -80,14 +80,25 @@ release, many users have reported that performance issues or bugs
 automatically resolve after updating their version of TimescaleDB.
 
 ## Disabling Telemetry
+
 Although we invite our community to help us keep improving our
-product, we do understand when users would like to disable telemetry. Note that
-disabling telemetry also disables the version checking functionality.
+product, we do understand when users would like to disable
+telemetry. Note that disabling telemetry also disables the version
+checking functionality.
 
-Telemetry is sent on a per-database basis, so users can disable telemetry for specific databases or for an entire instance.
+Telemetry is sent on a per-database basis, so users can disable
+telemetry for specific databases or for an entire instance.
 
-To turn off telemetry for an instance, simply include the following line
-in your `postgresql.conf` file:
+To disable telemetry, it is necessary to both remove the telemetry job
+from the job list and it is turn down the telemetry reporting
+level. Turning down the telemetry reporting level ensures that nothing
+is sent out even if the telemetry job is scheduled.  Removing the
+telemetry job from the jobs list will avoid it being scheduled for
+execution and hence prevent warnings and and errors resulting from
+scheduling an empty job. These two settings can be done independently.
+
+To turn down the telemetry level for an instance, include the
+following line in your `postgresql.conf` file:
 
 ```
 timescaledb.telemetry_level=off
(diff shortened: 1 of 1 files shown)


mkindahl:
> Not sure its a good idea to document this and it will be shortlived since this changes for 2.0. Also you should insert it with id 1 as there are some assumptions about this in the code

We need it for 1.7 branch anyway, and I think that branch will have to remain for a while, so good to document it.

Fixed the `id` (but the fact that we have such hard-coded assumptions make me wonder how easy it is to support this).

using-timescaledb/telemetry.md
```diff
@@ -99,25 +110,58 @@ Alternatively, in a `psql` console, run:
...
+INSERT INTO _timescaledb_config.bgw_job(id, application_name, job_type, schedule_interval, max_runtime_retries, retry_period) VALUES
+(1, 'Telemetry Reporter', 'telemetry_and_version_check_if_enabled', INTERVAL '24h', INTERVAL '100s', -1, INTERVAL '1h')
+```
+
+You can of course pick any interval you like, but these are the
+defaults for the telemetry job.
+
+>:WARNING: It is important that `job_type` and the `id` is set to these values.
+
+>:TIP: To ensure that newly created databases use telemetry, add the a
```
jflambert, line 164:
```suggestion
>:TIP: To ensure that newly created databases use telemetry, add the
```

either "a" or "the", not both

mkindahl: Fixed.

using-timescaledb/telemetry.md
```diff
@@ -99,25 +110,58 @@ Alternatively, in a `psql` console, run:
...
+INSERT INTO _timescaledb_config.bgw_job(id, application_name, job_type, schedule_interval, max_runtime_retries, retry_period) VALUES
+(1, 'Telemetry Reporter', 'telemetry_and_version_check_if_enabled', INTERVAL '24h', INTERVAL '100s', -1, INTERVAL '1h')
+```
+
+You can of course pick any interval you like, but these are the
+defaults for the telemetry job.
+
+>:WARNING: It is important that `job_type` and the `id` is set to these values.
+
+>:TIP: To ensure that newly created databases use telemetry, add the
```
svenklemm, line 164: I think this should be removed, it also only works if you installed timescaledb in template1 which has up- and downsides. Background workers are disabled for template-databases.

using-timescaledb/telemetry.md
```diff
@@ -99,25 +110,58 @@ Alternatively, in a `psql` console, run:
...
+To remove the telemetry job for a database, connect to the database
+using `psql` and execute:
+
+```sql
+DELETE FROM _timescaledb_config.bgw_job
+WHERE job_type = 'telemetry_and_version_check_if_enabled'
+```
+
+>:TIP: To ensure that newly created databases do not use telemetry,
+>remove the telemetry job from the `template1` database.
```
svenklemm, line 128: This should be removed, as this only applies to a subset of installations.

mkindahl: Well, the docker image install timescaledb into the template database as well, so I think this is good to mention. I do agree that it does not apply generally though, so we should probably qualify it.

using-timescaledb/telemetry.md
mkindahl: The docker image install `timescaledb` by default in the `template1` database, so if the user want to ensure that newly created databases use, or do not use, telemetry, you need to tweak the contents of `template1`.

using-timescaledb/telemetry.md
```diff
@@ -80,14 +80,25 @@ release, many users have reported that performance issues or bugs
...
+from the job list and it is turn down the telemetry reporting
+level. Turning down the telemetry reporting level ensures that nothing
+is sent out even if the telemetry job is scheduled.  Removing the
+telemetry job from the jobs list will avoid it being scheduled for
+execution and hence prevent warnings and and errors resulting from
+scheduling an empty job. These two settings can be done independently.
 
-To turn off telemetry for an instance, simply include the following line
-in your `postgresql.conf` file:
+To turn down the telemetry level for an instance, include the
```
svenklemm, line 100: Actually the GUC is what prevents telemetry from being sent, and should be set if you want to disable it. In addition you can remove the job but the job will be recreated in a dump/restore cycle. So after dump/restore you need to delete it again.

mkindahl: Yes, that is true, but do you suggest that we should add a notice about this?

using-timescaledb/telemetry.md
mkindahl: Removed the mention instead.

using-timescaledb/telemetry.md
mkindahl: Removed the mention instead.

mkindahl:
> I'm generally okay with this, but I wonder if we should move separate "job" instructions to a separate subsection?
> 
> So disabling telemetry section is just about configuring the GUC, which actually controls whether telemetry is sent / not sent. And then a new subsequent subsection, like:
> 
> ```
> ### Removing telemetry job
> 
> Once telemetry is disabled according to [these instructions][...],
> no telemetry will be sent from your database.  However, the background
> scheduler will still schedule a daily job, which will then immediately
> check the configuration's setting and, if disabled, do nothing.
> 
> This can lead to occasional information messages in your logs that the
> telemetry job is still scheduled (and warnings if you don't have enough
> background workers).  To completely remove this daily job, you can also
> directly remove the job from your database:
> 
> ...
> 
> >:TIP: These instructions are only to prevent the continued local telemetry
> check, and are not needed if you simply want to prevent telemetry from being sent.
> ```
> 
> ^ I wrote the above quickly, so can likely be improved. I think this will also reduce user confusion, as I worry that the current proposed instructions might suggest to users they need to take these more complex instruction to prevent telemetry from being sent.

I think it is better to have a separate section about this, so I rewrote it to have a separate section. There are also some other issues that is clearer with a separate section.

using-timescaledb/telemetry.md
```diff
@@ -79,15 +79,26 @@ version warnings. While you do not have to update immediately to the newest
...
-disabling telemetry also disables the version checking functionality.
+product, we do understand when users would like to disable
+telemetry. Note that disabling telemetry also disables the version
+checking functionality.
+
+Telemetry is sent on a per-database basis, so users can disable
+telemetry for specific databases or for an entire instance.
 
-Telemetry is sent on a per-database basis, so users can disable telemetry for specific databases or for an entire instance.
+To disable telemetry, it is necessary to both remove the telemetry job
```
gayyappan, line 92: This is a bug with our system. I strongly think we should not encourage users to modify catalog tables. In 2.0 we will have the ability to disable jobs.  Meanwhile, we have a couple of other options : a) A temporary bug fix might be to actually delete the job from the table if a user disables telemetry or b) Document the work around as part of the github issue but not as part of the documentation.

mkindahl:
Closing issue since we will have a disable flag in 2.0 and it will not be necessary to delete jobs.

PR #442, opened 2020-07-14
Update tutorial-grafana-dashboards.md

toymachiner62:
Updated to show entire query. Otherwise the query run as written returns an error.


toymachiner62:
Now that i'm further in the tutorial, I see where you actually have the full query with the group by clause.

PR #444, opened 2020-07-14
Hide enterprise

mfreed:
Given the new focus on Cloud and Community, does not make sense to make the Enterprise version feature so prominently in docs, as it leads to confusion.


mfreed:
Git history here had additional commit.  Created new clean PR in https://github.com/timescale/docs.timescale.com-content/pull/449

PR #448, opened 2020-07-15
WIP Remove mentioning private beta

k-rus:
WIP

Multinode is available in public beta and doesn't require to contact
us for testing it. Thus the private beta notes are removed. This
commit also removes dead file, which doesn't provide much information.


k-rus:
I will create new PR, since it's easier.

PR #452, opened 2020-07-21
Reorganize multi-node instructions

mfreed:
- Move scaling out and data node into "setting up multi-node" section in
  top-level "getting started" section

- Move details about creating and scaling distributed hypertables into
  top-level "using timescaledb" section


mfreed:
@mkindahl My previous PR missed two important files in the commit.  Please review again?  Thanks!

using-timescaledb/distributed-hypertables.md
```diff
@@ -0,0 +1,124 @@
...
+single machine can handle, you can distribute the data chunks over
+several machines by using *distributed hypertables*.
+
+Distributed hypertables are otherwise similar to normal hypertables, but
+distributed hypertables will also spread a hypertable's chunks across multiple
+data nodes.  Distributed hypertables also typically use an additional "space
+partitioning" of data (e.g., by server or container ID, by device ID, by ticker
+symbol, etc.) so that an individual time interval is spread over all data nodes
+(although this is not required).  Thus to create a distribted hypertable, you
+need to have added at least one data node before starting.
```
mkindahl, line 20: Suggestion:

> Thus to create a distributed hypertable, you need to have added at least one data node before starting. Distributed hypertables are typically also partitioned by an additional column, such as server or container ID, device ID, or ticker symbol. It is not strictly required, but can improve performance since individual time intervals are spread over several data nodes.

using-timescaledb/distributed-hypertables.md
```diff
@@ -0,0 +1,124 @@
...
+By default, distributed hypertables will use all available data nodes,
+but individual data nodes can be added and removed from already
+created distributed hypertables using
+[`add_data_node`][add_data_node]/[`attach_data_node`][attach_data_node] and
+[`detach_data_node`][detach_data_node].
+
+For example, if you add a new data node to the distributed database,
+it will not automatically be
+added to existing distributed hypertables, so it is necessary to attach
+it explicitly.
```
mkindahl, line 79: Suggestion to switch the order to make it flow better. The fact that data nodes are not added to existing hypertables is important, but comes very late and might be missed. IMHO, it is better to stress the point early.
> By default, distributed hypertables will use all available data nodes when created, but if you add new data nodes to the distributed database, they are *not* automatically added to existing distributed hypertables and you need to explicitly attach them using [`attach_data_node`][attach_data_node].

using-timescaledb/distributed-hypertables.md
```diff
@@ -0,0 +1,124 @@
...
+
+```sql
+SELECT add_data_node('node3', host => 'dn3.example.com');
+
+SELECT attach_data_node('node3', hypertable => 'conditions');
```
mkindahl, line 84:
```suggestion
SELECT add_data_node('node3', host => 'dn3.example.com');
SELECT attach_data_node('node3', hypertable => 'conditions');
```

using-timescaledb/distributed-hypertables.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+Once the data node has been attached, any new chunks will also be
+distributed to the hypertable on the new data node.
+
```
mkindahl, line 89: Seems repetitive, you can probably remove it.

using-timescaledb/distributed-hypertables.md
```diff
@@ -0,0 +1,124 @@
...
+```sql
+SELECT detach_data_node('node1', hypertable => 'conditions');
+```
+
+Note that you cannot detach a data node that still holds data for the
+hypertable. However, if the distributed hypertable is subject to a
+retention policy, it is possible to block new data on a data node and
+wait until the existing data on the data node moves outside the
+retention window. This is achieved by first blocking new chunks on the
+data node:
```
mkindahl, line 102: We should probably have a link to the data retention section and might have to add more about the steps there (as well). Not sure if the blocking is covered in the data retention section.

using-timescaledb/distributed-hypertables.md
```diff
@@ -0,0 +1,124 @@
...
+
+```sql
+SELECT block_new_chunks('node1', hypertable => 'conditions');
+```
+
+This will prohibit new data from being stored on the data node. Once
+the data node's data is outside the retention window,
+[`drop_chunks`][drop_chunks] can be used to delete the data on the data
+node.
```
mkindahl, line 111:
```suggestion
This will prevent data from being added or updated on the data node. Once
the data node's data is outside the retention window,
[`drop_chunks`][drop_chunks] can be used to delete the data on the data
node.
```

using-timescaledb/distributed-hypertables.md
```diff
@@ -0,0 +1,124 @@
+# Distributed Hypertables
```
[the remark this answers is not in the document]
mfreed: It's the title of the entire page, so I don't think needs an anchor?   None of the other similar pages seem to have an anchor at the top.

using-timescaledb/distributed-hypertables.md
mfreed: I think it adds some good information, so probably will leave.

using-timescaledb/distributed-hypertables.md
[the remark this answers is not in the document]
mkindahl: In that case it's probably fine. I don't know what we use to generate the pages.

PR #482, opened 2020-09-11
Add template for new API function entries

solugebefola:
Adds a template file for standard API entries to ensure consistency.

### DIFF
--- a/api/api-template.md
+++ b/api/api-template.md
@@ -1,0 +1,57 @@
+<!-- REMOVE ALL COMMENTS AFTER FILLING OUT TEMPLATE-->
+<!-- Add your function name in the headline below with an accompanying anchor name.  The anchor should be the same as the function name-->
+## function_name() [](function_name) 
+
+<!-- 
+Include a description of the function here.  It should succinctly describe what the function does and any important considerations with regards to function usage.  Include a link to the relevant section in Using TimescaleDB if one exists.
+-->
+
+
+<!-- Include a table with required arguments, including a name and description with type (e.g. REGCLASS, INTEGER)-->
+#### Required Arguments [](function_name-required-arguments)
+
+| Name                 | Description                                       |
+|----------------------|---------------------------------------------------|
+| `sample_argument`    | (TYPE) Quick and clear description of the argument|
+| `another_argument`   | (TYPE) This argument description spans more than one line, but that's ok as long as there is an ending symbol |
+
+
+<!-- Include a table of optional arguments if there are any, otherwise do not include this section in your entry -->
+#### Optional Arguments [](function_name-optional-arguments)
+
+| Name                 | Description                                       |
+|----------------------|---------------------------------------------------|
+| `third_argument`    | (TYPE) Quick and clear description of the argument|
+| `fourth_argument`   | (TYPE) This argument description spans more than one line, but that's ok as long as there is an ending symbol |
+
+
+<!-- Include a table of what the function returns, if appropriate, otherwise do not include this section in your entry-->
+#### Returns [](function_name-returns)
+
+| Column               | Description                                       |
+|----------------------|---------------------------------------------------|
+| `sample_column`      | Quick and clear description of the column         |
+| `another_column`     | Another clear description of a column             |
+
+
+<!-- Include the most possible errors that the function returns -->
+#### Errors/Limitations
+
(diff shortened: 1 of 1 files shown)


mfreed:
Do we have an example of these designed?  The current page can get pretty noisy.  Or are you confident with this markdown, the design can be arbitrarily tweaks?

https://docs.timescale.com/latest/api#continuous_aggregate-create_view

api/api-template.md
```diff
@@ -0,0 +1,57 @@
+<!-- REMOVE ALL COMMENTS AFTER FILLING OUT TEMPLATE-->
+<!-- Add your function name in the headline below with an accompanying anchor name.  The anchor should be the same as the function name-->
+## function_name() [](function_name)
```
mfreed, line 3: What about labels/badges like :community: or :multinode: anywhere?

api/api-template.md
```diff
@@ -0,0 +1,57 @@
...
+Create a foobar that wargles the first column in the named table
+
+<!-- Include a code block with the command and sample output (if any)-->
+```sql
+SELECT function_name('sample_argument', 'another_argument);
+
+--either real output or code block comment that states what output should be
+```
+
+<!-- Do NOT include best practices here.  That should be in the Using TimescaleDB section.  -->
```
mfreed, line 57: What about a way to include an optional/standard pointer to another page/tutorial that goes into more depth?  So the API call for compress_chunks points to the section on compression in Using TimescaleDB?

(We can ultimately also add some design flourish to this pointer.)

api/api-template.md
```diff
@@ -0,0 +1,57 @@
...
+#### Required Arguments [](function_name-required-arguments)
+
+| Name                 | Description                                       |
+|----------------------|---------------------------------------------------|
+| `sample_argument`    | (TYPE) Quick and clear description of the argument|
+| `another_argument`   | (TYPE) This argument description spans more than one line, but that's ok as long as there is an ending symbol |
+
+
+<!-- Include a table of optional arguments if there are any, otherwise do not include this section in your entry -->
+#### Optional Arguments [](function_name-optional-arguments)
```
mfreed, line 20: From readability perspective, thoughts on one argument section that includes required & optional, but label them somehow accordingly?

I find the "separate sections" sometime harder to read

https://docs.timescale.com/latest/api#continuous_aggregate-create_view-with-required

api/api-template.md
```diff
@@ -0,0 +1,57 @@
+<!-- REMOVE ALL COMMENTS AFTER FILLING OUT TEMPLATE-->
+<!-- Add your function name in the headline below with an accompanying anchor name.  The anchor should be the same as the function name-->
+## function_name() [](function_name) 
+
+<!-- 
+Include a description of the function here.  It should succinctly describe what the function does and any important considerations with regards to function usage.  Include a link to the relevant section in Using TimescaleDB if one exists.
+-->
+
```
mfreed, line 8: What about functions that have SQL synctax, vs. function arguments?

https://docs.timescale.com/latest/api#continuous_aggregate-create_view

mfreed (changes requested):

api/api-template.md
solugebefola: There's a comment about that in the description section

api/api-template.md
solugebefola: We don't have those badges yet, but we can easily add them all at once when we decide on which ones we will have.

solugebefola:
This function/set of functions (https://docs.timescale.com/latest/api#continuous_aggregate-create_view) is an exception to the template format b/c of the additional column added for `Type`.  Most of them should look more like e.g. https://docs.timescale.com/latest/api#time_bucket

solugebefola:
Definitely need to figure out what we're going to do with those exceptions though.

api/api-template.md
solugebefola: I think it'll get messier with a unified section since there would need to be some additional signifier to distinguish between them.  I think that design should play a role in making it easier to parse, controlling spacing between headline and content, maybe choosing a different font size for that level of header, maybe making a `#####` instead of `####` and styling it differently, etc.

mfreed:
Comment: Team concluded that we wanted to use "TRUE" and "FALSE" for boolean values in API docs (in plaintex), rather than "true" or "True", and not in a `codeblock` either.

PR #493, opened 2020-09-16
Update chapter on using continuous aggregates

mkindahl:
Updates to chapter on using continuous aggregates with the new syntax.

### DIFF
--- a/using-timescaledb/continuous-aggregates.md
+++ b/using-timescaledb/continuous-aggregates.md
@@ -18,15 +18,280 @@ re-calculate all of the data on every refresh. Only new and/or
 invalidated data will be calculated.  Since this re-aggregation is
 automatic, it doesn’t add any maintenance burden to your database.
 
-**How it Works:** A *materialization background job* regularly takes
-raw data from the hypertable and computes a partial aggregate that it
-stores (materializes) in the continuous aggregate.
-So whenever new data is inserted, updated, or deleted in the hypertable, the
-continuous aggregate will automatically decide what data needs to be
-re-materialized, and it schedules a re-materialization to happen the next time
-the background job runs. This way, the materalization is kept mostly
-up-to-date to the raw data (and the recency and frequency of these tasks
-are fully configurable; more below).
+### An introductory example [](quick-start)
+
+As a quick introductory example, let's create a hypertable
+`conditions` containing temperature data for devices and a continuous
+aggregate to compute the daily average, minimum, and maximum
+temperature. Start off by creating the hypertable and populate it with
+some random data:
+
+```sql
+CREATE TABLE conditions (
+      time TIMESTAMPTZ NOT NULL,
+      device INTEGER NOT NULL,
+      temperature FLOAT NOT NULL,
+      PRIMARY KEY(time, device)
+);
+SELECT * FROM create_hypertable('conditions', 'time', 'device', 3);
+
+INSERT INTO conditions
+SELECT time, (random()*30)::int, random()*80 - 40
+FROM generate_series(TIMESTAMP '2020-01-01 00:00:00',
+     		     TIMESTAMP '2020-06-01 00:00:00',
+		     INTERVAL '10 min') AS time;
+```
+
+You can then create a continuous aggregate view to compute the daily
+average, minimum, and maximum temperature:
+
+```sql
+CREATE MATERIALIZED VIEW conditions_summary_hourly
+WITH (timescaledb.continuous) AS
+SELECT device,
+       time_bucket(INTERVAL '1 hour', time) AS bucket,
+       AVG(temperature),
+       MAX(temperature),
+       MIN(temperature)
+FROM conditions
+GROUP BY device, bucket;
+```
+
+Lastly, you should add a policy to ensure that the continuous
+aggregate is refreshed on a regular basis.
+
+```sql
+SELECT add_continuous_aggregate_policy('conditions_summary_hourly',
+	start_offset => INTERVAL '1 month',
+	end_offset => INTERVAL '1 h',
+	schedule_interval => INTERVAL '5 min');
+```
+
+In this case, the continuous aggregate will be refreshed every hour
+and refresh the last month's data.
+
+You can now run a normal `SELECT` on the continuous aggregate and it
+will give you the aggregated data, for example, to select the daily
+averages for device 1 during the first three months:
+
+```sql
+SELECT bucket, avg
+  FROM conditions_summary_hourly
+ WHERE device = 1 AND bucket BETWEEN '2020-01-01' AND '2020-03-31'
+ORDER BY bucket;
+```
+
+### A detailed look at continuous aggregate views [](detailed-look)
+
+As shown above, creating a refreshing [continuous
+aggregate][api-continuous-aggs] is a two-step process. First, one
+needs to create a continuous aggregate view of the data using [`CREATE
+MATERIALIZED VIEW`][postgres-createview] with the
+`timescaledb.continuous` option. Second, a continuous aggregate
+policy needs to be created to keep it refreshed.
+
+You can create several continuous aggregates for the same
+hypertable. For example, we could create another continuous aggregate
+view that summarizes the hourly data.
+
+```sql
+CREATE MATERIALIZED VIEW conditions_summary_daily
+WITH (timescaledb.continuous) AS
+SELECT device,
+       time_bucket(INTERVAL '1 day', time) AS bucket,
+       AVG(temperature),
+       MAX(temperature),
+       MIN(temperature)
+FROM conditions
+GROUP BY device, bucket;
+
+-- Create the policy
+SELECT add_continuous_aggregate_policy('conditions_summary_daily',
+	start_offset => INTERVAL '1 month',
+	end_offset => INTERVAL '1 day',
+	schedule_interval => INTERVAL '1 hour');
+```
+
+A `time_bucket` on the time partitioning column of the hypertable is
+required in all continuous aggregate views. If you do not provide one,
+you will get an error.
+
+When the view is created, it will (by default) be populated with data
+so that it contains the aggregates computed across the entire
+`conditions` hypertable.
+
+It might, however, not always be desirable to populate the continuous
+aggregate. If the amount of data in `conditions` is large and new data
+is continuously being added, it is usually more meaningful to control
+the order in which the data is refreshed or combine manual refresh
+with a policy. For example, it might be more interesting to see the
+recent data but historical data can be deferred to later. In those
+cases, the `WITH NO DATA` option can be used to avoid aggregating the
+data during creation.
+
+You could then add a policy to control the refresh of the recent data,
+and also run manual refreshes using
+[`refresh_continuous_aggregate`][refresh_continuous_aggregate] to
+refresh the historical data in a controlled manner. For example, to
+refresh one month of data you could write:
+
+```sql
+CALL refresh_continuous_aggregate('conditions_summary_hourly', '2020-05-01', '2020-06-01');
+```
+
+Unlike a regular materialized view, the refresh will only recompute
+the data within the window that has changed in the underlying
+hypertable since the last refresh. Therefore, if only a few buckets
+need updating, then the refresh is quick.
+
+Note that the end range is exclusive and aligned to the buckets of the
+continuous aggregate, so this will refresh only the buckets that are
+fully in the date range `['2020-05-01', '2020-06-01')`, that is, up to
+but not including `2020-06-01`. While it is possible to use `NULL` to
+indicate an open-ended range, we do not in general recommend using
+it. For more information, see the [Advanced Usage](#advanced-usage)
+section below.
+
+However, note that this might materialize a lot of data, can affect
+other policies such as the data retention, and affects write
+amplification, so it should only be used rarely.
+
+Continuous aggregates are supported for most aggregates that can be
+[parallelized by PostgreSQL][postgres-parallel-agg], which includes
+the normal aggregation functions like `SUM` and `AVG`. However,
+aggregates using `ORDER BY` and `DISTINCT` cannot be used with
+continuous aggregates since they are not possible to parallelize by
+PostgreSQL. In addition, TimescaleDB continuous aggregates does not
+currently support the `FILTER` clause (not to be confused with
+`WHERE`) even though it is possible to parallelize but we might add
+support for this in a future version.
+
+#### Automatic refresh with a continuous aggregate policy
+
+You can refresh the continuous aggregate view manually as mentioned
+above, but you can also automate the refresh by adding a continuous
+aggregate policies to automatically refresh data.
+
+There are a few situations that you might want to automate:
+
+- You want the continuous aggregate and the hypertable to be in sync,
+  even when data is removed from the hypertable.
+- You want to aggregate data from the hypertable into the continuous
+  aggregate but want to keep the aggregated data in the continuous
+  aggregate when removing data from the hypertable.
+
+These cases can be automatied using a *continuous aggregate policy*,
+which are added using the function
+[`add_continuous_aggregate_policy`][add-continuous-aggregate-policy].
+
+This function takes takes three arguments:
+
+- The parameter `start_interval` indicate the start of the refresh
+  window relative to the current time when the policy executes.
+- The parameter `end_interval` indicate the end of the refresh window
+  relative to the current time when the policy executes.
+- The parameter `schedule_interval` indicate the refresh interval in
+  wall-clock time.
+
+Similar to the `refresh_continuous_aggregate` function, providing
+`NULL` to `start_interval` or `end_interval` makes the range
+open-ended and will extend to the beginning or end of time,
+respectively.
+
+For example, to create a policy for `conditions_summary_hourly` that
+keeps the continuous aggregate up to date with the underlying
+hypertable `conditions` and run every hour, you would write:
+
+```sql
+SELECT add_continuous_aggregate_policy('conditions_summary_hourly',
+	start_interval => NULL,
+	end_interval => INTERVAL '1 h',
+	schedule_interval => INTERVAL '1 h');
+```
+
+This will ensure that all data in the continuous aggregate is up to
+date with the hypertable except the last hour and also ensure that we
+do not try to refresh the last bucket of the continuous
+aggregate. Since we give an open-ended `start_interval`, any data that
+is removed from `conditions` (for example, by using `DELETE` or
+[`drop_chunks`][api-drop-chunks]) will also be removed from
+`conditions_summary_hourly`. In effect, the continuous aggregate will
+always reflect the data in the underlying hypertable.
+
+If you instead want to keep the continuous aggregate up to date for
+say only the last 30 days.  the continuous aggregate even if it is
+removed from the underlying hypertable, you can set a range for the
+`start_interval`. For example, if you have a [data retention
+policy][sec-data-retention] that removed data older than one month,
+you can set `start_interval` to one month (or less) and thereby not
+refresh data older than one month, which includes data that is
+removed.
+
+```sql
+SELECT add_continuous_aggregate_policy('conditions_summary_hourly',
+	start_interval => INTERVAL '1 month',
+	end_interval => INTERVAL '1 h',
+	schedule_interval => INTERVAL '1 h');
+```
+
+>:WARNING: It is important to consider data retention policies when
+>setting up continuous aggregate policies. If the continuous aggregate
+>policy window covers data that is removed by the data retention
+>policy, the aggregates for those buckets will be refreshed and
+>consequently the data will be removed. For example, if you have a
+>data retention policy that will remove all data older than 2 weeks,
+>the continuous aggregate policy above will only have data for the
+>last two weeks. A more reasonable data retention policy for this case
+>would then be to remove data that is older than 1 month.
+>
+>You can read more about data retention with continuous aggregates in
+>the [*Data retention*][sec-data-retention] section.
+
+Time-series data is typically ordered, so it is usually the last
+bucket that gets most of the updates. Recomputing the last bucket when
+new data arrives negates many of the benefits of using continuous
+aggregation. Older buckets rarely get updated and are usually
+aggregated only once.
+
+As a result, it is recommended to configure continuous aggregate
+policies with a positive `end_interval`, that is, the materialization
+will lag behind the most recent time by this amount. A recommended
+value is at least one time bucket.
+
+A continuous aggregate may be dropped by using the `DROP MATERIALIZED
+VIEW` command. It does not affect the data in the hypertable from
+which the continuous aggregate is derived (`conditions` in the example
+above).
+
+```sql
+DROP MATERIALIZED VIEW conditions_summary_hourly;
+```
+
+---
+
+### Using Continuous Aggregates [](using)
+
+To query data from a continuous aggregate, use a `SELECT` query on
+the continuous aggregate view. For instance, you can get the average,
+minimum, and maximum for the first quarter of 2020 for device 5:
+
+```sql
+SELECT * FROM conditions_summary_hourly
+WHERE device = 5
+  AND bucket >= '2020-01-01' AND bucket < '2020-04-01';
+```
+
+Or we can do more complex queries on the aggregates themselves, for instance, if
+we wanted to know the top 20 largest metric spreads in that quarter, we could do
+something like:
+```sql
+SELECT * FROM conditions_summary_hourly
+WHERE max - min > 1800
+  AND bucket >= '2020-01-01' AND bucket < '2020-04-01'
+ORDER BY bucket DESC, device_id DESC LIMIT 20;
+```
+
+---
 
 ### Real-Time Aggregates [](real-time-aggregates)
 
@@ -40,7 +305,7 @@ results while still enjoying the speedups of pre-computing a large
 portion of the result.
 
 As an example, continuous aggregates _without_ this real-time capability make
-it really fast to get aggregate answers by precomputing these values (such as
+it really fast to get aggregate answers by pre-computing these values (such as
 the min/max/average value over each hour). This way, if you are collecting raw
 data every second, querying hourly data over the past week means reading 24 x 7
 = 168 values from the database, as opposed to processing 60 x 60 x 24 x 7 =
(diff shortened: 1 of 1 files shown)


using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +139,67 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
 currently support the `FILTER` clause even though it is possible to
 parallelize but we might add support for this in a future version.
 
+#### Creating a continuous aggregate without refreshing it
```
erimatnor, line 142: Do we really need a separate section for this? Ii seems quite repetitive and I think we can just  cover this in the previous section.  I mean, to some extent it already mentions it:  "When the view is created, it will (by default) be refreshed so that it..." We can just continue with "Add the WITH NO DATA option to avoid materializing the view on creation. This might be preferred when there's a lot of data to materialize or the continuous aggregate is created as part of transaction block."

Otherwise we are just repeating pretty much the same example over-and-over again.

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +139,67 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+
+The policy will refresh a sliding window of the continuous aggregate
+that is given by a start and end interval, which are relative to
+[`CLOCK_TIMESTAMP`][clock_timestamp]. In the same was as for the
+`refresh_continuous_aggregate` function, providing `NULL` to either of
+the intervals will make it open-ended.
+
+The `schedule_interval` tells how often the policy job should run, so
+the most recent value in the continuous aggregate will be in the range
+`[end_interval, end_interval+schedule_interval)` relative to the
```
erimatnor, line 188: Maybe it is easier to say that the most recent bucket lags behind at most `schedule_interval+end_interval` (given that there is data to materialize in that bucket).

using-timescaledb/continuous-aggregates.md
```diff
@@ -203,80 +252,13 @@ Each continuous aggregate uses the *bucket width* of the `time_bucket`
...
-`ignore_invalidation_older_than` to `30 days` and can then drop old
-chunks from the table using `drop_chunks`:
-
-```sql
-ALTER VIEW device_readings SET (
-  timescaledb.ignore_invalidation_older_than = '30 days'
-);
-SELECT drop_chunks(INTERVAL '30 days', 'device_readings')
-```
+As a result, the materialization lags behind by one bucket and up to
```
erimatnor, line 259: This seems to repeat (or be related to) the ``[end_interval, end_interval+schedule_interval)` thing above. Maybe remove/harmonize.

using-timescaledb/continuous-aggregates.md
```diff
@@ -459,23 +412,25 @@ hypertable. For instance, we may want to set the materialization hypertable's
...
+the database will automatically create composite indexes on each
+column specified in the `GROUP BY` combined with the `time_bucket`
+column (i.e. in our example, we would create an index on `{device_id,
+bucket}` and if we had grouped by other columns, `foo` and `bar`, we
+would also create indexes on `{foo, bucket}` and `{bar,
+bucket}`). Setting `timescaledb.create_group_indexes` to `false` when
+creating the view will prevent this.  If we want to create additional
+indexes or drop some of the default ones, we can do so by creating or
+dropping the appropriate indexes on the materialization hypertable
+directly.
```
erimatnor, line 425: Add that the name of the materialization table is found in the `continuous_aggregates` view?

using-timescaledb/continuous-aggregates.md
```diff
@@ -522,35 +479,39 @@ or have other areas you think we should work on, please get in touch via our
 [GitHub][timescale-github] or [Slack][support-slack].
 
 
-**Parallelized Materializations:**
-Currently, materializations are populated by a single background worker per job.
-We plan to enable multiple workers to be launched to perform a single
-materialization. This can be especially helpful in the initial population of the
-continuous aggregate, if it's created after the raw hypertable has a significant
-amount of data.
+**Parallelized Materializations:** Currently, materializations are
```
erimatnor, line 482: I am wondering if we should remove the whole Future Work. Maybe replace with "Limitations".

It seems slightly odd to me to give this kind of roadmap that we aren't sure about when and if we will deliver on.

erimatnor (commented):
* The section on `REFRESH MATERIALIZED VIEW` needs to be updated. Still mentions that a background job will run, which is no longer true (actually don't think it was true previously either). We might also decide to block this command, but I guess we'll keep for now. We should also add a note/warning that this will refresh the entire aggregate and that we recommend `referesh_continuous_aggregate` that gives more control (e.g., to reduce write amplification by not moving the invalidation threshold further than necessary).
* `max_interval_per_job` is still mentioned in several places.
* The "Continuous Aggregates" section intro seems out-of-date. It mentions that manual refreshing is not needed. While still true, the intro doesn't give the right impression since manual refreshing is no possible and policies are optional.
* For the rest, I left inline comments.

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +139,67 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+       time_bucket(INTERVAL '1 day', time) AS bucket,
+       AVG(temperature),
+       MAX(temperature),
+       MIN(temperature)
+FROM conditions
+GROUP BY device, bucket
+WITH NO DATA;
+```
+
+#### Add a continuous aggregate policy
```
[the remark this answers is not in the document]
mkindahl: Per Mike's suggestion, I think a brief example in the beginning with just the basic nuts and bolts might handle that. Let me see if I can reorder this section and still keep it together.

using-timescaledb/continuous-aggregates.md
```diff
@@ -203,80 +252,13 @@ Each continuous aggregate uses the *bucket width* of the `time_bucket`
...
-
-```sql
-ALTER VIEW device_readings SET (
-  timescaledb.ignore_invalidation_older_than = '30 days'
-);
-SELECT drop_chunks(INTERVAL '30 days', 'device_readings')
-```
+As a result, the materialization lags behind by one bucket and up to
+an additional `schedule_interval` as a result of the job not running
+continuously.
```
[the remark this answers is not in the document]
mkindahl: This was there before, I just reformatted. Will take closer look at it. I was not very happy with several sections of the text, but was unsure where to draw the line regarding a re-write.

using-timescaledb/continuous-aggregates.md
mkindahl: All for removing the future work section, I do not think we should have any of those at all since they are hard to keep accurate.

mkindahl:
>     * The section on `REFRESH MATERIALIZED VIEW` needs to be updated. Still mentions that a background job will run, which is no longer true (actually don't think it was true previously either). We might also decide to block this command, but I guess we'll keep for now. We should also add a note/warning that this will refresh the entire aggregate and that we recommend `referesh_continuous_aggregate` that gives more control (e.g., to reduce write amplification by not moving the invalidation threshold further than necessary).

I have moved these changes to the PR #496 and will update the API documentation there.

>     * `max_interval_per_job` is still mentioned in several places.

In order to make the changes manageable, I am focusing on the using chapter in this PR. There is no mention of `max_interval_per_job` in that file.
 
>     * The "Continuous Aggregates" section intro seems out-of-date. It mentions that manual refreshing is not needed. While still true, the intro doesn't give the right impression since manual refreshing is no possible and policies are optional.

The section is in dire need of a rewrite, but I am reluctant to invest a lot of time in making drastic changes unless there is a general approval of doing a drastic rewrite.

>     * For the rest, I left inline comments.

using-timescaledb/continuous-aggregates.md
mkindahl: Reworded.

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +139,67 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+The policy will refresh a sliding window of the continuous aggregate
+that is given by a start and end interval, which are relative to
+[`CLOCK_TIMESTAMP`][clock_timestamp]. In the same was as for the
+`refresh_continuous_aggregate` function, providing `NULL` to either of
+the intervals will make it open-ended.
+
+The `schedule_interval` tells how often the policy job should run, so
+the most recent value in the continuous aggregate will be in the range
+`[end_interval, end_interval+schedule_interval)` relative to the
+current time.
```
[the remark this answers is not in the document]
mkindahl: Isn't that obvious?

api.md
```diff
@@ -1835,16 +1797,16 @@ all the updates to the hypertable.
 
 ---
 
-## DROP VIEW (Continuous Aggregate) :community_function: [](continuous_aggregate-drop_view)
-Continuous aggregate views can be dropped using `DROP VIEW` statement.
+## DROP MATERIALIZED VIEW (Continuous Aggregate) :community_function: [](continuous_aggregate-drop_view)
+Continuous aggregate views can be dropped using `DROP MATERIALIZED VIEW` statement.
```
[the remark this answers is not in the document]
mkindahl: This was partially updated as a result of changes to the Using chapter. I decided to move all the changes to a separate PR since I expected large changes.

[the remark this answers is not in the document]
mkindahl: Updated the section on `DROP MATERIALIZED VIEW` in #496. I have tested that it works without `CASCADE`.  It is not pushed yet.

api.md
```diff
@@ -1854,7 +1816,7 @@ DROP VIEW <view_name> CASCADE;
 #### Sample Usage [](continuous_aggregate-drop_view-examples)
 Drop existing continuous aggregate.
 ```sql
-DROP VIEW contagg_view CASCADE;
+DROP MATERIALIZED VIEW contagg_view CASCADE;
```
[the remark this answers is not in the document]
mkindahl: Removed in #496. Not pushed yet.

api.md
```diff
@@ -1644,17 +1644,17 @@ at predefined intervals and materialize the results. This is suitable for
...
-*  [ALTER VIEW](#continuous_aggregate-alter_view)
+*  [CREATE MATERIALIZED VIEW](#continuous_aggregate-create_view)
+*  [ALTER MATERIALIZED VIEW](#continuous_aggregate-alter_view)
 *  [REFRESH MATERIALIZED VIEW](#continuous_aggregate-refresh_view)
-*  [DROP VIEW](#continuous_aggregate-drop_view)
+*  [DROP MATERIALIZED VIEW](#continuous_aggregate-drop_view)
 
-## CREATE VIEW (Continuous Aggregate) :community_function: [](continuous_aggregate-create_view)
-`CREATE VIEW` statement is used to create continuous aggregates.
+## CREATE MATERIALIZED VIEW (Continuous Aggregate) :community_function: [](continuous_aggregate-create_view)
```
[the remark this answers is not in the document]
mkindahl: Added to #496.

api.md
```diff
@@ -1799,13 +1767,7 @@ ALTER VIEW <view_name> SET ( timescaledb.option =  <value> )
...
 Set the max interval processed by a materializer job (that updates the continuous aggregate) to 1 week.
 ```sql
-ALTER VIEW contagg_view SET (timescaledb.max_interval_per_job = '1 week');
-```
-Set the refresh lag to 1 hour, the refresh interval to 30 minutes and the max
-interval processed by a job to 1 week for the continuous aggregate.
-```sql
-ALTER VIEW contagg_view SET (timescaledb.refresh_lag = '1h', timescaledb.max_interval_per_job = '1 week', timescaledb.refresh_interval = '30m');
-
+ALTER MATERIALIZED VIEW contagg_view SET (timescaledb.max_interval_per_job = '1 week');
```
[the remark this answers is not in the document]

using-timescaledb/continuous-aggregates.md
```diff
@@ -18,118 +18,127 @@ re-calculate all of the data on every refresh. Only new and/or
...
+
+As you saw above, creating a refreshing [continuous
+aggregate][api-continuous-aggs] consists of creating a continuous
+aggregate view of the data using [`CREATE MATERIALIZED
+VIEW`][postgres-createview] with the `timescaledb.continuous` and
+adding a continuous aggregate policy for the continuous aggregate to
+keep it refreshed.
+
+You can create several continuous aggregates for the same
+hypertable. For example: to create a continuous aggregate view is
```
gayyappan, line 93: I think you mean in additiona to the earlier cagg ( the previous example) . Could make that more explicit. 
```suggestion
hypertable. For example, we could create another continuous aggregate view that summarizes the hourly data.
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -18,118 +18,127 @@ re-calculate all of the data on every refresh. Only new and/or
...
-You can then use [`CREATE VIEW`][api-continuous-aggs-create] to create
-a continuous aggregate using the `timescaledb.continuous` view
-option. This view aggregates the temperatures into hourly buckets
-using the [`time_bucket`][time-bucket] function.
+You can then create a continuous aggregate view to compute the daily
+average, minimum, and maximum temperature:
 
 ```sql
-CREATE VIEW conditions_summary_hourly
+CREATE MATERIALIZED VIEW conditions_summary_daily
```
gayyappan, line 49: Do we want to set materialized_only = true here?

using-timescaledb/continuous-aggregates.md
```diff
@@ -18,118 +18,127 @@ re-calculate all of the data on every refresh. Only new and/or
...
+changes to the hypertable and want to refresh an interval, you can
+explicity refresh it by calling
+[`refresh_continuous_aggregate`][refresh_continuous_aggregate] and
+give the range you want to refresh:
+
+```sql
+CALL refresh_continuous_aggregate('conditions_summary_daily, '2020-05-01', '2020-06-01');
+```
+
+Note that the end range is exclusive, so this will refresh the date
```
gayyappan, line 129: Need a note somewhre that all refreshes are bucket aligned. (perhaps advanced topics section?)

using-timescaledb/continuous-aggregates.md
```diff
@@ -18,118 +18,127 @@ re-calculate all of the data on every refresh. Only new and/or
...
+A `time_bucket` on the time partitioning column of the hypertable is
+required in all continuous aggregate views. If you do not provide one,
+you will get an error.
+
+When the view is created, it will (by default) be refreshed so that it
+contain an aggregate of the current data in `conditions`. If you make
+changes to the hypertable and want to refresh an interval, you can
+explicity refresh it by calling
+[`refresh_continuous_aggregate`][refresh_continuous_aggregate] and
+give the range you want to refresh:
```
gayyappan, line 123: Should note what a refresh means: we are not rebuilding from scratch. Regions that were invalidated will be rebuilt. But other regions are not touched. So it is efficient. (cc @erimatnor please comment if this is correct).

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +148,60 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
 currently support the `FILTER` clause even though it is possible to
 parallelize but we might add support for this in a future version.
 
+By default, a continuous aggregate is refreshed when created, which
```
gayyappan, line 151:
```suggestion
By default, a continuous aggregate is populated with data when it is created. 
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +148,60 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+means that it is immediately populated with data. Since this can be a
+lengthy process it is sometimes desirable to defer materialization to
+a manual refresh or a policy that does it in the background. You can
+disable the immediate materialization of data using the `WITH NO DATA`
+option to `CREATE MATERIALIZED VIEW`.
+
+#### Automatic refreshing with a continuous aggregate policy
+
+You can refresh the continuous aggregate view manually as above, but
+it is also possible to automate the refreshing by adding a continuous aggregate
```
gayyappan, line 161: typo: refreshing->refresh

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +148,60 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+a manual refresh or a policy that does it in the background. You can
+disable the immediate materialization of data using the `WITH NO DATA`
+option to `CREATE MATERIALIZED VIEW`.
+
+#### Automatic refreshing with a continuous aggregate policy
+
+You can refresh the continuous aggregate view manually as above, but
+it is also possible to automate the refreshing by adding a continuous aggregate 
+policy with [`add_continuous_aggregate_policy`][add-continuous-aggregate-policy]. 
+For example:
```
gayyappan, line 163: Could you elaborate what the policy does? Something along the lines of:  This policy would refresh the continuous aggregate from the beginning until upto 2 hours ago.

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +148,60 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+```
+
+When the policy runs it will refresh a window of data 
+that is given by a start and end interval, which are relative to the
+[`CLOCK_TIMESTAMP`][clock_timestamp] at the time of executing the policy. Similar to the
+`refresh_continuous_aggregate` function, providing `NULL` to either of
+the intervals will always refresh the entire continuous aggregate.
+
+The `schedule_interval` determines how often the policy job runs, so
+the most recent value in the continuous aggregate lags behind with at
```
gayyappan, line 179: I think this is confusing. Would suggest you drop the latter half of the sentence and restate as 
```suggestion
The schedule_interval determines how often the policy job runs.
```
An example that modifies the schedule_interval using alter_job would be pertinent here.

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +148,60 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+SELECT add_continuous_aggregate_policy('device_summary',
+	start_interval => NULL,
+	end_interval => INTERVAL '2 h',
+	schedule_interval => INTERVAL '2 h');
+```
+
+When the policy runs it will refresh a window of data 
+that is given by a start and end interval, which are relative to the
+[`CLOCK_TIMESTAMP`][clock_timestamp] at the time of executing the policy. Similar to the
+`refresh_continuous_aggregate` function, providing `NULL` to either of
```
gayyappan, line 175: We should also add a case wiht non-null start_interval (say: start_interval=> '30 days', end_interval=> '1 day' ) and explain how that is different from the previous one.

using-timescaledb/continuous-aggregates.md
```diff
@@ -18,118 +18,127 @@ re-calculate all of the data on every refresh. Only new and/or
...
-data every second, querying hourly data over the past week means reading 24 x 7
-= 168 values from the database, as opposed to processing 60 x 60 x 24 x 7 =
-604,800 values at query time.
-
-But this type of continuous aggregate does not incorporate the very
-latest data, _i.e._, since the last time the asynchronous aggregation job ran
-inside the database. So if you are generating hourly rollups, you might only
-run this materialization job every hour.
+As a quick introductory example, let's create a hypertable
+`conditions` containing temperature data for devices and continuous
```
erimatnor, line 24:
```suggestion
`conditions` containing temperature data for devices and a continuous
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -18,118 +18,127 @@ re-calculate all of the data on every refresh. Only new and/or
...
+
+In this case, the continuous aggregate will be refreshed every hour
+and refresh the last month's data.
 
-You can create multiple continuous aggregates on the same
-hypertable. For example, you can create a continuous aggregate
-`condition_summary_daily` over the data with daily buckets:
+You can now run a normal `SELECT` on the continuous aggregate and it
+will give you the aggregated data, for example, to select the averages
+for the first three months:
```
erimatnor, line 75:
```suggestion
for device 1 during the first three months:
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -18,118 +18,127 @@ re-calculate all of the data on every refresh. Only new and/or
...
+```
+
+Note that the end range is exclusive, so this will refresh the date
+range `['2020-05-01', '2020-06-01')`, that is, up to but not including
+`2020-06-01`. You can use `NULL` to indicate an open-ended range, so
+to refresh the full range of the continuous aggregate, use:
+
+```sql
+CALL refresh_continuous_aggregate('conditions_summary_daily, NULL, NULL);
+```
```
erimatnor, line 136:
```suggestion
`2020-06-01`. While it is possible to use `NULL` to indicate an open-ended range,  we do not recommend doing so for tables that see continuous ingest of new data. Instead, never refresh the recent time intervals that still see a lot of writes. Those intervals will not produce accurate aggregates anyway and refreshing them will unnecessarily increase write amplification, which will slow down the ingest rate of the hypertable.

```sql
CALL refresh_continuous_aggregate('conditions_summary_daily, NULL, NULL);
```
```

File: using-timescaledb/continuous-aggregates.md [hunk 25]
```diff
@@ -139,19 +148,60 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
 currently support the `FILTER` clause even though it is possible to
 parallelize but we might add support for this in a future version.
 
+By default, a continuous aggregate is populated with data when it is created. 
+means that it is immediately populated with data. Since this can be a
```
erimatnor, line 152:
```suggestion
By default, a continuous aggregate is populated with data when it is created, which
means that it is immediately populated with data. Since this can be a
```

erimatnor: If you choose to adopt my suggestion above, this paragraph can probably be removed.

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +148,60 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+
+#### Automatic refreshing with a continuous aggregate policy
+
+You can refresh the continuous aggregate view manually as above, but
+it is also possible to automate the refreshing by adding a continuous aggregate 
+policy with [`add_continuous_aggregate_policy`][add-continuous-aggregate-policy]. 
+For example:
+
+```sql
+SELECT add_continuous_aggregate_policy('device_summary',
```
erimatnor, line 166: The parameter choices here seem picked at random. I suggest we start by explaining the use case, i.e., what we (or the users) are trying to achieve. We need to setup a context where the reader understands that these parameter choices make sense. For instance, setting a `NULL` start interval will ensure that the aggregate will stay fully up-to-date with the source data from the beginning of time, including any deletions of source data. This is non-obvious as presented.

Therefore, I believe we need to focus on a couple of key use cases, as described in design docs. For instance:


1. The user wants the cagg to be always up-to-date with the source data, including following retention.
2. The user wants separate retention for caggs and source hypertable (keep cagg for longer, while dropping source data)
3. The user wishes to see recent aggregates immediately and either does not care about historical data or can aggregate it later (e.g., manually)

I think we should formulate our examples around such use cases and be clear with what we're trying to achieve with a particular policy.

I do realize that refocusing on use cases is a big change to this section, so I am happy to merge what we have and then rewrite again, if that's preferred.

using-timescaledb/continuous-aggregates.md
erimatnor: Then let's remove. At least the "Synchronous Invalidation" section doesn't seem up-to-date with recent changes so it is confusing to keep that part.

using-timescaledb/continuous-aggregates.md
mkindahl: Added a comment about that.

using-timescaledb/continuous-aggregates.md
mkindahl: I think it is more suitable for the advanced topics.

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +148,60 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+When the policy runs it will refresh a window of data 
+that is given by a start and end interval, which are relative to the
+[`CLOCK_TIMESTAMP`][clock_timestamp] at the time of executing the policy. Similar to the
+`refresh_continuous_aggregate` function, providing `NULL` to either of
+the intervals will always refresh the entire continuous aggregate.
+
+The `schedule_interval` determines how often the policy job runs, so
+The schedule_interval determines how often the policy job runs.
+most `end_interval` + `schedule_interval` relative to the current
+time.
```
erimatnor, line 181: This paragraph has repetitions.

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +148,60 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+
+A continuous aggregate may be dropped by using the `DROP MATERIALIZED
+VIEW` command, which deletes the hypertable that stores the
+materialized data for the continuous aggregate; it does not affect the
+data in the hypertable from which the continuous aggregate is derived
+(`conditions` in the example above).
```
erimatnor, line 187:
```suggestion
A continuous aggregate and all its aggregates may be dropped by using the `DROP MATERIALIZED
VIEW` command; it does not affect the
data in the hypertable from which the continuous aggregate is derived
(`conditions` in the example above).
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +148,60 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+
 ---
 
 ### Using Continuous Aggregates [](using)
 
-To select data from a continuous aggregate, use a
-`SELECT` query on the continuous aggregate view. For instance, you can get the
-average, minimum, and maximum for the first quarter of 2018 for device
-5:
+To select data from a continuous aggregate, use a `SELECT` query on
```
erimatnor, line 197:
```suggestion
To query data from a continuous aggregate, use a `SELECT` query on
```

using-timescaledb/continuous-aggregates.md
mkindahl: Changed.

using-timescaledb/continuous-aggregates.md
mkindahl: I added a comment before, but it is already outlined after the query, so it does look redundant.

using-timescaledb/continuous-aggregates.md
mkindahl: Isn't that quite obvious from the description following?

using-timescaledb/continuous-aggregates.md
```diff
@@ -18,118 +18,162 @@ re-calculate all of the data on every refresh. Only new and/or
...
+>  hypertables AS (
+>    SELECT
+>      id AS hypertable_id,
+>      format('%1$I.%2$I', schema_name, table_name)::regclass AS hypertable
+>    FROM
+>      _timescaledb_catalog.hypertable
+>  )
+>SELECT
+>  hypertable,
+>  _timescaledb_internal.to_timestamp(lowest_modified_value) AS start_time,
```
erimatnor, line 170: There are several issues with this query.

1. It only works on timestamp-based tables
2. It only shows invalidations in the hypertable invalidation log and not the invalidations in the materialization invalidation log.
3. Showing raw invalidations might show a lot of duplicates, since the logs aren't deduped until a refresh runs.

using-timescaledb/continuous-aggregates.md
```diff
@@ -18,118 +18,127 @@ re-calculate all of the data on every refresh. Only new and/or
...
        AVG(temperature),
        MAX(temperature),
        MIN(temperature)
 FROM conditions
 GROUP BY device, bucket;
+
+-- Create the policy
+SELECT add_continuous_aggregate_policy('conditions_summary_hourly',
+	start_interval => INTERVAL '1 month',
+	end_interval => NULL,
```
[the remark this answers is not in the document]
erimatnor: Don't we need to rationalize NULL in the same way?

using-timescaledb/continuous-aggregates.md
```diff
@@ -18,118 +18,162 @@ re-calculate all of the data on every refresh. Only new and/or
...
+	schedule_interval => INTERVAL '5 min');
 ```
 
+A `time_bucket` on the time partitioning column of the hypertable is
+required in all continuous aggregate views. If you do not provide one,
+you will get an error.
+
+When the view is created, it will (by default) be populated with data
+so that it contain an aggregate of the current data in
```
erimatnor, line 120:
```suggestion
so that it contains an aggregate of the current data in
```

using-timescaledb/continuous-aggregates.md
erimatnor: @gayyappan it's correct.

using-timescaledb/continuous-aggregates.md
```diff
@@ -18,118 +18,162 @@ re-calculate all of the data on every refresh. Only new and/or
...
+
+```sql
+CALL refresh_continuous_aggregate('conditions_summary_daily, '2020-05-01', '2020-06-01');
+```
+
+This will update the buckets in the range that were invalidated by
+changes in the underlying hypertable, but not touch other buckets of
+the continuous aggregate, so it does not necessarily mean that all
+buckets in the range are touched. If there are just a few
+invalidations in the range, the refresh will be quick.
```
erimatnor, line 135: I would not start talking about invalidations here as it doesn't seem necessary to convey the main message (i.e., that refreshes are efficient). Instead, I suggest talking about how things work in the abstract.

```suggestion
Unlike a regular materialized view, the refresh will only materialize the data within the window that has changed in the underlying hypertable since the last refresh. Therefore, if only a few buckets need updating, then the refresh is quick.
```

using-timescaledb/continuous-aggregates.md
erimatnor: I agree with @gayyappan. Since we might not refresh the given range, which we go into depth about here, I think it is misleading to elide the fact that we will only refresh the fully covered buckets.

I think a quick note about this is sufficient and we can go into further depth in advanced topics. For instance

"Note that the refresh will only materialize the buckets that are fully covered by the given refresh window. So, unless the window is aligned with bucket boundaries, the function will likely refresh _less_ data than specified."

using-timescaledb/continuous-aggregates.md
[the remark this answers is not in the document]
gayyappan: I agree with Erik here. This is not the standard usage and should not be in the docs. 
Generally policies would be : start_interval= NULL, end_interval = xxxx. This mimics the old cagg semantics : keep everything up to the end_interval up to date.
The other recommended usage is: start_interval = xxx and end_interval= yyy.  
We could argue it is possible to set end_interval=NULL, but users shouldn't be doing this without thinking through what that means and how that interacts with real time aggregates etc.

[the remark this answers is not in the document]
mkindahl: Changed it to 1 hour.

using-timescaledb/continuous-aggregates.md
```diff
@@ -18,118 +18,127 @@ re-calculate all of the data on every refresh. Only new and/or
...
+A `time_bucket` on the time partitioning column of the hypertable is
+required in all continuous aggregate views. If you do not provide one,
+you will get an error.
+
+When the view is created, it will (by default) be refreshed so that it
+contain an aggregate of the current data in `conditions`. If you make
+changes to the hypertable and want to refresh an interval, you can
+explicity refresh it by calling
+[`refresh_continuous_aggregate`][refresh_continuous_aggregate] and
+give the range you want to refresh:
```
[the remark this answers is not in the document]
mkindahl: Wrote something. PTAL.

using-timescaledb/continuous-aggregates.md
mkindahl: Removed.

using-timescaledb/continuous-aggregates.md
mkindahl: Clarified in paragraph.

erimatnor (approved):
I think we can merge this and continue improving.

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +175,98 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+the continuous aggregate even if it is removed from the underlying
+hypertable, you can set a range for the `start_interval`. For example,
+if you have a [data retention policy][sec-data-retention] that removed
+data older than one month, you can set `start_interval` to one month
+(or less) and thereby not refresh data older than one month, which
+includes data that is removed.
+
+```sql
+SELECT add_continuous_aggregate_policy('conditions_summary_hourly',
+	start_interval => INTERVAL '1 month',
```
gayyappan, line 244: Just realized that since we are renaming these should be start_offset and end_offset?

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +175,98 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+	end_interval => INTERVAL '1 h',
+	schedule_interval => INTERVAL '1 h');
+```
+
+A continuous aggregate and all its aggregates may be dropped by using
```
gayyappan, line 249:
```suggestion
A continuous aggregate may be dropped by using
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -160,123 +275,91 @@ something like:
...
 ```
 
+---
+
+### Real-Time Aggregates [](real-time-aggregates)
+
+Real-time aggregates are a capability (first introduced in TimescaleDB 1.7)
+whereby querying the *continuous aggregate view* will then compute
+fully up-to-date aggregate results by combining the materialized
+partial aggregate with recent data from the hypertable that has yet to
```
gayyappan, line 289: we have removed "how this works" section. Without that the reader does not know what is the materialized partial aggregate. 
Adding a couple of sentences on what is stored in the materialization  should suffice as an introduction to the partial aggregate.

using-timescaledb/continuous-aggregates.md
```diff
@@ -160,123 +275,91 @@ something like:
...
-`ignore_invalidation_older_than` to `30 days` and can then drop old
-chunks from the table using `drop_chunks`:
-
-```sql
-ALTER VIEW device_readings SET (
-  timescaledb.ignore_invalidation_older_than = '30 days'
-);
-SELECT drop_chunks(INTERVAL '30 days', 'device_readings')
-```
+As a result, it is recommended to configure continuous aggregate policies with a positive `end_interval`, i.e., the materialization will lag behind the most recent time by this amount. A recommended value is at least one time bucket.
```
gayyappan, line 362: Adding the comment here as the following section is no longer applicable. We should discuss the caveats with saving down sampled data and the danger of an explicit refresh to the area.
i..e user drops data for [1, 20]
cagg has all data from [1, 40]. But refreshed only [20, 40]
Now an explicit refresh to  [ 10, 20] will drop the data for that region.

using-timescaledb/continuous-aggregates.md
```diff
@@ -18,118 +18,154 @@ re-calculate all of the data on every refresh. Only new and/or
...
+refresh of buckets that are not yet filled completely.
+	
+You should avoid refreshing time intervals that still see a lot of
+writes, which is usually the last bucket of the continuous
+aggregate. These intervals are still changing and will not produce
+accurate aggregate anyway and refreshing unnecessarily will increase
+the write amplification, which will slow down the ingest rate of the
+hypertable.
+
+```sql
```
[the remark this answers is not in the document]
mfreed: If we believe there should at least suggest usage, maybe just describe in text without example? 

"If `NULL` arguments are supplied for the start or end interval, refreshing the continuous aggregate will start at the beginning or end, respectively, of the hypertable.  This might materialize a lot ...

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +175,98 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
 currently support the `FILTER` clause even though it is possible to
 parallelize but we might add support for this in a future version.
```
mfreed, line 176: We've often had people confuse the `FILTER` keyword with a `WHERE` clause, which is often described as a filter predicate.  Way to clarify?

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +175,98 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
 
+By default, a continuous aggregate is populated with data when it is
```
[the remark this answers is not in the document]
mfreed: +1

using-timescaledb/continuous-aggregates.md
mfreed, line 244: Do we support months, or just 30 days?  I noticed that Gayathri suggested updating -- any reason for that?

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +175,98 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+(or less) and thereby not refresh data older than one month, which
+includes data that is removed.
+
+```sql
+SELECT add_continuous_aggregate_policy('conditions_summary_hourly',
+	start_interval => INTERVAL '1 month',
+	end_interval => INTERVAL '1 h',
+	schedule_interval => INTERVAL '1 h');
+```
+
```
mfreed, line 248: Maybe add something:

"It is important to consider the implications of continuous aggregate policies and data retention policies together, to ensure that dropping data from the underlying hypertable doesn't lead to overwriting aggregates in the continuous aggregate view with NULL values (unless desired).  For example, in this above example, if the data retention policy was instead  2 weeks while the continuous aggregate policy remained 1 month, after data more than 2 weeks is dropped, the next time the continuous aggregate is refreshed, the aggregate data from 2-4 weeks ago will be effectively deleted from the view."

Or something like that?

using-timescaledb/continuous-aggregates.md
mfreed: +1

using-timescaledb/continuous-aggregates.md
```diff
@@ -160,123 +275,91 @@ something like:
...
+performance improvements.
+
+Real-time aggregates are now the default behavior for any continuous
+aggregates. To revert to the previous behavior, in which the query
+touches materialized data only and doesn't combine with the latest raw
+data, add the parameter `timescaledb.materialized_only=true` when
+creating the continuous aggregate view.
+
+You can also use this in conjunction with the `ALTER MATERIALIZED
+VIEW` to turn this feature on or off at any time.
```
mfreed, line 321: Create hyperlink to API docs?

using-timescaledb/continuous-aggregates.md
```diff
@@ -160,123 +275,91 @@ something like:
...
-background worker resources while it is running.
+The `schedule_interval` option to `add_continuous_aggregate_policy`
+controls how frequently materialization jobs will be launched. Setting
+a shorter interval will mean materializations happen more frequently
+but each job consumes background worker resources while it is running.
 
 When querying the continuous aggregate view (for example,
 `conditions_summary_daily`) by default you will get a complete view of
 the data including both the partially aggregated data and newer,
 unaggregated data in the hypertable (for example, `conditions`).
```
mfreed, line 341: Maybe something like:
```suggestion
unaggregated data in the hypertable (for example, `conditions`),
based on the [real-time aggregate capability](#real-time-aggregate).
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -160,123 +275,91 @@ something like:
...
+
+>:TIP: To upgrade continuous aggregates that were created in a version
+earlier than TimescaleDB 1.7 to use real-time aggregates, ALTER the
+view to set `timescaledb.materialized_only=false`.  All subsequent
+queries to the view will immediately use the real-time aggregate
+feature.
+
+---
+
 ### Advanced Usage
```
mfreed, line 331: What's the difference between "Advanced Usage" and "Advanced Topics"?

using-timescaledb/continuous-aggregates.md
```diff
@@ -160,123 +275,91 @@ something like:
...
-```sql
-ALTER VIEW device_readings SET (
-  timescaledb.ignore_invalidation_older_than = '30 days'
-);
-SELECT drop_chunks(INTERVAL '30 days', 'device_readings')
-```
+As a result, it is recommended to configure continuous aggregate policies with a positive `end_interval`, i.e., the materialization will lag behind the most recent time by this amount. A recommended value is at least one time bucket.
 
 You can read more about data retention with continuous aggregates in
 the [*Data retention*][sec-data-retention] section.
```
mfreed, line 365: Move the views stuff out of "advanced usage"?  It also seems like the only "grouped" section here; otherwise, organize the rest of advanced usage under subsections (like "Using  `timescaledb.information` Views) ?

using-timescaledb/continuous-aggregates.md
```diff
@@ -292,65 +375,14 @@ tuning other options noted below.
...
-Usually, continuous aggregates are defined on a [date/time-type](https://www.postgresql.org/docs/current/datatype-datetime.html) column, but it is
-also possible to create your own custom scheme for handling
-aggregation for tables that are using an integer time
+Usually, continuous aggregates are defined on a
+[date/time-type](https://www.postgresql.org/docs/current/datatype-datetime.html)
+column, but it is also possible to create your own custom scheme for
+handling aggregation for tables that are using an integer time
 column. This can be useful if you have tables that use other measures
 of time that can be represented as integer values, such as nanosecond
 epochs, minutes since founding date, or whatever is suitable for your
```
mfreed, line 388: Is there a reason this example (and others) typically have a `PRIMARY KEY`?  That's usually an anti-pattern...especially setting a primary (i.e., UNIQUE) key on _time_.

using-timescaledb/continuous-aggregates.md
```diff
@@ -459,23 +491,38 @@ hypertable. For instance, we may want to set the materialization hypertable's
...
-to `false` when creating the view will prevent this.  If we want to create
-additional indexes or drop some of the default ones, we can do so by creating or
-dropping the appropriate indexes on the materialization hypertable directly.
+**Creating Indexes on the Materialization Hypertable:** By default,
+the database will automatically create composite indexes on each
+column specified in the `GROUP BY` combined with the `time_bucket`
+column (i.e. in our example, we would create an index on `{device_id,
+bucket}` and if we had grouped by other columns, `foo` and `bar`, we
+would also create indexes on `{foo, bucket}` and `{bar,
+bucket}`). Setting `timescaledb.create_group_indexes` to `false` when
```
mfreed, line 500:
```suggestion
column (i.e., in our example, because the continuous aggregate view
is defined as `GROUP BY device, bucket`, we would automatically
create a composite index on `{device, bucket}`.  If we had additionally
grouped by additional columns (e.g., `GROUP BY device, foo, bar, bucket`),
we would create additional indexes as well (`{foo, bucket}` and
`{bar, bucket}`). Setting `timescaledb.create_group_indexes` to `false` when
```

mfreed (changes requested):
Left a number of comments.

High-level question that was unresolved if the notion of invalidation and the cost of refreshing an interval.  Is this going to ship in 2.0, and if so, I think we need to discuss it somewhere.

If it isn't going to ship in 2.0, we probably need to discuss this also somewhere in "Advanced Users" for users that upgrade from 1.x and expect this functionality.

using-timescaledb/continuous-aggregates.md
```diff
@@ -18,118 +18,154 @@ re-calculate all of the data on every refresh. Only new and/or
...
+used to avoid aggregating the data during creation.
+
+You could then add a policy to control the refresh of the recent data,
+and also run manual refreshes using
+[`refresh_continuous_aggregate`][refresh_continuous_aggregate] to
+refresh the historical data in a controlled manner. For example, to
+refresh one month of data you could write:
+
+```sql
+CALL refresh_continuous_aggregate('conditions_summary_daily, '2020-05-01', '2020-06-01');
```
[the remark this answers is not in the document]
erimatnor: PostgreSQL has functions and procedures. The former is called using `SELECT` and the latter with `CALL`. The difference between functions and procedures is, among other things, that procedures have transactional control, i.e., they can run across several transactions, and can therefore also not return any result.

The way refresh works, it needs to run across several transactions, and therefore must be a procedure. FWIW, we also have procedures in the jobs API.

using-timescaledb/continuous-aggregates.md
erimatnor: FWIW, 1 month is not equal to 30 days. If you want to policy to be deterministic, then maybe 30 days is actually more appropriate. I don't feel strongly though, as this is not the focus.

using-timescaledb/continuous-aggregates.md
erimatnor: I would not say that we overwrite with NULL values since that implies we actually write values as opposed to deleting.

using-timescaledb/continuous-aggregates.md
erimatnor: I think this whole section is bloated, complicated, and confusing. I suggest folding this into the intro. Probably best left for a separate PR though.

using-timescaledb/continuous-aggregates.md
```diff
@@ -160,123 +275,91 @@ something like:
...
+materialized data in this way, one gets accurate and up-to-date
+results while still enjoying the speedups of pre-computing a large
+portion of the result.
+
+As an example, continuous aggregates _without_ this real-time capability make
+it really fast to get aggregate answers by pre-computing these values (such as
+the min/max/average value over each hour). This way, if you are collecting raw
+data every second, querying hourly data over the past week means reading 24 x 7
+= 168 values from the database, as opposed to processing 60 x 60 x 24 x 7 =
+604,800 values at query time.
```
erimatnor, line 300: Do we need this? Makes my head spin.

using-timescaledb/continuous-aggregates.md
erimatnor: The last use case (3) is commonly requested among users. For instance here: https://github.com/timescale/timescaledb/issues/1634#issuecomment-633861621

I think it would be a mistake to not include this somehow. But it need not happen as part of this PR. I imagine we'll continuous tweaking.

using-timescaledb/continuous-aggregates.md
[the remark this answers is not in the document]
mfreed: Sounds good.  So if it's not, this should be clarified in the docs API.  Thanks!

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +175,98 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+
+#### Automatic refresh with a continuous aggregate policy
+
+You can refresh the continuous aggregate view manually as mentioned
+above, but you can also automate the refresh by adding a continuous
+aggregate policies to automatically refresh data.
+
+There are a few situations that you might want to automate:
+
+- You want the continuous aggregate to be up to date with the
```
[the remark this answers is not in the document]
mkindahl: Reformulated the points, but they are already describe later in the text.

using-timescaledb/continuous-aggregates.md
```diff
@@ -139,19 +175,98 @@ PostgreSQL. In addition, TimescaleDB continuous aggregates does not
...
+	start_interval => NULL,
+	end_interval => INTERVAL '1 h',
+	schedule_interval => INTERVAL '1 h');
+```
+
+This will ensure that all data in the continuous aggregate is up to
+date with the hypertable except the last hour and also ensure that we
+do not try to refresh the last bucket of the continuous
+aggregate. Since we give an open-ended `start_interval`, any data that
+is removed from `conditions` will also be removed from
```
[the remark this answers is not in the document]
mkindahl: Mentioned `drop_chunks`.

using-timescaledb/continuous-aggregates.md
mkindahl: Set it to 1 month.

using-timescaledb/continuous-aggregates.md
mkindahl: Added a warning about this, along the lines of what @mike suggested.

using-timescaledb/continuous-aggregates.md
mkindahl: Leaving it for now. I will handle it as a separate PR.

using-timescaledb/continuous-aggregates.md
mkindahl: Added link to API documentation.

using-timescaledb/continuous-aggregates.md
mkindahl: There is none. Moved the heading so that there is only one and also removed some stuff and moved some stuff around.

using-timescaledb/continuous-aggregates.md
mkindahl: Removed part of this text and moved some to the discussion about the continuous aggregates above.

using-timescaledb/continuous-aggregates.md
mkindahl: Added this comment to the warning above, and the preceding text to the discussion on using continuous aggregates.

using-timescaledb/continuous-aggregates.md
mkindahl: Most comes from old documentation, we can easily remove it, but why is setting a primary key on time a problem?

mfreed:
We probably need to add a section to bottom of 2.0 docs on "Migrating from earlier continuous aggregates" or something, to have a built-in tutorial for migration from 1.x to 2.x

using-timescaledb/continuous-aggregates.md
mfreed: Erik doesn't like my example =]

mfreed (approved):
I think good as version, might update further in future PRs.

PR #496, opened 2020-09-17
Update continuous aggregate API functions

mkindahl:
Updating API documentation to fix continuous aggregate constructions.

### DIFF
--- a/api.md
+++ b/api.md
@@ -1644,20 +1644,21 @@ at predefined intervals and materialize the results. This is suitable for
 frequently used queries. For a more detailed discussion of this capability,
 please see [using TimescaleDB Continuous Aggregates][using-continuous-aggs].
 
-*  [CREATE VIEW](#continuous_aggregate-create_view)
-*  [ALTER VIEW](#continuous_aggregate-alter_view)
+*  [CREATE MATERIALIZED VIEW](#continuous_aggregate-create_view)
+*  [ALTER MATERIALIZED VIEW](#continuous_aggregate-alter_view)
 *  [REFRESH MATERIALIZED VIEW](#continuous_aggregate-refresh_view)
-*  [DROP VIEW](#continuous_aggregate-drop_view)
+*  [DROP MATERIALIZED VIEW](#continuous_aggregate-drop_view)
 
-## CREATE VIEW (Continuous Aggregate) :community_function: [](continuous_aggregate-create_view)
-`CREATE VIEW` statement is used to create continuous aggregates.
+## CREATE MATERIALIZED VIEW (Continuous Aggregate) :community_function: [](continuous_aggregate-create_view)
+`CREATE MATERIALIZED VIEW` statement is used to create continuous aggregates.
 
 The syntax is:
 ``` sql
-CREATE VIEW <view_name> [ ( column_name [, ...] ) ]
-WITH ( timescaledb.continuous [, timescaledb.<option> = <value> ] )
-AS
-<select_query>
+CREATE MATERIALIZED VIEW <view_name> [ ( column_name [, ...] ) ]
+  WITH ( timescaledb.continuous [, timescaledb.<option> = <value> ] )
+  AS
+    <select_query>
+  [WITH [NO] DATA]
 ```
 
 `<select_query>` is of the form :
@@ -1685,54 +1686,40 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
 |---|---|---|
 |`timescaledb.continuous`|||
 |**Description**|**Type**|**Default**|
-|If timescaledb.continuous is not specified, then this is a regular PostgresSQL view. | `BOOLEAN` ||
+|If timescaledb.continuous is not specified, then this is a regular PostgresSQL materialized view. | `BOOLEAN` ||
 
 #### Optional `WITH` clause options [](continuous_aggregate-create_view-with-optional)
 
 |**Name**|||
 |---|---|---|
-|`timescaledb.refresh_lag`|||
-|**Description**|**Type**|**Default**|
-| Refresh lag controls the amount by which the materialization will lag behind the current time. The continuous aggregate view lags behind by `bucket_width` + `refresh_lag` value. `refresh_lag` can be set to positive and negative values. | Same datatype as the `bucket_width` argument from the `time_bucket` expression.| The default value is twice the bucket width (as specified by the `time_bucket` expression).|
-|   |   |   |
-|**Name**|||
-|`timescaledb.refresh_interval`|||
-|**Description**|**Type**|**Default**|
-| Refresh interval controls how often the background materializer is run. Note that if `refresh_lag` is set to `-<bucket_width>`, the continuous aggregate will run whenever new data is received, regardless of what the `refresh_interval` value is. | `INTERVAL`|By default, this is set to twice the bucket width (if the datatype of the bucket_width argument from the `time_bucket` expression is an `INTERVAL`), otherwise it is set to 12 hours.|
-|   |   |   |
-|**Name**|||
 |`timescaledb.materialized_only`|||
 |**Description**|**Type**|**Default**|
-| Return only materialized data when querying the continuous aggregate view. | `BOOLEAN` | false |
+| Return only materialized data when querying the continuous aggregate view. See more in section on [real-time aggregates][real-time-aggregates]. | `BOOLEAN` | false |
 |   |   |   |
-|**Name**|||
-|`timescaledb.max_interval_per_job`|||
-|**Description**|**Type**|**Default**|
-| Max interval per job specifies the amount of data processed by the background materializer job when the continuous aggregate is updated. | Same datatype as the `bucket_width` argument from the `time_bucket` expression.| The default value is `20 * bucket width`.|
-|   |   |   |
-|**Name**|||
 |`timescaledb.create_group_indexes`|||
 |**Description**|**Type**|**Default**|
 | Create indexes on the materialization table for the group by columns (specified by the `GROUP BY` clause of the `SELECT` query). | `BOOLEAN` | Indexes are created by default for every group by expression + time_bucket expression pair.|
-|   |   |   |
-|**Name**|||
-|`timescaledb.ignore_invalidation_older_than`|||
-|**Description**|**Type**|**Default**|
-| Time interval after which invalidations are ignored.| Same datatype as the `bucket_width` argument from the `time_bucket` expression. | By default all invalidations are processed.|
 
->:TIP: Say, the continuous aggregate uses time_bucket(INTERVAL '2h', time_column) and we want to keep the view up to date with the data. We can do this by modifying the `refresh_lag` setting. Set refresh_lag to `-2h`. E.g. `ALTER VIEW contview set (timescaledb.refresh_lag = '-2h');` Please refer to the [caveats][].
+#### Notes
 
-#### Restrictions
+- The view will be automatically refreshed (as outlined under
+  [`refresh_continuous_aggregate`](#refresh_continuous_aggregate))
+  unless `WITH NO DATA` is given (`WITH DATA` is the default).
 - `SELECT` query should be of the form specified in the syntax above.
-- The hypertable used in the `SELECT` may not have [row-level-security policies][postgres-rls] enabled.
--  `GROUP BY` clause must include a time_bucket expression. The [`time_bucket`](#time_bucket) expression must use the time dimension column of the hypertable.
+- The hypertable used in the `SELECT` may not have [row-level-security
+  policies][postgres-rls] enabled.
+-  `GROUP BY` clause must include a time_bucket expression. The
+   [`time_bucket`](#time_bucket) expression must use the time
+   dimension column of the hypertable.
 - [`time_bucket_gapfill`](#time_bucket_gapfill) is not allowed in continuous
   aggs, but may be run in a `SELECT` from the continuous aggregate view.
-- In general, aggregates which can be [parallelized by PostgreSQL][postgres-parallel-agg] are allowed in the view definition, this
-  includes most aggregates distributed with PostgreSQL. Aggregates with `ORDER BY`,
-  `DISTINCT` and `FILTER` clauses are not permitted.
-* All functions and their arguments included in `SELECT`, `GROUP BY` and `HAVING` clauses must be [immutable][postgres-immutable].
-- Queries with `ORDER BY` are disallowed.
+- In general, aggregates which can be [parallelized by
+  PostgreSQL][postgres-parallel-agg] are allowed in the view
+  definition, this includes most aggregates distributed with
+  PostgreSQL. Aggregates with `ORDER BY`, `DISTINCT` and `FILTER`
+  clauses are not permitted.
+- All functions and their arguments included in `SELECT`, `GROUP BY`
+  and `HAVING` clauses must be [immutable][postgres-immutable].
 - The view is not allowed to be a [security barrier view][postgres-security-barrier].
 
 [postgres-immutable]:https://www.postgresql.org/docs/current/xfunc-volatility.html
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -1874,6 +1862,44 @@ are meant to implement data retention or perform tasks that will improve query
 performance on older chunks. Each policy is assigned a scheduled job
 which will be run in the background to enforce it.
 
+## add_continuous_aggregate_policy() :community_function> [](add_continuous_aggregate_policy)
+
+Create a policy that refresh a windows in a continuous aggregate.
```
erimatnor, line 1867:
```suggestion
Create a policy that automatically refreshes a continuous aggregate.
```

api.md
```diff
@@ -1874,6 +1862,44 @@ are meant to implement data retention or perform tasks that will improve query
...
+
+#### Required Arguments [](add_continuous_aggregate_policy-required-arguments)
+
+|Name|Description|
+|---|---|
+| `continuous_aggregate` | (REGCLASS) The continuous aggregate to add the policy for. |
+| `start_interval` | (INTERVAL) Start of the window relative to current time to refresh. |
+| `end_interval` | (INTERVAL) End of the window relative to current time to refresh. |
```
erimatnor, line 1875:
```suggestion
| `start_interval` | (INTERVAL) Start of the refresh window relative to the time when the policy is executed. |
| `end_interval` | (INTERVAL) End of the refresh window relative to the time when the policy is executed. |
```

api.md
```diff
@@ -1874,6 +1862,44 @@ are meant to implement data retention or perform tasks that will improve query
...
+| `schedule_interval` | (INTERVAL) Interval between refresh executions. |
+
+#### Optional Arguments [](add_continuous_aggregate_policy-optional-arguments)
+
+|Name|Description|
+|---|---|
+| `if_not_exists` | (BOOLEAN) Set to true to avoid throwing an error if the continuous aggregate_policy already exists. A notice is issued instead. Defaults to false. |
```
erimatnor, line 1882:
```suggestion
| `if_not_exists` | (BOOLEAN) Set to true to avoid throwing an error if the continuous aggregate policy already exists. A notice is issued instead. Defaults to false. |
```

api.md
```diff
@@ -1854,13 +1829,26 @@ DROP VIEW <view_name> CASCADE;
...
+buckets are partially outside the window either at the beginning or
+the end will be excluded from the refresh.
+
+#### Required Arguments [](refresh_continuous_aggregate-required-arguments)
+
+|Name|Description|
+|---|---|
+| `continuous_aggregate` | (REGCLASS) The continuous aggregate to refresh. |
+| `window_start` | (INTERVAL) Start of the window relative to current time to refresh. |
+| `window_end` | (INTERVAL) End of the window relative to current time to refresh. |
```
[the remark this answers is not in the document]
erimatnor: I also think we need to describe somewhere what reasonable values for these parameters are. For instance, window_start should be before window_end. And, schedule_interval should (must) be smaller than window start. Perhaps for a later PR, but I am worried that this will be lost.

api.md
```diff
@@ -1874,6 +1862,44 @@ are meant to implement data retention or perform tasks that will improve query
...
+|`job_id`| (INTEGER)  TimescaleDB background job id created to implement this policy|
+
+
+#### Sample Usage [](add_continuous_aggregate_policy-examples)
+
+To add a policy that refreshes the last month once an hour:
+```sql
+SELECT add_continuous_aggregate_policy('conditions_summary',
+	start_interval => INTERVAL '1 month',
+	end_interval => INTERVAL '1 hour',
```
erimatnor, line 1897: The setting of end_interval is not motivated. And we never describe why you'd want to have a refresh offset/lag like this.

mkindahl: Not sure what you mean. All the parameters given here are mandatory and have to be given.

api.md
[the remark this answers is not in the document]
mkindahl: I have deliberately skipped discussing reasonable values for the parameters here since it is not clear what we should recommend. I think we can handle it in a later PR. I added a note about that to the documentation issue.

api.md
```diff
@@ -1854,13 +1829,26 @@ DROP VIEW <view_name> CASCADE;
...
-DROP VIEW contagg_view CASCADE;
+DROP MATERIALIZED VIEW contagg_view;
 ```
->:WARNING: `CASCADE` will drop those objects that depend on the continuous
-aggregate, such as views that are built on top of the continuous aggregate view.
 
+## refresh_continuous_aggregate() :community_function: [](refresh_continuous_aggregate)
+
+Refresh all buckets of a continuous aggregate inside a window. Note
+that only buckets fully in the window will be refreshed, so if any
```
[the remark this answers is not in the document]
mkindahl: Removed it.

api.md
erimatnor: Above you say `add a policy that refreshes the last month once an hour`, which isn't strictly true since you have an "offset" of 1 hour via the `end_interval` (which takes 1 hour off of the month). This begs the question: "why is there an offset of 1 hour?". We never explain in the API documentation or the example why you'd like to have this 1 hour offset, so it just comes out of nowhere. It is there to avoid write amplification, but no such context or explanation is given.

The other parameters are explained, self-evident, or somewhat motivated by the text and given context.

api.md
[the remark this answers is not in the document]
erimatnor: I am not talking about recommended "defaults", but rather configuration that doesn't make sense. For instance, for the policy to make sense:

* window_start should be larger than window_end
* schedule_interval should be smaller than window_start

and so on.

api.md
mkindahl: It is not clear if you want:
* To rewrite the documentation and add a discussion about the write amplification. I have deliberately skipped this because 1) it requires a discussion about recommended values and what yo use, which is not obvious, and 2) it is not suitable material for the API reference documentation and is more suitable for the "Using" section.
* To remove `end_interval`, which is not possible because it is mandatory to the function.
* Set `end_interval` to NULL, which raises a discussion about the write amplification again, and the API documentation is not a suitable place to discuss such things.

api.md
```diff
@@ -1874,6 +1860,44 @@ are meant to implement data retention or perform tasks that will improve query
...
+#### Returns [](add_continuous_aggregate_policy-returns)
+
+|Column|Description|
+|---|---|
+|`job_id`| (INTEGER)  TimescaleDB background job id created to implement this policy|
+
+
+#### Sample Usage [](add_continuous_aggregate_policy-examples)
+
+To add a policy that refreshes the last month once an hour:
```
erimatnor, line 1891:
```suggestion
Add a policy that refreshes the last month once an hour, excluding the latest hour from the aggregate (for performance reasons, it is recommended to exclude buckets that still see lots of writes):
```

The above can be complemented by linking to a deeper discussion on recommended settings.

erimatnor (approved):

api.md
```diff
@@ -1685,53 +1686,41 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
...
 |**Description**|**Type**|**Default**|
 | Create indexes on the materialization table for the group by columns (specified by the `GROUP BY` clause of the `SELECT` query). | `BOOLEAN` | Indexes are created by default for every group by expression + time_bucket expression pair.|
-|   |   |   |
-|**Name**|||
-|`timescaledb.ignore_invalidation_older_than`|||
-|**Description**|**Type**|**Default**|
-| Time interval after which invalidations are ignored.| Same datatype as the `bucket_width` argument from the `time_bucket` expression. | By default all invalidations are processed.|
 
->:TIP: Say, the continuous aggregate uses time_bucket(INTERVAL '2h', time_column) and we want to keep the view up to date with the data. We can do this by modifying the `refresh_lag` setting. Set refresh_lag to `-2h`. E.g. `ALTER VIEW contview set (timescaledb.refresh_lag = '-2h');` Please refer to the [caveats][].
+#### Notes
```
mfreed, line 1703: @solugebefola Do you have thoughts on any standard template format for API docs with such type of "notes"?

api.md
```diff
@@ -1685,53 +1686,41 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
...
-  `DISTINCT` and `FILTER` clauses are not permitted.
-* All functions and their arguments included in `SELECT`, `GROUP BY` and `HAVING` clauses must be [immutable][postgres-immutable].
+- The hypertable used in the `SELECT` may not have [row-level-security
+  policies][postgres-rls] enabled.
+-  `GROUP BY` clause must include a time_bucket expression. The
+   [`time_bucket`](#time_bucket) expression must use the time
+   dimension column of the hypertable.
+- [`time_bucket_gapfill`](#time_bucket_gapfill) is not allowed in
+  continuous aggs, but may be run in a `SELECT` from the continuous
+  aggregate view.
```
mfreed, line 1716:
```suggestion
- [`time_bucket_gapfill`](#time_bucket_gapfill) is not allowed in
  continuous aggs, but may be executed by a `SELECT` on the continuous
  aggregate view.
```

api.md
```diff
@@ -1746,23 +1735,17 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
...
-CREATE VIEW continuous_aggregate_view( timec, minl, sumt, sumh )
-WITH ( timescaledb.continuous,
-    timescaledb.refresh_lag = '5 hours',
-    timescaledb.refresh_interval = '1h' )
-AS
+CREATE MATERIALIZED VIEW continuous_aggregate_view( timec, minl, sumt, sumh )
+WITH (timescaledb.continuous) AS
     SELECT time_bucket('1day', timec), min(location), sum(temperature), sum(humidity)
         FROM conditions
         GROUP BY time_bucket('1day', timec)
```
mfreed, line 1742: I would standardize on 2 space indents.  The section actually has a mix (this is old issue =)

api.md
```diff
@@ -1685,53 +1686,41 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
...
-|   |   |   |
-|**Name**|||
-|`timescaledb.refresh_interval`|||
-|**Description**|**Type**|**Default**|
-| Refresh interval controls how often the background materializer is run. Note that if `refresh_lag` is set to `-<bucket_width>`, the continuous aggregate will run whenever new data is received, regardless of what the `refresh_interval` value is. | `INTERVAL`|By default, this is set to twice the bucket width (if the datatype of the bucket_width argument from the `time_bucket` expression is an `INTERVAL`), otherwise it is set to 12 hours.|
-|   |   |   |
-|**Name**|||
 |`timescaledb.materialized_only`|||
 |**Description**|**Type**|**Default**|
 | Return only materialized data when querying the continuous aggregate view. | `BOOLEAN` | false |
```
mfreed, line 1697: Should this section have any pointer to [real-time aggregates][real-time-aggregates] ?  I don't think the reader just looking up this API spec will understand this term.  If so, where should the pointer go?

api.md
```diff
@@ -1829,22 +1809,16 @@ Update the continuous aggregate view immediately.
...
-## DROP VIEW (Continuous Aggregate) :community_function: [](continuous_aggregate-drop_view)
-Continuous aggregate views can be dropped using `DROP VIEW` statement.
+## DROP MATERIALIZED VIEW (Continuous Aggregate) :community_function: [](continuous_aggregate-drop_view)
+Continuous aggregate views can be dropped using the `DROP MATERIALIZED VIEW` statement.
 
-This deletes the hypertable that stores the materialized data for the
-continuous aggregate; it does not affect the data in the underlying hypertable
-from which the continuous aggregate is derived (i.e., the raw data).  The
-`CASCADE` parameter is required for this command.
+This statement deletes the continuous aggregate and all its internal objects. To also delete other dependent objects, add the `CASCADE` option. Dropping a continuous aggregate does not affect the data in the
```
mfreed, line 1817: Can you give an example of what dependent objects might be?

"To also delete other dependent objects (e.g., <...>), add the ..."

api.md
```diff
@@ -1854,13 +1828,24 @@ DROP VIEW <view_name> CASCADE;
...
-DROP VIEW contagg_view CASCADE;
+DROP MATERIALIZED VIEW contagg_view;
 ```
->:WARNING: `CASCADE` will drop those objects that depend on the continuous
-aggregate, such as views that are built on top of the continuous aggregate view.
 
+## refresh_continuous_aggregate() :community_function: [](refresh_continuous_aggregate)
+
+Refresh all buckets of a continuous aggregate between two points of
+time. Note that the function expects the parameter values to have the same time type as used in the continuous aggregate's time bucket expression.
```
mfreed, line 1837:
```suggestion
Refresh all buckets of a continuous aggregate between two points of
time. 

The function expects the parameter values to have the same time type
as used in the continuous aggregate's time bucket expression (e.g., if
the time bucket specifies in `timestamptz`, then the start and end time
supplied should also be `timestamptz`).
```

api.md
mfreed: (Not sure exactly what @erimatnor is proposing.  Agree that longer discussion should be in a "Using" section.  Why don't we revisit after we have a draft of that section, to see how we might reference it?)

erimatnor: @mfreed my suggested change was already adopted as part of another comment, which is no longer visible.

api.md
mkindahl: Added an example as subphrase.

api.md
mkindahl: I reformatted this code block and the next code block.

api.md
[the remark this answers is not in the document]
mkindahl: I added a notice about this to the `add_continuous_aggregate_policy` but I do not understand how this affects `refresh_continuous_aggregate`. It might not even be the case that there is a policy defined.

api.md
mkindahl: Added a pointer to the section on real-time aggregates.

api.md
```diff
@@ -1829,22 +1809,20 @@ Update the continuous aggregate view immediately.
...
-Continuous aggregate views can be dropped using `DROP VIEW` statement.
+## DROP MATERIALIZED VIEW (Continuous Aggregate) :community_function: [](continuous_aggregate-drop_view)
+Continuous aggregate views can be dropped using the `DROP MATERIALIZED VIEW` statement.
 
-This deletes the hypertable that stores the materialized data for the
-continuous aggregate; it does not affect the data in the underlying hypertable
-from which the continuous aggregate is derived (i.e., the raw data).  The
-`CASCADE` parameter is required for this command.
+This statement deletes the continuous aggregate and all its internal
+objects. To also delete other dependent objects&mdash;such as a view
```
mfreed, line 1818: Does `&mdash;` work in the markdown parser we are using?

api.md
```diff
@@ -1685,53 +1686,41 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
...
-|**Name**|||
-|`timescaledb.refresh_interval`|||
-|**Description**|**Type**|**Default**|
-| Refresh interval controls how often the background materializer is run. Note that if `refresh_lag` is set to `-<bucket_width>`, the continuous aggregate will run whenever new data is received, regardless of what the `refresh_interval` value is. | `INTERVAL`|By default, this is set to twice the bucket width (if the datatype of the bucket_width argument from the `time_bucket` expression is an `INTERVAL`), otherwise it is set to 12 hours.|
-|   |   |   |
-|**Name**|||
 |`timescaledb.materialized_only`|||
 |**Description**|**Type**|**Default**|
-| Return only materialized data when querying the continuous aggregate view. | `BOOLEAN` | false |
+| Return only materialized data when querying the continuous aggregate view. See more in section on [real-time aggregates][real-time-aggregates]. | `BOOLEAN` | false |
```
mfreed, line 1697: Have you done a local build, and does the visual format support such a wide line in table?

api.md
```diff
@@ -1685,53 +1686,41 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
...
+  continuous aggs, but may be executed by a `SELECT` on the continuous
+  aggregate view.
+- In general, aggregates which can be [parallelized by
+  PostgreSQL][postgres-parallel-agg] are allowed in the view
+  definition, this includes most aggregates distributed with
+  PostgreSQL. Aggregates with `ORDER BY`, `DISTINCT` and `FILTER`
+  clauses are not permitted.
+- All functions and their arguments included in `SELECT`, `GROUP BY`
+  and `HAVING` clauses must be [immutable][postgres-immutable].
 - Queries with `ORDER BY` are disallowed.
```
mfreed, line 1724: Minor nit:  What's the difference between these line:

     Queries with `ORDER BY` are disallowed.

And the above:

    Aggregates with `ORDER BY`, `DISTINCT` and `FILTER` clauses are not permitted.

api.md
```diff
@@ -1854,13 +1832,29 @@ DROP VIEW <view_name> CASCADE;
...
+the time bucket specifies in `timestamptz`, then the start and end time
+supplied should also be `timestamptz`).
+
+#### Required Arguments [](refresh_continuous_aggregate-required-arguments)
+
+|Name|Description|
+|---|---|
+| `continuous_aggregate` | (REGCLASS) The continuous aggregate to refresh. |
+| `window_start` | Start of the window to refresh, has to be before `window_end`. |
+| `window_end` | End of the window to refresh, has to be after `window_start`. |
```
mfreed, line 1855: Any sample usage?

mfreed (approved):
A few small nits, but otherwise LGTM.

api.md
```diff
@@ -1874,6 +1868,50 @@ are meant to implement data retention or perform tasks that will improve query
...
+Create a policy that automatically refreshes a continuous aggregate.
+
+#### Required Arguments [](add_continuous_aggregate_policy-required-arguments)
+
+|Name|Description|
+|---|---|
+| `continuous_aggregate` | (REGCLASS) The continuous aggregate to add the policy for. |
+| `start_interval` | Start of the refresh window as an interval relative to the time when the policy is executed. |
+| `end_interval` | End of the refresh window as an interval relative to the time when the policy is executed. |
+| `schedule_interval` | Interval between refresh executions in wall-clock time. |
```
mfreed, line 1882: Question:  was there discussion to make `schedule_interval` an optional argument, default to to the time_bucket of the continuous agg?

api.md
mkindahl: I don't know for sure, but this is what the reference says (https://daringfireball.net/projects/markdown/syntax#html):
> For any markup that is not covered by Markdown’s syntax, you simply use HTML itself. There’s no need to preface it or delimit it to indicate that you’re switching from Markdown to HTML; you just use the tags.

api.md
mkindahl: Seems like it was accidentally duplicated. Removed the second one.

api.md
mkindahl: Nope, I have not done a local build, but there's been quite long lines elsewhere.

api.md
mkindahl: There has been discussions about if any of these parameters should be made optional, but for the time being, they are mandatory. If it changes, we have to change the documentation.

api.md
mkindahl: Added one.

PR #498, opened 2020-09-17
Add custom job documentation

svenklemm:


using-timescaledb/custom_jobs.md
```diff
@@ -0,0 +1,214 @@
+# Custom Jobs
```
svenklemm, line 1: @mfreed Should we call this custom policies?

using-timescaledb/custom_jobs.md
```diff
@@ -0,0 +1,214 @@
...
+
+Reenables automatic scheduling of the job with id 1000.
+
+```sql
+SELECT delete_job(1000);
+```
+
+Deletes the job with id 1000 from the automation framework.
+
+## Generic Retention Policy
```
k-rus, line 81: I suggest to have a separate section for examples and then level down each example. For example:
```suggestion
## Examples of custom policies

This section provides several example how to create custom policies.

### Generic Retention Policy
```

using-timescaledb/custom_jobs.md
```diff
@@ -0,0 +1,214 @@
...
+
```
k-rus, line 82: Can you describe the example before the code, so it will be easier for the reader to follow the code?

using-timescaledb/custom_jobs.md
```diff
@@ -0,0 +1,214 @@
...
+  lag INTERVAL;
+  chunk REGCLASS;
+BEGIN
+  SELECT jsonb_object_field_text(config, 'lag')::interval INTO STRICT lag;
+
+  IF lag IS NULL THEN
+	  RAISE EXCEPTION 'Config must have lag';
+  END IF;
+
+	PERFORM drop_chunks(format('%I.%I',table_schema,table_name), older_than => lag) FROM timescaledb_information.hypertables;
```
k-rus, line 97: Should it be the same indentation as `IF`?

k-rus (changes requested):
Great start! Few notices:

This PR introduces new doc page. Can you make this page reachable? It seems to me it should be added to the TOC, so it appears in using section. If I don't mistake, this page needs to be updated https://github.com/timescale/docs.timescale.com-content/blob/master/page-index/page-index.js

Existing pages in Using section describes code examples before the code. I think this page should not be an exception.

Another comment that this page should link to function descriptions in API doc. I guess it is necessary to guess the links until the other page is added.

Can you added the examples as regression tests to TimescaleDB if it wasn't done yet?

PR #499, opened 2020-09-21
Add size related functions

gayyappan:
Add hypertable_detailed_size, chunks_detailed_size.
Add hypertable_size, hypertable_index_size.
Remove chunk_relation_size, chunk_relation_size_pretty,
hypertable_relation_size, hypertable_relation_size_pretty.
Remove indexes_relation_size.

To reviewers:
This PR has 2 commits.

### DIFF
--- a/api.md
+++ b/api.md
@@ -16,8 +16,7 @@
 > - [attach_tablespace](#attach_tablespace)
 > - [block_new_chunks](#block_new_chunks)
 > - [chunk_compression_stats](#chunk_compression_stats)
-> - [chunk_relation_size](#chunk_relation_size)
-> - [chunk_relation_size_pretty](#chunk_relation_size_pretty)
+> - [chunks_detailed_size](#chunks_detailed_size)
 > -	[compress_chunk](#compress_chunk)
 > - [create_distributed_hypertable](#create_distributed_hypertable)
 > - [create_hypertable](#create_hypertable)
@@ -35,10 +34,9 @@
 > - [get_telemetry_report](#get_telemetry_report)
 > - [histogram](#histogram)
 > - [hypertable_compression_stats](#hypertable_compression_stats)
-> - [hypertable_relation_size](#hypertable_relation_size)
-> - [hypertable_relation_size_pretty](#hypertable_relation_size_pretty)
-> - [indexes_relation_size](#indexes_relation_size)
-> - [indexes_relation_size_pretty](#indexes_relation_size_pretty)
+> - [hypertable_detailed_size](#hypertable_size)
+> - [hypertable_index_size](#hypertable_index_size)
+> - [hypertable_size](#hypertable_size)
 > - [interpolate](#interpolate)
 > - [last](#last)
 > - [locf](#locf)
@@ -673,7 +671,7 @@ corresponds to increased planning latency for some types of queries.
 both the underlying data size *and* any indexes, so some care might be
 taken if you make heavy use of expensive index types (e.g., some
 PostGIS geospatial indexes).  During testing, you might check your
-total chunk sizes via the [`chunk_relation_size`](#chunk_relation_size)
+total chunk sizes via the [`chunks_detailed_size`](#chunks_detailed_size)
 function.
 
 **Space partitions:** In most cases, it is advised for users not to use
@@ -3138,98 +3136,6 @@ total_failures         | 0
 ```
 ---
 
-## chunk_relation_size() [](chunk_relation_size)
-
-Get relation size of the chunks of an hypertable.
-
-#### Required Arguments [](chunk_relation_size-required-arguments)
-
-|Name|Description|
-|---|---|
-| `main_table` | Identifier of hypertable to get chunk relation sizes for.|
-
-#### Returns [](chunk_relation_size-returns)
-|Column|Description|
-|---|---|
-|chunk_id|TimescaleDB id of a chunk|
-|chunk_table|Table used for the chunk|
-|partitioning_columns|Partitioning column names|
-|partitioning_column_types|Types of partitioning columns|
-|partitioning_hash_functions|Hash functions of partitioning columns|
-|dimensions|Partitioning dimension names|
-|ranges|Partitioning ranges for each dimension|
-|table_bytes|Disk space used by main_table|
-|index_bytes|Disk space used by indexes|
-|toast_bytes|Disk space of toast tables|
-|total_bytes|Disk space used in total|
-
-#### Sample Usage [](chunk_relation_size-examples)
-```sql
-SELECT * FROM chunk_relation_size('conditions');
-```
-or, to reduce the output, a common use is:
-```sql
-SELECT chunk_table, table_bytes, index_bytes, total_bytes
-FROM chunk_relation_size('conditions');
-```
-The expected output:
-```
-                 chunk_table                 | table_bytes | index_bytes | total_bytes
----------------------------------------------+-------------+-------------+-------------
- "_timescaledb_internal"."_hyper_1_1_chunk"  |    29220864 |    37773312 |    67002368
- "_timescaledb_internal"."_hyper_1_2_chunk"  |    59252736 |    81297408 |   140558336
- ...
-```
-
-Where `chunk_table` is the table that contains the data, `table_bytes` is the size of that table, `index_bytes` is the size of the indexes of the table, and `total_bytes` is the size of the table with indexes.
-
----
-
-## chunk_relation_size_pretty() [](chunk_relation_size_pretty)
-
-Get relation size of the chunks of an hypertable.
-
-#### Required Arguments [](chunk_relation_size_pretty-required-arguments)
-
-|Name|Description|
-|---|---|
-| `main_table` | Identifier of hypertable to get chunk relation sizes for.|
-
-#### Returns [](chunk_relation_size_pretty-returns)
-|Column|Description|
-|---|---|
-|chunk_id|TimescaleDB id of a chunk|
-|chunk_table|Table used for the chunk|
-|partitioning_columns|Partitioning column names|
-|partitioning_column_types|Types of partitioning columns|
-|partitioning_hash_functions|Hash functions of partitioning columns|
-|ranges|Partitioning ranges for each dimension|
-|table_size|Pretty output of table_bytes|
-|index_size|Pretty output of index_bytes|
-|toast_size|Pretty output of toast_bytes|
-|total_size|Pretty output of total_bytes|
-
-#### Sample Usage [](chunk_relation_size_pretty-examples)
-```sql
-SELECT * FROM chunk_relation_size_pretty('conditions');
-```
-or, to reduce the output, a common use is:
-```sql
-SELECT chunk_table, table_size, index_size, total_size
-FROM chunk_relation_size_pretty('conditions');
-```
-The expected output:
-```
-                chunk_table                  | table_size | index_size | total_size
----------------------------------------------+------------+------------+------------
- "_timescaledb_internal"."_hyper_1_1_chunk"  | 28 MB      | 36 MB      | 64 MB
- "_timescaledb_internal"."_hyper_1_2_chunk"  | 57 MB      | 78 MB      | 134 MB
- ...
-```
-Where `chunk_table` is the table that contains the data, `table_size` is the size of that table, `index_size` is the size of the indexes of the table, and `total_size` is the size of the table with indexes.
-
----
-
 ## get_telemetry_report() [](get_telemetry_report)
 
 If background [telemetry][] is enabled, returns the string sent to our servers.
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -35,10 +34,9 @@
 > - [get_telemetry_report](#get_telemetry_report)
 > - [histogram](#histogram)
 > - [hypertable_compression_stats](#hypertable_compression_stats)
-> - [hypertable_relation_size](#hypertable_relation_size)
-> - [hypertable_relation_size_pretty](#hypertable_relation_size_pretty)
-> - [indexes_relation_size](#indexes_relation_size)
-> - [indexes_relation_size_pretty](#indexes_relation_size_pretty)
+> - [hypertable_detailed_size](#hypertable_size)
+> - [hypertable_size](#hypertable_size)
+> - [hypertable_index_size](#hypertable_index_size)
```
mfreed, line 39:
```suggestion
> - [hypertable_detailed_size](#hypertable_detailed_size)
> - [hypertable_index_size](#hypertable_index_size)
> - [hypertable_size](#hypertable_size)
```

api.md
```diff
@@ -3251,7 +3157,7 @@ SELECT get_telemetry_report(always_display_report := true);
 ```
 ---
 
-## approximate_row_count() [](approximate_row_count)
+## approximate_row_count() :community_function:  [](approximate_row_count)
```
mfreed, line 3160: Isn't this in Apache-2 codebase, not TSL?  If so, it shoudn't be marked :community_function:

api.md
```diff
@@ -3406,138 +3312,143 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
 total | 48 kB
 
 ```
-## hypertable_relation_size() [](hypertable_relation_size)
+---
+
+## hypertable_detailed_size() :community_function: [](hypertable_detailed_size)
```
mfreed, line 3317: Isn't this in Apache-2 codebase, not TSL?  If so, it shoudn't be marked :community_function:

(Same comment throughout)

api.md
```diff
@@ -3406,138 +3312,143 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
...
 
-Get relation size of hypertable like `pg_relation_size(hypertable)`.
+Get size of hypertable like `pg_relation_size(hypertable)`.
+If this is a distributed hypertable, the function returns size
+information per node. All sizes are reported in bytes.
```
mfreed, line 3321:
```suggestion
Get size of hypertable like `pg_relation_size(hypertable)`, returning 
size information for the table itself, any indexes on the table, any 
toast tables, and the total size of all. All sizes are reported in bytes.

If this is a distributed hypertable, the function returns size
information as a separate row per node. 
```

api.md
```diff
@@ -3406,138 +3312,143 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
...
+chunk, if this is a distributed hypertable. Meta data for a chunk can be
+viewed using `timescaledb_information.chunks` view.
 
-#### Required Arguments [](hypertable_relation_size_pretty-required-arguments)
+#### Required Arguments [](chunks_detailed_size-required-arguments)
 
 |Name|Description|
 |---|---|
-| `main_table` | Identifier of hypertable to get relation size for.|
+| `main_table` | Name of the hypertable|
```
mfreed, line 3363: Given the description, this only works for hypertables, and not regular tables, right?

Why isn't the name of the argument "hypertable"?

api.md
```diff
@@ -3406,138 +3312,143 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
...
+Get size of hypertable like `pg_relation_size(hypertable)`.
+If this is a distributed hypertable, the function returns size
+information per node. All sizes are reported in bytes.
 
-#### Required Arguments [](hypertable_relation_size-required-arguments)
+#### Required Arguments [](hypertable_detailed_size-required-arguments)
 
 |Name|Description|
 |---|---|
 | `main_table` | Identifier of hypertable to get relation size for.|
```
mfreed, line 3327: Why isn't the name of the argument "hypertable"?

Also, in the chunks_* docs, it describes this as:  "Name of the hypertable", but here as "Identifier of hypertable to get relation size for."

Should this just be "Name of the hypertable"?

It can be confusing if we describe things differently even though we mean the same thing.  (Also, "identifier" here is slightly odd choice of language.)

(Same question for other functions here.)

api.md
```diff
@@ -3406,138 +3312,143 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
...
 ---
 
-## hypertable_relation_size_pretty() [](hypertable_relation_size_pretty)
+## chunks_detailed_size()  :community_function: [](chunks_detailed_size)
 
-Get relation size of hypertable like `pg_relation_size(hypertable)`.
+Get per chunk size information of hypertable.
+This function will return multiple rows of output for a single
+chunk, if this is a distributed hypertable. Meta data for a chunk can be
+viewed using `timescaledb_information.chunks` view.
```
mfreed, line 3357: I like re-using language across docs for consistency whenever possible.

```suggestion
Get size information about the chunks belonging to a hypertable, returning 
size information for each chunk table itself, any indexes on the chunk, any 
toast tables, and the total size associated with the chunk. All sizes are reported in bytes.

If this is a distributed hypertable, the function returns [size
information as a separate row per node - SEE COMMENT].

Additional metadata associated with a chunk can be accessed 
via the `timescaledb_information.chunks` view.
```

I don't understand this:  
```
This function will return multiple rows of output for a single
chunk, if this is a distributed hypertable. 
```
Is this referring to the case when replication_factor > 1?  Otherwise, shouldn't there only be a single copy of a chunk, on a specific data node?

api.md
```diff
@@ -3406,138 +3312,143 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
...
-SELECT * FROM hypertable_relation_size_pretty('conditions');
-```
-or, to reduce the output, a common use is:
+#### Sample Usage [](chunks_detailed_size-examples)
 ```sql
-SELECT table_size, index_size, toast_size, total_size
-FROM hypertable_relation_size_pretty('conditions');
+SELECT * FROM chunks_detailed_size('dist_table')
+ORDER BY chunk_name, node_name;
+     chunk_schema      |      chunk_name       | table_bytes | index_bytes | toast_bytes | total_bytes |       node_name
```
mfreed, line 3380:
```suggestion
SELECT * FROM chunks_detailed_size('dist_table')
  ORDER BY chunk_name, node_name;
  
     chunk_schema      |      chunk_name       | table_bytes | index_bytes | toast_bytes | total_bytes |       node_name
```

api.md
```diff
@@ -3406,138 +3312,143 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
...
+-----------------------+-----------------------+-------------+-------------+-------------+-------------+-----------------------
+ _timescaledb_internal | _dist_hyper_1_1_chunk |        8192 |       32768 |           0 |       40960 | db_node1
+ _timescaledb_internal | _dist_hyper_1_1_chunk |        8192 |       32768 |           0 |       40960 | db_node2
```
mfreed, line 3383: See my question above about why a chunk appears on multiple data nodes?  Given that (at least for now) we don't recommend replication_factor > 1, if that's the only case for it, I would simplify this example to show replication_factor = 1.  We can extend these docs once we launch more complete support for replication.

api.md
```diff
@@ -3406,138 +3312,143 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
...
 ---
 
-## indexes_relation_size() [](indexes_relation_size)
+## hypertable_size() [](hypertable_size)
 
-Get sizes of indexes on a hypertable.
+Get the total disk space used bya  table, including all indexes and TOAST tables.
+The size is reported in bytes.
+This is equivalent to computing the sum of `total_bytes` column from the output of 
+`hypertable_detailed_size` function.
```
mfreed, line 3394: Please align language with other docs.

api.md
```diff
@@ -3406,138 +3312,143 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
...
-| `main_table` | Identifier of hypertable to get indexes size for.|
+| `main_table` | Name of the hypertable|
 
-#### Returns [](indexes_relation_size-returns)
-|Column|Description|
-|---|---|
-|index_name|Index on hypertable|
-|total_bytes|Size of index on disk|
+#### Returns [](hypertable_size-returns)
+Total disk space used by the specified table, including all indexes and TOAST data|
```
mfreed, line 3403:
```suggestion
Total disk space used by the specified table, including all indexes and TOAST data
```

api.md
```diff
@@ -3406,138 +3312,143 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
...
-```
-The expected output:
-```
-              index_name              | total_bytes
---------------------------------------+-------------
- public.conditions_device_id_time_idx |  1198620672
- public.conditions_time_idx           |   487358464
-
+SELECT hypertable_size('disttable') ;
+ hypertable_size
```
mfreed, line 3409:
```suggestion
SELECT hypertable_size('disttable') ;

 hypertable_size
```

api.md
```diff
@@ -3406,138 +3312,143 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
...
+-----------------
+           73728
 ```
-
 ---
 
-## indexes_relation_size_pretty() [](indexes_relation_size_pretty)
+## hypertable_index_size() [](indexes_relation_size)
```
mfreed, line 3415: Shouldn't the anchor tag be "hypertable_index_size" ?

api.md
```diff
@@ -3406,138 +3312,143 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
...
-Get sizes of indexes on a hypertable.
+Get size of an index on a hypertable.
 
-#### Required Arguments [](indexes_relation_size_pretty-required-arguments)
+#### Required Arguments [](indexes_relation_size-required-arguments)
 
 |Name|Description|
 |---|---|
-| `main_table` | Identifier of hypertable to get indexes size for.|
+| `main_table` | Name of hypertable|
```
mfreed, line 3423: Isn't the required arg "index" not "hypertable" name?

api.md
```diff
@@ -3406,138 +3312,143 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
...
+ volume | integer                  |           |          | 
+Indexes:
+    "second_index" btree ("time")
+    "test_table_time_idx" btree ("time" DESC)
+    "third_index" btree ("time")
+
+select pg_size_pretty(hypertable_index_size('second_index'));
+ pg_size_pretty 
+----------------
+ 160 kB
```
mfreed, line 3448: Maybe should example first without pg_size_pretty?

(Also, for consistency, I think we are capitalizing "SELECT")

api.md
```diff
@@ -3613,7 +3524,7 @@ psql [your connect flags] -d your_timescale_db < dump_meta_data.sql > dumpfile.t
 and then inspect `dump_file.txt` before sending it together with a bug report or support question.
 
 [Slack]: https://slack-login.timescale.com
-[chunk relation size]: #chunk_relation_size
+[chunk detailed size]: #chunk_detialed_size
```
mfreed, line 3527:
```suggestion
[chunk detailed size]: #chunk_detailed_size
```

mfreed (changes requested):

api.md
gayyappan: yes, this is Apache. Removed.

api.md
gayyappan: @mfreed The size functions and some of the new utility functions make calls to tsl code in case the table is a distributed table. But this is not the case for regular hypertables.
So it would work for regular hypertable with Apache. But the function could call tsl code in some circumstances. So is this Apache or Community?

api.md
gayyappan: Modified. I wasn't aware that we don't recommend this usage.

api.md
svenklemm: you can make the function apache and still have the call to tsl code

svenklemm: i mean have the call be conditional and run tsl code only in community edition we have a couple of places that do that already like telemetry that gets additional information

gayyappan: This calls tsl code for distributed hypertables. Since we cannot have distributed hypertables in apache, is this good enough?
```
        CASE WHEN is_distributed THEN
            RETURN QUERY SELECT * FROM _timescaledb_internal.hypertable_remote_size(schema_name, table_name);
        ELSE
            RETURN QUERY SELECT *, NULL::name FROM _timescaledb_internal.hypertable_local_size(schema_name, table_name);
        END CASE;
```
The remote_size function invokes the tsl code.

svenklemm: yes i think this is fine

svenklemm (approved):

api.md
mfreed: I believe this should then be updated to remove the :community_function: label?

api.md
```diff
@@ -3406,138 +3298,158 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
...
+ table_bytes | index_bytes | toast_bytes | total_bytes |  node_name
+-------------+-------------+-------------+-------------+-------------
+       16384 |       32768 |           0 |       49152 | data_node_1
+        8192 |       16384 |           0 |       24576 | data_node_2
+
 ```
 ---
 
-## hypertable_relation_size_pretty() [](hypertable_relation_size_pretty)
+## chunks_detailed_size()  :community_function: [](chunks_detailed_size)
```
mfreed, line 3340: Remove :community_function:

api.md
```diff
@@ -3406,138 +3298,158 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
...
+     chunk_schema      |      chunk_name       | table_bytes | index_bytes | toast_bytes | total_bytes |       node_name
+-----------------------+-----------------------+-------------+-------------+-------------+-------------+-----------------------
+ _timescaledb_internal | _dist_hyper_1_1_chunk |        8192 |       32768 |           0 |       40960 | db_node1
+ _timescaledb_internal | _dist_hyper_1_2_chunk |        8192 |       32768 |           0 |       40960 | db_node2
+ _timescaledb_internal | _dist_hyper_1_3_chunk |        8192 |       32768 |           0 |       40960 | db_node3
+```
 ---
 
-## indexes_relation_size() [](indexes_relation_size)
+## hypertable_size() :community_function: [](hypertable_size)
```
mfreed, line 3383: Remove :community_function:

api.md
```diff
@@ -3406,138 +3298,158 @@ SELECT pg_size_pretty(after_compression_total_bytes) as total
...
 
+ hypertable_size
+-----------------
+           73728
 ```
-
 ---
 
-## indexes_relation_size_pretty() [](indexes_relation_size_pretty)
+## hypertable_index_size() :community_function: [](hypertable_index_size)
```
mfreed, line 3410: Remove :community_function:

api.md
```diff
@@ -3613,7 +3525,7 @@ psql [your connect flags] -d your_timescale_db < dump_meta_data.sql > dumpfile.t
 and then inspect `dump_file.txt` before sending it together with a bug report or support question.
 
 [Slack]: https://slack-login.timescale.com
-[chunk relation size]: #chunk_relation_size
+[chunk detailed size]: #chunk_detailed_size
```
mfreed, line 3528: I don't /think/ this is used anywhere?  But regardless not incorrect.

mfreed (approved):
A few tweaks with removing :community_function: label, but otherwise LGTM.

api.md
gayyappan: removed community_function label.

PR #506, opened 2020-10-01
Add timescaledb.license API section

pmwkaa:
This PR brings back `timescaledb.license` guc description and updates it according to the recent changes

### DIFF
--- a/api.md
+++ b/api.md
@@ -61,6 +61,7 @@
 > - [timescaledb_information.hypertables](#timescaledb_information-hypertables)
 > - [timescaledb_information.job_stats](#timescaledb_information-job_stats)
 > - [timescaledb_information.reorder_policies](#timescaledb_information-reorder_policies)
+> - [timescaledb.license](#timescaledb_license)
 > - [timescaledb_pre_restore](#timescaledb_pre_restore)
 > - [timescaledb_post_restore](#timescaledb_post_restore)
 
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -3518,6 +3519,28 @@ SELECT timescaledb_post_restore();
 ```
 ---
 
+## timescaledb.license [](timescaledb_license)
+
+View or set currently used TimescaleDB license.
+
+It is possible to downgrade currently used license from `timescale` to `apache` by setting
+the `timescaledb.license` GUC in the server configuration file or using the
+server command line.
```
k-rus, line 3528: Can you give more details or even give example in sample session?

k-rus (approved):
LGTM but might be good to elaborate how to change the value.

api.md
erimatnor, line 3528: I am not sure we should focus on downgrades here. I suggest adding a bit more nuance:

```suggestion
It is possible to limit access to features by license by changing the `timescaledb.license` settings parameter in the server configuration file or on the server command line. For instance, by setting `timescaledb.license`  to `apache`, it is only possible to use TimescaleDB features that are implemented under the Apache 2 license. The default value is `timescale`, however, which allows access to all features.
```

mfreed (approved):
I like the suggestions, but otherwise LGTM.

PR #509, opened 2020-10-02
Clarify single hypertable in continuous aggregate

k-rus:
There is no explicit note in continuous aggregate that it can select only from one hypertable. This commit adds notes and clarifiesthis.

### DIFF
--- a/api.md
+++ b/api.md
@@ -1610,6 +1610,11 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
          [ optional grouping exprs>]
 [HAVING ...]
 ```
+Note that continuous aggregates have some limitations of what types of
+queries they can support, described in more length below.  For example,
+the `FROM` clause must provide only one hypertable, i.e., no joins, views or 
+subqueries ares upported, the `GROUP BY` clause must include a time bucket on 
+the hypertable's time column, and all aggregates must be parallelizable.
 
 #### Parameters
 |Name|Description|
@@ -1644,10 +1649,14 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
 - The view will be automatically refreshed (as outlined under
   [`refresh_continuous_aggregate`](#refresh_continuous_aggregate))
   unless `WITH NO DATA` is given (`WITH DATA` is the default).
-- `SELECT` query should be of the form specified in the syntax above.
+- The `SELECT` query should be of the form specified in the syntax above, which is discussed in
+  the following items.
+- Only a single hypertable can be specified in the `FROM` clause of the 
+  `SELECT` query. This means that including more hypertables, joins, tables, views, subqueries
+  is not supported.
 - The hypertable used in the `SELECT` may not have [row-level-security
   policies][postgres-rls] enabled.
--  `GROUP BY` clause must include a time_bucket expression. The
+-  The `GROUP BY` clause must include a time_bucket expression. The
    [`time_bucket`](#time_bucket) expression must use the time
    dimension column of the hypertable.
 - [`time_bucket_gapfill`](#time_bucket_gapfill) is not allowed in continuous


mkindahl (approved):

api.md
```diff
@@ -1669,14 +1669,15 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
          [ optional grouping exprs>]
 [HAVING ...]
 ```
+Note that only hypertable and only one is allowed in `FROM` clause.
```
erimatnor, line 1672: Current sentence reads a bit strange. I suggest:
```suggestion
Note that only a single hypertable is supported in the `FROM` clause.
```

api.md
```diff
@@ -1669,14 +1669,15 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
...
 
 #### Parameters
 |Name|Description|
 |---|---|
 | `<view_name>` | Name (optionally schema-qualified) of continuous aggregate view to be created.|
 | `<column_name>`| Optional list of names to be used for columns of the view. If not given, the column names are deduced from the query.|
 | `WITH` clause | This clause specifies [options](#continuous_aggregate-create_view-with) for the continuous aggregate view.|
-| `<select_query>`| A `SELECT` query that uses the specified syntax. |
+| `<select_query>`| A `SELECT` query that uses the specified syntax, which allows only a hypertable in `FROM` clause. |
```
erimatnor, line 1680:
```suggestion
| `<select_query>`| A `SELECT` query that uses the specified syntax, which allows only a single hypertable in the `FROM` clause. |
```

api.md
```diff
@@ -1704,6 +1705,8 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
   [`refresh_continuous_aggregate`](#refresh_continuous_aggregate))
   unless `WITH NO DATA` is given (`WITH DATA` is the default).
 - `SELECT` query should be of the form specified in the syntax above.
+- Only hypertable and only one can be specified in `FROM` clause of `SELECT`
+  query.
```
erimatnor, line 1709:
```suggestion
- Only a single hypertable  can be specified in the `FROM` clause of a `SELECT`
  query.
```

api.md
k-rus: I agree it is a little bit strange, but the suggestion might not emphasise `hypertable` strong enough. But I will go for it.

erimatnor (approved):

mfreed (commented):
Hrm, I'm not sure I agree with these changes.

It seems like you are trying to address the limitation that you can't perform a JOIN as part of the continuous agg?   I think that's a totally great thing to include in the limitations section, but these changes seem to mention them 3 times, which seem excessive (as you don't want to mention every limitation multiple times, and I'm not sure what's special about the "one hypertable" rule)?

Also, the proposed changes say "a single _hypertable_".  But presumably you also mean a JOIN between a hypertable and a regular table?

(I proposed something in the language)

k-rus:
> Also, the proposed changes say "a single _hypertable_". But presumably you also mean a JOIN between a hypertable and a regular table?

I actually formulated more explicit regarding this, however @erimatnor thinks it is unnecessary: https://github.com/timescale/docs.timescale.com-content/pull/509#discussion_r498751871



> It seems like you are trying to address the limitation that you can't perform a JOIN as part of the continuous agg?

It's about both: only hypertable, no, e.g., subqeruy, and only one, i.e., no joins.

Yes, 3 times might be too much. However, I don't think it is good enough just to mention in limitation. Many users miss this limitation.

mfreed (changes requested):
> > Also, the proposed changes say "a single _hypertable_". But presumably you also mean a JOIN between a hypertable and a regular table?
> 
> I actually formulated more explicit regarding this, however @erimatnor thinks it is unnecessary: [#509 (comment)](https://github.com/timescale/docs.timescale.com-content/pull/509#discussion_r498751871)

I think Erik's reply is mostly that the original text was hard to parse.

> 
> > It seems like you are trying to address the limitation that you can't perform a JOIN as part of the continuous agg?
> 
> It's about both: only hypertable, e.g., no subquery, and only one, i.e., no joins.
> 
> Yes, 3 times might be too much. However, I don't think it is good enough just to mention in limitation. Many users miss this limitation.

How about we keep at top and button; a "limitation" does not belong in the definition of the "required argument".

And let's just make it explicit.  I'll make suggestions inline.

api.md
```diff
@@ -1669,14 +1669,15 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
...
+Note that a single hypertable is supported in the `FROM` clause.
 
 #### Parameters
 |Name|Description|
 |---|---|
 | `<view_name>` | Name (optionally schema-qualified) of continuous aggregate view to be created.|
 | `<column_name>`| Optional list of names to be used for columns of the view. If not given, the column names are deduced from the query.|
 | `WITH` clause | This clause specifies [options](#continuous_aggregate-create_view-with) for the continuous aggregate view.|
-| `<select_query>`| A `SELECT` query that uses the specified syntax. |
+| `<select_query>`| A `SELECT` query that uses the specified syntax, which allows only a singe hypertable in the `FROM` clause. |
```
mfreed, line 1680: Let's not make this change.  Two reminders is enough, and very long parameter args just are hard to read.  Plus, the "suggestion" as written is actually confusing.

api.md
```diff
@@ -1669,14 +1669,15 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
          [ optional grouping exprs>]
 [HAVING ...]
 ```
+Note that a single hypertable is supported in the `FROM` clause.
```
mfreed, line 1672:
```suggestion
Note that continuous aggregates have some limitations of what types of
queries they can support, described in more length below.  For example,
the `FROM` clause must provide a single hypertable name (no JOINs or subqueries),
the `GROUP BY` must include a time bucket on the hypertable's time column, and
all aggregates must be parallelizable.
```

mfreed (changes requested):

api.md
k-rus: Removed

api.md
k-rus: Used the above text with some modifications.

api.md
```diff
@@ -1703,10 +1704,12 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
 - The view will be automatically refreshed (as outlined under
   [`refresh_continuous_aggregate`](#refresh_continuous_aggregate))
   unless `WITH NO DATA` is given (`WITH DATA` is the default).
-- `SELECT` query should be of the form specified in the syntax above.
+- The `SELECT` query should be of the form specified in the syntax above.
+- Only a single hypertable can be specified in the `FROM` clause of the 
+  `SELECT` query.
```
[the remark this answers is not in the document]
k-rus: Made longer clarifications

api.md
```diff
@@ -1610,6 +1610,11 @@ GROUP BY <time_bucket( <const_value>, <partition_col_of_hypertable> ),
          [ optional grouping exprs>]
 [HAVING ...]
 ```
+Note that continuous aggregates have some limitations of what types of
+queries they can support, described in more length below.  For example,
+the `FROM` clause must provide only one hypertable, i.e., no joins, views or 
+subqueries ares upported, the `GROUP BY` clause must include a time bucket on
```
mfreed, line 1616:
```suggestion
subqueries are supported, the `GROUP BY` clause must include a time bucket on 
```

mfreed (approved):

PR #515, opened 2020-10-04
Update compression docs to reflect 2.0 changes

WireBaron:
This change updates the api doc to include the new `if_not_exists`
parameter for add/remove_compression policy. It also renames the
`time_interval` parameter of add_compression_policy to
`compress_after`.

It also updates the compression document under the using-timescale
section of the documentation to use the 2.0 apis and views.

### DIFF
--- a/api.md
+++ b/api.md
@@ -1457,19 +1457,29 @@ ALTER TABLE metrics SET (timescaledb.compress, timescaledb.compress_orderby = 't
 
 ## add_compression_policy() :community_function: [](add_compression_policy)
 Allows you to set a policy by which the system will compress a chunk
-automatically in the background after it reaches a given age.
+automatically in the background after it reaches a given age. 
+
+Note that compression policies can only be created on hypertables that already
+have compression enabled, e.g., via the [`ALTER TABLE`](/api#compression_alter-table) command
+to set `timescaledb.compress` and other configuration parameters.
 
 #### Required Arguments [](add_compression_policy-required-arguments)
 
 |Name|Description|
 |---|---|
 | `hypertable` | (REGCLASS) Name of the hypertable|
-| `compress_after` | (INTERVAL or integer) The age after which the policy job will compress chunks|
+| `compress_after` | (INTERVAL or INTEGER) The age after which the policy job will compress chunks|
 
 The `compress_after` parameter should be specified differently depending on the type of the time column of the hypertable:
 - For hypertables with TIMESTAMP, TIMESTAMPTZ, and DATE time columns: the time interval should be an INTERVAL type
 - For hypertables with integer-based timestamps: the time interval should be an integer type (this requires
 the [integer_now_func](#set_integer_now_func) to be set).
+
+#### Optional Arguments [](add_compression_policy-optional-arguments)
+
+|Name|Description|
+|---|---|
+| `if_not_exists` | (BOOLEAN) Setting to true will cause the command to fail with a warning instead of an error if a compression policy already exists on the hypertable. Defaults to false.|
 
 #### Sample Usage [](add_compression_policy-sample-usage)
 Add a policy to compress chunks older than 60 days on the 'cpu' hypertable.
@@ -1492,6 +1502,12 @@ If you need to remove the compression policy. To re-start policy basd compressio
 |Name|Description|
 |---|---|
 | `hypertable` | (REGCLASS) Name of the hypertable the policy should be removed from.|
+
+#### Optional Arguments [](remove_compression_policy-optional-arguments)
+
+|Name|Description|
+|---|---|
+| `if_exists` | (BOOLEAN) Setting to true will cause the command to fail with a notice instead of an error if a compression policy does not exist on the hypertable. Defaults to false.|
 
 #### Sample Usage [](remove_compression_policy-sample-usage)
 Remove the compression policy from the 'cpu' table:
--- a/using-timescaledb/compression.md
+++ b/using-timescaledb/compression.md
@@ -271,15 +271,16 @@ You can see the results of the compression of that given chunk by running the fo
 
 ``` sql
 SELECT *
-FROM timescaledb_information.compressed_chunk_stats;
+FROM chunk_compression_stats('conditions');
 ```
 
-| hypertable_name | chunk_name | compression_status | uncompressed_heap_bytes | uncompressed_index_bytes | uncompressed_toast_bytes | uncompressed_total_bytes | compressed_heap_bytes | compressed_index_bytes | compressed_toast_bytes | compressed_total_bytes |
-| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |  --- |
-| my_hypertable | _timescaledb_internal._hyper_1_1_chunk | Compressed | 8192 bytes | 16 kB | 8192 bytes |   | 32 kB  | 8192 bytes   | 16 kB  | 8192 bytes | 32 kB |
-| my_hypertable | _timescaledb_internal._hyper_1_20_chunk | Uncompressed | | | | | | | | |
+| chunk_schema | chunk_name | compression_status | before_compression_table_bytes | before_compression_index_bytes | before_compression_toast_bytes | before_compression_total_bytes | after_compression_table_bytes | after_compression_index_bytes | after_compression_toast_bytes | after_compression_total_bytes | node_name |
+| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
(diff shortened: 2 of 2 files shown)


api.md
```diff
@@ -1453,20 +1453,28 @@ ALTER TABLE metrics SET (timescaledb.compress, timescaledb.compress_orderby = 't
 
 ## add_compression_policy() :community_function: [](add_compression_policy)
 Allows you to set a policy by which the system will compress a chunk
-automatically in the background after it reaches a given age.
+automatically in the background after it reaches a given age. Note that
+a hypertable must have the `timescaledb.compress` storage parameter set
+in order to be compressed.
```
mfreed, line 1458: This instruction feels confusing, at it refers to a specific "storage parameter", which (a) isn't in _this_ API call, and (b) isn't in the `create_hypertable` call.

I guess I understand that when users use `ALTER TABLE` to set a compression policy, this "storage parameter" gets set, but I wonder if this is too "in the weeds" and instead we should uplevel it.

```suggestion
automatically in the background after it reaches a given age. 

Note that compression policies can only be created on hypertables that already
have compression enabled, e.g., via the [`ALTER TABLE`](/api#compression_alter-table) command
to set `timescaledb.compress` and other configuration parameters.
```

api.md
```diff
@@ -1453,20 +1453,28 @@ ALTER TABLE metrics SET (timescaledb.compress, timescaledb.compress_orderby = 't
...
 #### Required Arguments [](add_compression_policy-required-arguments)
 
 |Name|Description|
 |---|---|
 | `table_name` | (REGCLASS) Name of the table that the policy will act on.|
-| `time_interval` | (INTERVAL or integer) The age after which the policy job will compress chunks.|
+| `compress_after` | (INTERVAL or integer) The age after which the policy job will compress chunks.|
 
-The `time_interval` parameter should be specifified differently depending on the type of the time column of the hypertable:
+The `compress_after` parameter should be specifified differently depending on the type of the time column of the hypertable:
```
mfreed, line 1467:
```suggestion
The `compress_after` parameter should be specified differently depending on the type of the time column of the hypertable:
```

using-timescaledb/compression.md
```diff
@@ -271,15 +271,16 @@ You can see the results of the compression of that given chunk by running the fo
...
-| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |  --- |
-| my_hypertable | _timescaledb_internal._hyper_1_1_chunk | Compressed | 8192 bytes | 16 kB | 8192 bytes |   | 32 kB  | 8192 bytes   | 16 kB  | 8192 bytes | 32 kB |
-| my_hypertable | _timescaledb_internal._hyper_1_20_chunk | Uncompressed | | | | | | | | |
+| chunk_schema | chunk_name | compression_status | before_compression_table_bytes | before_compression_index_bytes | before_compression_toast_bytes | before_compression_total_bytes | after_compression_table_bytes | after_compression_index_bytes | after_compression_toast_bytes | after_compression_total_bytes | node_name |
+| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
+| _timescaledb_internal | _hyper_1_1_chunk | Compressed | 8192 bytes | 16 kB | 8192 bytes |  32 kB  | 8192 bytes   | 16 kB  | 8192 bytes | 32 kB | |
+| _timescaledb_internal | _hyper_1_20_chunk | Uncompressed | | | | | | | | | |
```
mfreed, line 280: Given the width, perhaps these should be shown in `\x` mode?

using-timescaledb/compression.md
```diff
@@ -305,15 +306,16 @@ to re-compress the chunks that we are currently working on (not the desired resu
 To accomplish this we first find the job_id of the policy using:
 
 ```sql
-SELECT job_id, table_name
-FROM _timescaledb_config.bgw_policy_compress_chunks p
-INNER JOIN _timescaledb_catalog.hypertable h ON (h.id = p.hypertable_id);
+SELECT s.job_id
+FROM timescaledb_information.jobs j
+INNER JOIN timescaledb_information.job_stats s ON j.job_id = s.job_id
+WHERE j.proc_name = 'policy_compression' AND s.hypertable_name = <target table>;
```
mfreed, line 312:
```suggestion
SELECT s.job_id
  FROM timescaledb_information.jobs j
  INNER JOIN timescaledb_information.job_stats s ON j.job_id = s.job_id
  WHERE j.proc_name = 'policy_compression' AND s.hypertable_name = <target table>;
```

using-timescaledb/compression.md
```diff
@@ -339,10 +341,15 @@ Once your backfill and update operations are complete we can simply re-enable
 our compression policy job:
 
 ``` sql
-SELECT alter_job_schedule(<job_id>, next_start=>now());
+SELECT alter_job(<job_id>, scheduled => true);
+```
+
+This job will re-compress any chunks that were decompressed during your backfilling
+operation the next time it runs. To have it run immediately you can do the following:
```
mfreed, line 348:
```suggestion
operation the next time it runs. To have it run immediately, you can expressly execute
the command via [`run_job`](/api#run_job):
```

k-rus:
@WireBaron `if_not_exists` is not a new argument. It is present in `add_compress_chunks_policy` in 1.7. It seems that it was just not documented. Can you update the commit message?

using-timescaledb/compression.md
k-rus: @mfreed Your proposal is not consistent with other examples in this and other docs. The top select clauses are usually indented in the same way as `SELECT`, however the join should have an additional indentation. I believe it should be:
```suggestion
SELECT s.job_id
FROM timescaledb_information.jobs j
  INNER JOIN timescaledb_information.job_stats s ON j.job_id = s.job_id
WHERE j.proc_name = 'policy_compression' AND s.hypertable_name = <target table>;
```
or
```suggestion
SELECT s.job_id
FROM timescaledb_information.jobs j
     INNER JOIN timescaledb_information.job_stats s ON j.job_id = s.job_id
WHERE j.proc_name = 'policy_compression' AND s.hypertable_name = <target table>;
```

api.md
```diff
@@ -1453,20 +1453,28 @@ ALTER TABLE metrics SET (timescaledb.compress, timescaledb.compress_orderby = 't
...
+a hypertable must have the `timescaledb.compress` storage parameter set
+in order to be compressed.
 
 #### Required Arguments [](add_compression_policy-required-arguments)
 
 |Name|Description|
 |---|---|
 | `table_name` | (REGCLASS) Name of the table that the policy will act on.|
-| `time_interval` | (INTERVAL or integer) The age after which the policy job will compress chunks.|
+| `compress_after` | (INTERVAL or integer) The age after which the policy job will compress chunks.|
```
svenklemm, line 1465:
```suggestion
| `compress_after` | (INTERVAL or INTEGER) The age after which the policy job will compress chunks.|
```

svenklemm (approved):

using-timescaledb/compression.md
WireBaron: I think in this instance, printing in `\x` would be highlighting the wrong thing.  Really we're just demonstrating the command, the actual numbers in the result aren't interesting, so I feel like minimizing the vertical space and allowing the user to scroll is a better fit.

using-timescaledb/compression.md
WireBaron: Doing a quick scan of the docs, it looks like there isn't a standard usage here.  Most common appears to be having nothing indented, though there are some examples where only joins are indented and a few others where everything is indented.  Personally I find Ruslan's suggestion a bit more readable than Mike's here (though I find the larger indentation a bit glaring), so I'm going with that for the moment.  I expect we'll have to revisit this if we decide to standardize our docs sql.

mfreed:
I'm not actually seeing the changes that were "fixed"?

using-timescaledb/compression.md
```diff
@@ -339,10 +341,16 @@ Once your backfill and update operations are complete we can simply re-enable
 our compression policy job:
 
 ``` sql
-SELECT alter_job_schedule(<job_id>, next_start=>now());
+SELECT alter_job(<job_id>, scheduled => true);
+```
+
+This job will re-compress any chunks that were decompressed during your backfilling
+operation the next time it runs. To have it run immediatelyimmediately, you can expressly execute
```
mfreed, line 348:
```suggestion
operation the next time it runs. To have it run immediately, you can expressly execute
```

mfreed (approved):
One tiny nit, but after that's fixed, looks good to merge.

using-timescaledb/compression.md
WireBaron: fixed

using-timescaledb/compression.md
WireBaron: Looking at that example doc, it does look like there's a effort to make them more readable.  However, there still doesn't seem to be much consistency between examples.

PR #516, opened 2020-10-05
Update parameter names for several functions

mkindahl:
This updates the parameter names for functions changed in
timescale/timescaledb#2472.

### DIFF
--- a/api.md
+++ b/api.md
@@ -125,7 +125,7 @@ queries.
 
 |Name|Description|
 |---|---|
-| `main_table` | Identifier of hypertable to add the dimension to.|
+| `hypertable` | Hypertable to add the dimension to.|
 | `column_name` | Name of the column to partition by.|
 
 #### Optional Arguments [](add_dimension-optional-arguments)
@@ -403,7 +403,7 @@ still work on the resulting hypertable.
 
 |Name|Description|
 |---|---|
-| `main_table` | Identifier of table to convert to hypertable |
+| `relation` | Identifier of table to convert to hypertable. |
 | `time_column_name` | Name of the column containing time values as well as the primary column to partition by. |
 
 #### Optional Arguments [](create_hypertable-optional-arguments)
@@ -614,7 +614,7 @@ when creating distributed hypertables.
 
 |Name|Description|
 |---|---|
-| `main_table` | Identifier of table to convert to hypertable |
+| `relation` | Identifier of table to convert to hypertable. |
 | `time_column_name` | Name of the column containing time values as well as the primary column to partition by. |
 
 #### Optional Arguments [](create_distributed_hypertable-optional-arguments)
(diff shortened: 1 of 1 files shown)


svenklemm (approved):

api.md
```diff
@@ -403,7 +403,7 @@ still work on the resulting hypertable.
 
 |Name|Description|
 |---|---|
-| `main_table` | Identifier of table to convert to hypertable |
+| `relation` | Identifier of table to convert to hypertable |
```
erimatnor, line 406: Assuming that punctuation is our standard:

```suggestion
| `relation` | Identifier of table to convert to hypertable. |
```

api.md
```diff
@@ -614,7 +614,7 @@ when creating distributed hypertables.
...
```
erimatnor, line 617:
```suggestion
| `relation` | Identifier of table to convert to hypertable. |
```

api.md
```diff
@@ -958,7 +958,7 @@ specified one.
 
 |Name|Description|
 |---|---|
-| `hypertable_or_cagg` | Hypertable or continuous aggregate from which to drop chunks.
+| `relation` | Hypertable or continuous aggregate from which to drop chunks.
```
erimatnor, line 961:
```suggestion
| `relation` | Hypertable or continuous aggregate from which to drop chunks. |
```

api.md
```diff
@@ -3265,7 +3265,7 @@ All sizes are in bytes.
 
 |Name|Description|
 |---|---|
-| `main_table` | (REGCLASS) Name of the hypertable |
+| `hypertable` | (REGCLASS) Hypertable to show stats for |
```
erimatnor, line 3268:
```suggestion
| `hypertable` | Hypertable to show stats for. |
```

api.md
```diff
@@ -3394,7 +3394,7 @@ information as a separate row per node.
 
 |Name|Description|
 |---|---|
-| `main_table` | (REGCLASS) Name of the hypertable |
+| `hypertable` | (REGCLASS) Hypertable to show detailed size of |
```
erimatnor, line 3397:
```suggestion
| `hypertable` | Hypertable to show detailed size of. |
```

api.md
```diff
@@ -3473,7 +3473,7 @@ output of `hypertable_detailed_size` function.
 
 |Name|Description|
 |---|---|
-| `main_table` | (REGCLASS) Name of the hypertable |
+| `hypertable` | (REGCLASS) Hypertable to show size of |
```
erimatnor, line 3476:
```suggestion
| `hypertable` | Hypertable to show size of. |
```

erimatnor (approved):
Added some nits for consistency. I looks like some API function parameters document the type while many don't. Do we want to make this consistent too?

api.md
mfreed: I believe the "preferred" format is to include type argument's TYPE (we aren't great here, but I guess would like to trend towards better).

api.md
mfreed: Is there a reason you don't just use "table" here?  I mean, I guess `relation` is correct from the format database semantics perspective, but I think will be confusing for the average user....

mkindahl: > Is there a reason you don't just use "table" here? I mean, I guess `relation` is correct from the format database semantics perspective, but I think will be confusing for the average user....

We had a discussion in the team and decided to use `hypertable` when it was a hypertable, and `relation` for other cases (e.g., a Postgres table, hypertable or cagg, etc.)

PR #517, opened 2020-10-05
Update introduction to continuous aggregates

erimatnor:
This PR contains tweaks and changes to the section on continuous aggregates.

### DIFF
--- a/api.md
+++ b/api.md
@@ -3085,7 +3085,7 @@ proc_schema       | _timescaledb_internal
 proc_name         | policy_refresh_continuous_aggregate
 owner             | postgres
 scheduled         | t
-config            | {"end_offset": "20 days", "start_offset": "10 
+config            | {"start_offset": "20 days", "end_offset": "10 
 days", "mat_hypertable_id": 2}
 next_start        | 2020-10-02 12:38:07.014042-04
 hypertable_schema | _timescaledb_internal
--- a/api.md
+++ b/api.md
@@ -3085,7 +3085,7 @@ proc_schema       | _timescaledb_internal
 proc_name         | policy_refresh_continuous_aggregate
 owner             | postgres
 scheduled         | t
-config            | {"end_interval": "20 days", "start_interval": "10 
+config            | {"end_offset": "20 days", "start_offset": "10 
 days", "mat_hypertable_id": 2}
 next_start        | 2020-10-02 12:38:07.014042-04
 hypertable_schema | _timescaledb_internal
--- a/using-timescaledb/continuous-aggregates.md
+++ b/using-timescaledb/continuous-aggregates.md
@@ -2,29 +2,27 @@
 
 Aggregate queries which touch large swathes of time-series data can
 take a long time to compute because the system needs to scan large
-amounts of data on every query execution. TimescaleDB continuous
-aggregates automatically calculate the results of a query in the
-background and materialize the results. Queries to the continuous
-aggregate view are then significantly faster as they touch less raw
-data in the hypertable and instead mostly use the pre-computed
-aggregates to build the view.
+amounts of data on every query execution. To make such queries faster,
+continuous aggregates allows pre-computing (or materializing) the
+aggregates, while also providing means to continuously, and without
+much overhead, keep them up-to-date as the underlying source data
+changes.
 
-Continuous aggregates are somewhat similar to PostgreSQL [materialized
-views][postgres-materialized-views], but unlike a materialized view,
-continuous aggregates do not need to be refreshed manually; the view
-will be refreshed automatically in the background as new data is
-added, or old data is modified. Additionally, it does not need to
-re-calculate all of the data on every refresh. Only new and/or
-invalidated data will be calculated.  Since this re-aggregation is
-automatic, it doesn’t add any maintenance burden to your database.
-
+Continuous aggregates are somewhat similar to PostgreSQL's
+[materialized views][postgres-materialized-views], but, unlike a
+materialized view, a continuous aggregate can be updated continuously
+and incrementally, either via manual refreshing or a policy that runs
+in the background. A refresh can cover the entire aggregate or just a
+specific time range. In either case, the refresh only recomputes the
+aggregate buckets that have changed since the last refresh.
+ 
 ### An introductory example [](quick-start)
 
 As a quick introductory example, let's create a hypertable
 `conditions` containing temperature data for devices and a continuous
 aggregate to compute the daily average, minimum, and maximum
 temperature. Start off by creating the hypertable and populate it with
-some random data:
+some data:
 
 ```sql
 CREATE TABLE conditions (
@@ -81,7 +79,7 @@ SELECT bucket, avg
 ORDER BY bucket;
 ```
 
-### A detailed look at continuous aggregate views [](detailed-look)
+### A detailed look at continuous aggregates [](detailed-look)
 
 As shown above, creating a refreshing [continuous
 aggregate][api-continuous-aggs] is a two-step process. First, one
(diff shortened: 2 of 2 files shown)


using-timescaledb/continuous-aggregates.md
```diff
@@ -149,88 +146,85 @@ continuous aggregate, so this will refresh only the buckets that are
...
-you can set `start_interval` to one month (or less) and thereby not
-refresh data older than one month, which includes data that is
-removed.
+If you instead want to keep data in the continuous aggregate even if
+the source data is removed from the underlying hypertable, you also
+need to set the `start_offset` in way that is compatible with the
+[data retention policy][sec-data-retention] on the source
+hypertable. For example, if you have a retention policy that removes
+data older than one month, you need to set `start_offset` to one month
+(or less) and thereby not refresh into to region of dropped data.
```
mfreed, line 222:
```suggestion
(or less) and thereby not refresh the region of dropped data.
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -353,47 +312,26 @@ continuous aggregate, write:
...
-ALTER MATERIALIZED VIEW conditions_summary_hourly SET (
-    timescaledb.materialized_only = true
-);
-```
+controls how frequently materialization jobs will be run. Setting a
+shorter interval will refresh more frequently but at the same time
+consume more background worker resources.
 
-**Using `timescaledb.information` Views:**
+#### Using `timescaledb.information` Views####
```
mfreed, line 334:
```suggestion
#### Using `timescaledb.information` Views
```

mfreed (approved):

api.md
```diff
@@ -3068,7 +3068,7 @@ proc_schema       | _timescaledb_internal
 proc_name         | policy_refresh_continuous_aggregate
 owner             | postgres
 scheduled         | t
-config            | {"end_interval": "20 days", "start_interval": "10 
+config            | {"end_offset": "20 days", "start_offset": "10
```
erimatnor, line 3071: Hmm, this config doesn't make sense. I guess end_offset should be smaller than start_offset?

using-timescaledb/continuous-aggregates.md
```diff
@@ -2,29 +2,26 @@
...
-views][postgres-materialized-views], but unlike a materialized view,
-continuous aggregates do not need to be refreshed manually; the view
-will be refreshed automatically in the background as new data is
-added, or old data is modified. Additionally, it does not need to
-re-calculate all of the data on every refresh. Only new and/or
-invalidated data will be calculated.  Since this re-aggregation is
-automatic, it doesn’t add any maintenance burden to your database.
-
+amounts of data on every query execution. To make such queries faster,
+continuous aggregates allows pre-computing (or materializing) the
```
mkindahl, line 6: People that are accustomed to databases are probably familiar with what materialization is, so not sure it is good to use "precomputing" instead of "materialization".

using-timescaledb/continuous-aggregates.md
```diff
@@ -2,29 +2,26 @@
...
+aggregates, while also providing means to continuously, and with low
+overhead, keep them up-to-date as the underlying source data changes.
+
+Continuous aggregates are somewhat similar to PostgreSQL's
+[materialized views][postgres-materialized-views], but, unlike a
+materialized view, a continuous aggregate can be continuously and
+incrementally refreshed, either manually or via a policy that runs in
+the background. A refresh can cover the entire aggregate or just a
+specific time range. In either case, the refresh only recomputes the
+aggregate buckets that have changed since the last refresh.
```
mkindahl, line 16: That first sentence is really long and many nested phrases with commas between making it harder to read. I think you should split it up as it was in the original (you can replace the `;` with a `.`). Something like this:
```suggestion
Continuous aggregates are somewhat similar to PostgreSQL's
[materialized views][postgres-materialized-views] but, unlike a
materialized view, they do not need to be manually refreshed. A continuous aggregate can manually refreshed or refreshed incrementally using a policy that runs in
the background. A refresh can cover the entire aggregate or just a
specific time range. In either case, the refresh only recomputes the
aggregate buckets that have changed since the last refresh.
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -91,8 +88,8 @@ MATERIALIZED VIEW`][postgres-createview] with the
 policy needs to be created to keep it refreshed.
 
 You can create several continuous aggregates for the same
-hypertable. For example, we could create another continuous aggregate
-view that summarizes the hourly data.
+hypertable. For example, you could create another continuous aggregate
+view for daily data.
```
mkindahl, line 92: :+1: I really think we should use more third-person singular.

using-timescaledb/continuous-aggregates.md
```diff
@@ -121,116 +118,111 @@ so that it contains the aggregates computed across the entire
...
-open-ended and will extend to the beginning or end of time,
-respectively.
+`NULL` to `start_offset` or `end_offset` makes the range open-ended
+and will extend to the beginning or end of time,
+respectively. However, it seldom makes sense to use `NULL` for the
+`end_offset`. Instead, it is recommended to set the `end_offset` such
+that at least one time bucket is excluded. For time-series data that
+see mostly in-order writes, the time buckets that still see lots of
+writes will quickly have out-of-date aggregates. Eliding the buckets
+will also provide better performance.
```
mkindahl, line 192: Slight rewrite. I do not think you need that last sentence though
```suggestion
and will extend to the beginning or end of time,
respectively. However, it seldom makes sense to use `NULL` for the
`end_offset`. Instead, it is recommended to set the `end_offset` so
that at least the last time bucket is excluded. Since time-series data usually
see mostly writes to the end of the table, any aggregate on this bucket will quickly become out-of-date. Eliding the last buckets
will also provide better performance.
```

using-timescaledb/continuous-aggregates.md
```diff
@@ -353,47 +310,26 @@ continuous aggregate, write:
...
-ALTER MATERIALIZED VIEW conditions_summary_hourly SET (
-    timescaledb.materialized_only = true
-);
-```
+controls how frequently materialization jobs will be run. Setting a
+shorter interval will refresh more frequently but at the same time
+consume more background worker resources.
 
-**Using `timescaledb.information` Views:**
+#### Using `timescaledb.information` Views
```
mkindahl, line 332: Aren't there more of these in the document?

mkindahl (approved):
Some suggestions, but in general looks good.

using-timescaledb/continuous-aggregates.md
erimatnor: FWIW, I wanted to avoid the use of "last bucket" because it is a bit ambiguous and there is really no choice as to what buckets `end_offset` will exclude. I did some tweaks to the text, however.

using-timescaledb/continuous-aggregates.md
erimatnor: I think so, but I changed this because the content there didn't fit within the current section which is about real-time aggregation. So, it needs to be its own subsection.

using-timescaledb/continuous-aggregates.md
mkindahl: Using "newest buckets" then? The text just mention "buckets", which makes it confusing to what bucket is actually referenced.

using-timescaledb/continuous-aggregates.md
mkindahl: I'm only worried about consistency. Are there other sections that use `**whatever**` style header that need to be changed to `####whatever`?

Issue #492, opened 2020-09-16
[Update] Multi-node and distributed hypertables


mfreed:
**New Intro / Getting Started**

- [x] Introduce single, primary/backup, and multinode configuration in updated intro - https://docs.timescale.com/latest/introduction [@mfreed]
- [x] Create new "quick-start / quick configure" guide to setting up multi-node, new doc for "getting started" section - https://docs.timescale.com/beta-v2.0.0/getting-started/setup-multi-node [@solugebefola][ #525 ]
- [x] Add "quick start" for setting up a new postgres user (simplified for data auth): https://docs.timescale.com/beta-v2.0.0/getting-started/setup-multi-node#data-node-auth [@mkindahl, @solugebefola ][ #525 ]

**New Using Distributed Hypertables**  [@erimatnor]
https://docs.timescale.com/beta-v2.0.0/using-timescaledb/distributed-hypertables

- [x] New intro to section with architectural picture of ANs / DNs and brief description of distributed hypertable [@erimatnor]
- [x] Subsection:  Easy configuration for ANs and DNs, adding data nodes to access node
- [x] Subsection:  Creating distributed hypertable - https://docs.timescale.com/beta-v2.0.0/using-timescaledb/distributed-hypertables#create  [@erimatnor]
- [x] Subsection:  Changing the Number of Data Nodes for a Distributed Hypertable - https://docs.timescale.com/beta-v2.0.0/using-timescaledb/distributed-hypertables#changing-data-nodes  [@erimatnor]
- [x] Subsection: Advanced topic:  Configuring data nodes - https://docs.timescale.com/beta-v2.0.0/getting-started/setup-multi-node#config-data-nodes
- [x] Subsection: Advanced topic: Data authentication:  Edit in light of quick-start, including some instructions about when/if each step needs to be taken https://docs.timescale.com/beta-v2.0.0/getting-started/setup-multi-node#data-node-auth [ @solugebefola  ][ #525 ]
- [x] Subsection: Current Status and Limitations (see below)
- ? Advanced topic: Schedule transaction cleanup (e.g., via custom job): https://github.com/timescale/timescaledb-private/issues/545 [ @pmwkaa ]

**Current Status and Limitations**
- [x] Describe consistency model: https://github.com/timescale/timescaledb-private/issues/514
- [x] Roles and permissions management
- [x] Compression (only `compress_chunk`, no policies) [@gayyappan ]
[...]

PR #540, opened 2020-10-19
Add description of replication factor

mkindahl:
Part of #492

PR instructions

1. Describe your PR right below:

---
2. Add label for the earliest DB version your changes apply to -->>

3. If you are changing page names or in-page anchors, you must update the
`page-index.js` file

4. Please add at least one content reviewer. If you don't add an
additional reviewer with knowledge about the area you are writing about,
one of the codeowners may add one

5. If this is in response to a posted GitHub issue, please add "Fixes <issue #>"
below so that it's referenced (i.e. "Fixes #122")

5. If you are making simple corrections, reviewers may add comments to your
PR without officially requesting changes. This is to avoid additional
request/review cycles for simple changes. Make sure to address these
comments (i.e. with changes) before your PR is ready to merge

6. Feel free to delete these instructions in your PR message after everything
else is filled out
---
Additional notes:

The default reviewers (CODEOWNERS) are added to each PR.  They will
primarily be reviewing on formatting, not content.  You only need ONE of them
to approve in order to merge your PR

### DIFF
--- a/using-timescaledb/distributed-hypertables.md
+++ b/using-timescaledb/distributed-hypertables.md
@@ -1,5 +1,10 @@
 # Distributed Hypertables
 
+>:WARNING: Distributed hypertables are currently in BETA and
+are not yet meant for production use. For more information, please
+[contact us][contact] or join the #multinode-beta channel in our
+[community Slack][slack].
+
 Hypertables handle large amounts of data by breaking it up into
 smaller pieces (chunks), allowing operations to execute
 efficiently. When the amount of data is expected to be beyond what a
@@ -98,6 +103,65 @@ the data node's data is outside the retention window,
 [`drop_chunks`][drop_chunks] can be used to delete the data on the data
 node.
 
+### Native Replication for Distributed Hypertables [](native-replication)
+
+A hypertable can have a *replication factor* that tells how many
+replicas of chunks that should be created. The default replication
+factor is 1, which means that each row of data is written to a single
+chunk. If you set the replication factor of a hypertable to a higher
+value, chunks will be replicated and each row of data will be written
+to that many chunks.
+
+To create a hypertable with a replication factor higher than 1, you
+can use the `replication_factor` parameter to
+[`create_distributed_hypertable`][create_distributed_hypertable]. For
+example, if you want to create a hypertable where each chunk has three
+replicas:
+
+```sql
+SELECT create_distributed_hypertable('conditions', 'time', 'location',
+	replication_factor => 3);
+```
+
+If you already have a hypertable and instead want to change the
+replication factor, you can use the function
+[`set_replication_factor`][set_replication_factor].
+
+```sql
+SELECT set_replication_factor('conditions', 3);
+```
+
+Having a replication factor larger than one is a simple way to ensure
+that you have redundancy of the data and that a crash of a data node
+will not cause you to lose the data. This, however, comes at the cost
+of an increased average commit time and more data to manage.
+
+Read-only query execution is not affected by the redundant data: the
+query will just be split up and send the sub-query to *one* of the
+replicas (not all of them), so the replication factor does not affect
+query execution time in any significant manner.
+
+When executing a commit on the access node, the commit will
+acknowledge the transaction until it has received acknowledgements
+that the transaction is committed on all data nodes. If the
+transaction fails on one more data nodes, or a connection times out,
+the transaction will be aborted.
+
+#### Limitations
+
+The distributed commit protocol is stable and ensures transactional
+semantics even in the presence of data node failures, but if a data
+node crashes permanently, it has to be replaced with a new data node
+that is consistent with the other data nodes.
+
+To be able to recover the system into a operational state, you
+therefore have to ensure that you have a standby of each data node. On
+failure, you then need to promote it to be the primary and add a new
+standby to handle further crashes. The standby can be either a
+[warm][warm-standby] or a [hot standby][hot-standby].
+
+[hot-standby]: https://www.postgresql.org/docs/current/hot-standby.html
+[warm-standby]: https://www.postgresql.org/docs/current/warm-standby.html
 [getting-started-multi-node]: /getting-started/setup-multi-node
 [add_data_node]: /api#add_data_node
 [drop_chunks]: /api#drop_chunks


using-timescaledb/distributed-hypertables.md
```diff
@@ -103,6 +103,90 @@ the data node's data is outside the retention window,
...
+receives an acknowledgement that the transaction is committed on all
+data nodes, or a connection times out, whichever comes first. This
+means that if one of the data nodes crash during a commit (or is
+unresponsive for some other reason) the connection will eventually
+time out and abort the transaction.
+
+Read-only query execution is not affected by the redundant data: the
+query will just be split up and send the sub-query to *one* of the
+replicas (not all of them), so the replication factor does not affect
+query execution time in any significant manner.
```
mfreed, line 173: I think we probably need to cut/shorten this section significantly, to find the right level of detail given the current _preliminary_ support for replication factors (namely, during recovery).

Flagging for additional discussion.

mfreed (changes requested):

using-timescaledb/distributed-hypertables.md
```diff
@@ -103,6 +103,90 @@ the data node's data is outside the retention window,
...
+
+A hypertable can have a *replication factor* that tells how many
+replicas of chunks that should be created. The default replication
+factor is 1, which means that each row of data is written to a single
+chunk. If you set the replication factor of a hypertable to a higher
+value, chunks will be replicated and each row of data will be written
+to that many chunks.
+
+With a replication factor of one, data is written to the data nodes
+using a one-phase commit, which basically means that every transaction
```
erimatnor, line 116: This is incorrect, we never use one-phase commit, unless this is forced in a GUC setting.

Also, replication factor has nothing to do with transactional semantics. Even if replication_factor is 1, a transaction can involve writes to multiple chunks on different data nodes and still need to governed by two-phase commit.

One-phase commit would only be possible if the transaction involves only one data node, which, again, has nothing to do with replication factor.

using-timescaledb/distributed-hypertables.md
```diff
@@ -103,6 +103,90 @@ the data node's data is outside the retention window,
 [`drop_chunks`][drop_chunks] can be used to delete the data on the data
 node.
 
+### Replication Factor for Distributed Hypertables [](replication-factor)
```
erimatnor, line 106: IMO, a section like this should be about native replication in general and not the "replication factor" specifically.

using-timescaledb/distributed-hypertables.md
```diff
@@ -103,6 +103,90 @@ the data node's data is outside the retention window,
...
+to that many chunks.
+
+With a replication factor of one, data is written to the data nodes
+using a one-phase commit, which basically means that every transaction
+on the access node opens a similar transaction on the data
+node. Committing the transaction involves sending a commit message to
+the data node and no additional coordination.
+
+With a higher replication factor, data will be written to the chunks
+using two-phase commit. This protocol is more complicated, but it
```
erimatnor, line 122: Again, the replication_factor has no connection to transactional semantics and commit protocol used.

using-timescaledb/distributed-hypertables.md
```diff
@@ -103,6 +103,90 @@ the data node's data is outside the retention window,
...
+time out and abort the transaction.
+
+Read-only query execution is not affected by the redundant data: the
+query will just be split up and send the sub-query to *one* of the
+replicas (not all of them), so the replication factor does not affect
+query execution time in any significant manner.
+
+#### Limitations
+
+The two-phase commit protocol is stable and handles failures without
```
erimatnor, line 177: There seems to be a fundamental misunderstanding here: two phase commit has no connection to replication factor.

Also, I think this section is misleading. AFAIK, two-phase commit only guarantees transactional semantics (atomicity) but doesn't "handle" failures (at least not in any sense of resolving them). It only ensures that, in case of a failure, a distributed operation fails globally and not only locally on some nodes.

using-timescaledb/distributed-hypertables.md
```diff
@@ -103,6 +103,90 @@ the data node's data is outside the retention window,
...
+  compared to one-phase commit. This is, however, minor and does not
+  significantly affect the commit time.
+- Since there are now more transactions to be committed, the commit
+  time of the two-phase commit (also known as the *global commit*) is
+  the maximum time of the commits on the data nodes (also known as
+  *local commits*), which on average is going to be higher than the
+  average of the commit of each transaction on the data nodes. (The
+  proof of this is left as an excercies for the reader.)
+
+When executing a two-phase commit, the commit will block until it
```
erimatnor, line 163: I think saying it will "block" is misleading; in particular, the client can be notified of success as soon as the coordinator has received "agree to commit" from all nodes.

In any case, I don't think we should go into this amount of detail about the two-phase commit protocol in our documentation. There are better resources online to point to.

erimatnor (changes requested):
I agree with Mike's assessment that his is too much in-depth. And, the section also seems to incorrectly tie two-phase commit to the replication factor (see inline comments).

I think there are useful bits here that we can use for a more general section on native replication. A separate section should deal with the consistency model we use; (transactional guarantees, two-phase commit, etc.)

using-timescaledb/distributed-hypertables.md
mkindahl: Right, I was thinking a single data node, but that is not what it is.

using-timescaledb/distributed-hypertables.md
mkindahl: Changed header.

using-timescaledb/distributed-hypertables.md
mkindahl: Removed paragraphs.

using-timescaledb/distributed-hypertables.md
mkindahl: I think it is useful to give a brief explanation what causes the delay and also leads into the limitations below (that the data node needs to be replaced). The intention is not to go into details, but give a superficial understanding. Reworded.

using-timescaledb/distributed-hypertables.md
mkindahl: I was just tired and got into my head that single replica == single node, which is not the case. I used "handle" in a loose sense that it would ensure transactional semantics, not that it would resolve the problem, but I reworded to avoid the confusion.

mkindahl:
Closing since it's already described in #555

PR #541, opened 2020-10-19
Update add_data_node() API description

pmwkaa:
Restyle and update arguments description according to the latest changes

### DIFF
--- a/api.md
+++ b/api.md
@@ -260,25 +260,13 @@ error or a notice depending on the value of `if_not_exists`.
 
 If `bootstrap` is true, the function will attempt to bootstrap the
 data node by:
-1. Connecting to the database given in `bootstrap_database`.
-2. Creating the database given in `database` that will serve as the
+1. Creating the database given in `database` that will serve as the
    new data node.
-3. Loading the TimescaleDB extension in the new database.
-4. Setting metadata to make the data node part of the distributed
+2. Loading the TimescaleDB extension in the new database.
+3. Setting metadata to make the data node part of the distributed
    database.
 
 See [boostrapping description][multinode-bootstrap] for details.
-
-#### Errors
-
-An error will be given if:
-* The function is executed inside a transaction.
-* The function is executed in a database that is already a data node.
-* The data node already exists and `if_not_exists` is `FALSE`.
-* The port number is not valid for use as a port number.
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -260,7 +260,7 @@ error or a notice depending on the value of `if_not_exists`.
 
 If `bootstrap` is true, the function will attempt to bootstrap the
 data node by:
-1. Connecting to the database given in `bootstrap_database`.
+1. Connecting to the PostgreSQL maintance database (`postgres` or `template1`).
```
mfreed, line 263: What is a "maintenance" database?  This isn't a term we used before, and sounds like a new type of organizational role which seems confusing / not sure if we want to use?

(Regardless, spelling issue:
```suggestion
1. Connecting to the PostgreSQL mainetance database (`postgres` or `template1`).
```

api.md
```diff
@@ -295,7 +284,6 @@ An error will be given if:
 | `port`               | Port to use on the remote data node. The default is the PostgreSQL port used by the access node on which the function is executed. |
 | `if_not_exists`      | Do not fail if the data node already exists. The default is `FALSE`. |
 | `bootstrap`          | Bootstrap the remote data node. The default is `TRUE`. |
-| `bootstrap_database` | Database to use when bootstrapping as described above. The default is `'postgres'`. |
```
mfreed, line 298: Is this optional at all?  Or it /only/ supports `postgres` or `template1`?  (If both postgres and template1 supported, which one?

pmwkaa: Logic here is equal to the way how PostgreSQL itself handles it. I think we support only those two choices for now. It will try to connect to the `postgres` database first and if it fails then try `template1`: https://github.com/timescale/timescaledb/blob/master/tsl/src/data_node.c#L552

api.md
pmwkaa: I believe "maintenance" is the PostgreSQL terminology, at least they are using it in the source code:  https://github.com/postgres/postgres/blob/master/src/bin/scripts/common.c#L173.

It serves bootstrapping purposes. For example when you need to create a database your client application (libpq/psql) needs to connect to some database first in order to execute `CREATE DATABASE` command.

pmwkaa: I think we could also remove the first step completely

mfreed (approved):

PR #544, opened 2020-10-20
Update release notes with RC1 and RC2 info

mfreed:
Update release notes with information about RC1 and RC2, as well as update about 2.0 description.

### DIFF
--- a/release-notes.md
+++ b/release-notes.md
@@ -12,34 +12,134 @@ can view active developments on GitHub at any time.
 
 ### What to expect from our next release
 
-The team is actively working on the multi-node version
-of TimescaleDB which is currently in beta. You can read more about our
-architecture and design for distributed hypertables
+The team is finishing up work on the first release of distributed hypertables
+supported by multi-node TimescaleDB, which will be part of the TimescaleDB 2.0
+release.  We are currently in the middle of formal Release Candidates (RC) for
+2.0; these RC releases support in-place upgrades from 1.x and subsequent
+upgrades to 2.0.
+
+You can read more about our architecture and design for distributed hypertables
 [here](https://docs.timescale.com/v2.0/introduction/architecture#distributed-hypertables).
 To test out the beta version for yourself, join our #multinode channel on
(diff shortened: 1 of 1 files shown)


release-notes.md
```diff
@@ -12,27 +12,128 @@ can view active developments on GitHub at any time.
...
+multiple instances of TimescaleDB, configured so that one instance serves
+as an access node and multiple others as data nodes. All queries for a
+distributed hypertable are issued to the access node, but inserted data
+and queries are pushed down across data nodes for greater scale and
+performance.
+
+This release also adds support for user-defined actions allowing users to
+define actions that are run by the TimescaleDB automation framework.
+
+In addition to these major new features, the 2.0 branch introduces _breaking_ changes
```
solugebefola, line 95: I think the fact that there are breaking changes should be more emphasized, maybe with a minor headline?

release-notes.md
```diff
@@ -12,27 +12,128 @@ can view active developments on GitHub at any time.
...
+to APIs and existing features, such as continuous aggregates. These changes are not
+backwards compatible and might require changes to clients and/or scripts that rely on
```
solugebefola, line 97:
```suggestion
backwards compatible and may require changes to clients and/or scripts that rely on
```

solugebefola (approved):

release-notes.md
mfreed: Given that this is already in github history, and we typically just copy github notes into the docs, I'm going to skip this grammatical change.  Thanks for careful read, though.

release-notes.md
mfreed: Yeah, will add to front-piece, and we'll be sure to include in 2.0 announcement.

mfreed: Actually, @ryanbooz has some pointers in his PR to the *changes* document; I think I'll highlight the breaking changes after I merge the two (i.e., in a later PR).

ryanbooz (approved):

PR #554, opened 2020-10-22
First draft of chapter on GUC

mkindahl:
This contains a brief dump of existing GUC variables and their
descriptions and default values. The section for each variable needs to
be filled in.


mfreed (approved):
Looks good.  Will be moving location in nav/index in follow-on PR.

PR #562, opened 2020-10-27
Add multinode role-adding instructions

WireBaron:
This change adds a section to the basic multinode setup document
that describes how to add roles to a permissive multinode setup.

It also updates several distributed_exec invocations to use CALL
instead of select.

PR #569, opened 2020-10-28
Fix distributed_exec to correspond to the implementation

k-rus:
distributed_exec was fixed in timescale/timescaledb#2520 to allow
execution of statements outside transactions in addition to
transactional executions. It was also changed from being a function to
be a procedure. This commit aligns the description of distributed_exec
and examples to use CALL.

### DIFF
--- a/api.md
+++ b/api.md
@@ -909,10 +909,16 @@ SELECT detach_tablespaces('conditions');
 ---
 ## distributed_exec() [](distributed_exec)
 
-This function is used on an access node to execute a SQL command
+This procedure is used on an access node to execute a SQL command
 across the data nodes of a distributed database. For instance, one use
 case is to create the roles and permissions needed in a distributed
 database.
+
+The procedure can run distributed commands transactionally, so a command
+is executed either everywhere or nowhere. However, not all SQL commands can run in a 
+transaction. This can be toggled with the argument `transactional`. Note if the execution 
+is not transactional, a failure on one of the data node will require manual dealing with
+the appeared inconsistency.
 
 Note that the command is _not_ executed on the access node itself and
 it is not possible to chain multiple commands together in one call.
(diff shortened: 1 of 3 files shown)


api.md
```diff
@@ -928,23 +928,30 @@ it is not possible to chain multiple commands together in one call.
...
 | `node_list` | An array of data nodes where the command should be executed. Defaults to all data nodes if not specified. |
+| `transactional` | Allows to specify if the execution of the statement should be transactional or not. Defaults to TRUE. |
 
 #### Sample Usage [](distributed_exec-examples)
 
 Create the role `davide` across all data nodes in a distributed database:
 
 ```sql
-SELECT distributed_exec($$CREATE USER davide WITH PASSWORD 'jw8s0F4'$$);
+CALL distributed_exec($$CREATE USER davide WITH PASSWORD 'jw8s0F4'$$);
```
pmwkaa, line 938: Maybe use some kind of password template here instead? Feels like some users might create roles using this password :)

api.md
```diff
@@ -928,23 +928,30 @@ it is not possible to chain multiple commands together in one call.
...
 
 ```
 
 Create the role `davide` on two specific data nodes:
```
pmwkaa, line 942: Likely creating roles only on a limited set of data nodes will require to set grants specific to them as well. I am wondering if it needs to be covered somewhere

pmwkaa (approved):
Not sure if need to update the description, but there are some limitations when using transaction mode. For example some DDL's cannot be run in a multi-statement transaction. And not using an active transaction might require manual cleanup in case if a data node fails.

api.md
k-rus: Thank you for noting. I switched the example to use the one from the certificate authorisation setup.

pmwkaa: One more thing. I think this example should be in sync with @WireBaron update on authentication setup, regarding required usage of `LOGIN`, e.g.:

`CALL distributed_exec($$ CREATE USER testrole WITH LOGIN PASSWORD 'password' $$); `

pmwkaa: It might be also a good idea to put a link to that document

k-rus:
> Not sure if need to update the description, but there are some limitations when using transaction mode. For example some DDL's cannot be run in a multi-statement transaction. And not using an active transaction might require manual cleanup in case if a data node fails.

@pmwkaa Good point! I added a paragraph to deal with this. Can you check it?

api.md
k-rus: @pmwkaa I think your comment is more relevant to https://github.com/timescale/docs.timescale.com-content/pull/562

api.md
```diff
@@ -909,11 +909,17 @@ SELECT detach_tablespaces('conditions');
...
-This function is used on an access node to execute a SQL command
+This procedure is used on an access node to execute a SQL command
 across the data nodes of a distributed database. For instance, one use
 case is to create the roles and permissions needed in a distributed
 database.
 
+The procedure can run distributed commands transactionally, so a command
+is executed either everywhere or nowhere. However, not all SQL commands can run in a 
+transaction. This can be toggled with the argument `transactional`. Note if the execution 
+is not transactional, a failure on one of the data node will require manual dealing with
```
pmwkaa, line 920: `a failure on one of the data node might require manual cleanup` maybe?

pmwkaa (approved):

api.md
k-rus: @pmwkaa It is not necessary cleanup. The user might use different strategy, e.g., manually redoing the command on the failed node.

api.md
k-rus: @pmwkaa I use the example from the certification authorisation. No password is needed in such case.

api.md
```diff
@@ -909,11 +909,17 @@ SELECT detach_tablespaces('conditions');
...
+the appeared inconsistency.
```
WireBaron, line 921:
```suggestion
any introduced inconsistency.
```

WireBaron (approved):

mfreed (approved):

PR #581, opened 2020-11-10
Add Forge installation page

solugebefola:


getting-started/installation-timescale-forge.md
```diff
@@ -0,0 +1,26 @@
+>:WARNING: Our scaling out capabilities are currently in BETA and
```
mfreed, line 1: I think we should remove this warning on multi-node?  Not sure why it's specific to Forge?

mfreed (approved):

PR #582, opened 2020-11-11
Update instructions for adding new data nodes

erimatnor:
Clarify instructions on how to add new data nodes, in particular how
to configure password authentication.

The new `password` parameter has been added to `add_data_node` and a
brief mention is made of how user mappings can be used as an
alternative to a password file.

The need to create a `ssl_cert` role when setting up certificate
authentication has been removed.

### DIFF
--- a/api.md
+++ b/api.md
@@ -280,27 +280,37 @@ with [`add_data_node`](#add_data_node) first before attaching it.
 ---
 ## add_data_node() [](add_data_node)
 
-Add a new data node to the database to be used for creating
-distributed hypertables.
-
-Newly created distributed hypertables will be able to make use of this
-data node for apportioning data across the distributed database. Note
-that existing distributed hypertables will not automatically use the
-newly added data node. For existing distributed hypertables to use
-added data nodes, use [`attach_data_node`](#attach_data_node).
+Add a new data node on the access node to be used by distributed
+hypertables. The data node will automatically be used by distributed
+hypertables that are created after the data node has been added, while
+existing distributed hypertables require an additional
+[`attach_data_node`](#attach_data_node).
 
 If the data node already exists, the command will abort with either an
 error or a notice depending on the value of `if_not_exists`.
 
-If `bootstrap` is true, the function will attempt to bootstrap the
-data node by:
+For security purposes, only superusers or users with necessary
+privileges can add data nodes (see below for details). When adding a
+data node, the access node will also try to connect to the data node
+and therefore needs a way to authenticate with it. TimescaleDB
+currently supports several different such authentication methods for
+flexibility (including trust, user mappings, password, and certificate
+methods). Please refer to [Setting up Multi-Node
+TimescaleDB][multinode] for more information about node-to-node
+authentication.
+
+Unless `bootstrap` is false, the function will attempt to bootstrap
+the data node by:
 1. Creating the database given in `database` that will serve as the
    new data node.
 2. Loading the TimescaleDB extension in the new database.
 3. Setting metadata to make the data node part of the distributed
    database.
 
-See [boostrapping description][multinode-bootstrap] for details.
+Note that user roles are not automatically created on the new data
+node during bootstrapping. The [`distributed_exec`](#distributed_exec)
+procedure can be used to create additional roles on the data node
+after it is added.
 
 #### Required Arguments [](add_data_node-required-arguments)
 
@@ -317,6 +327,7 @@ See [boostrapping description][multinode-bootstrap] for details.
 | `port`               | Port to use on the remote data node. The default is the PostgreSQL port used by the access node on which the function is executed. |
 | `if_not_exists`      | Do not fail if the data node already exists. The default is `FALSE`. |
 | `bootstrap`          | Bootstrap the remote data node. The default is `TRUE`. |
+| `password`           | Password for authenticating with the remote data node during bootstrapping or validation. A password only needs to be provided if the data node requires password authentication and a password for the user does not exist in a local password file on the access node. If password authentication is not used, the specified password will be ignored. |
 
 #### Returns [](add_data_node-returns)
 
@@ -336,9 +347,25 @@ An error will be given if:
 * The function is executed inside a transaction.
 * The function is executed in a database that is already a data node.
 * The data node already exists and `if_not_exists` is `FALSE`.
-* The port number is not valid for use as a port number.
+* The access node cannot connect to the data node due to a network
+  failure or invalid configuration (e.g., wrong port, or there is no
+  way to authenticate the user).
 * If `bootstrap` is `FALSE` and the database was not previously
   bootstrapped.
+
+#### Privileges
+
+To add a data node, you must be a superuser or have the `USAGE`
+privilege on the `timescaledb_fdw` foreign data wrapper. To grant such
+privileges to a regular user role, do:
+
+```sql
+GRANT USAGE ON FOREIGN DATA WRAPPER timescaledb_fdw TO <newrole>;
+```
+
+Note, however, that superuser privileges might still be necessary on
+the data node in order to bootstrap it, including creating the
+TimescaleDB extension on the data node unless it is already installed.
 
 #### Sample usage [](add_data_node-examples)
 
@@ -3717,5 +3744,4 @@ and then inspect `dump_file.txt` before sending it together with a bug report or
 [telemetry]: /using-timescaledb/telemetry
 [caveats]: /using-timescaledb/continuous-aggregates
 [backup-restore]: /using-timescaledb/backup#pg_dump-pg_restore
-[multinode-bootstrap]: /multinode/bootstrapping
 [real-time-aggregates]: /using-timescaledb/continuous-aggregates#real-time-aggregates
--- a/getting-started/setup-multi-node-basic.md
+++ b/getting-started/setup-multi-node-basic.md
@@ -3,18 +3,19 @@
 A multi-node TimescaleDB implementation consists of: 
 - One access node to handle ingest, data routing and act as an entry 
 point for user access; 
-- One or more data nodes to store and organize distributed data.     
+- One or more data nodes to store and organize distributed data.
 
-All nodes begin as single-node TimescaleDBs, i.e., hosts with a running 
-PostgreSQL instance and a loaded TimescaleDB extension. This is assumed for 
-"access node" and "data node" in the instructions. More detail on 
-the architecture can be found in the [Architecture][architecture] section.
+All nodes begin as standalone TimescaleDB instances, i.e., hosts with
+a running PostgreSQL server and a loaded TimescaleDB extension. This
+is assumed for "access node" and "data node" in the instructions. More
+detail on the architecture can be found in the
+[Architecture][architecture] section.
 
 TimescaleDB multi-node can be created as part of a self-managed deployment
 or (coming soon) as a managed cloud deployment.  In order to set up a
 self-managed cluster, including how to configure the nodes for secure
 communication and creating users/roles across servers, please follow
-[these instructions][advanced setup].
+[these instructions][advanced setup] before proceeding.
 
 In the case of Timescale Cloud and Forge, the created services already contain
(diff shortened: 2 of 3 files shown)


pmwkaa (approved):
Looks good overall, yet I am slightly skeptical about recommending the scram-256 for password based auth. I am curious if we ever tested it

api.md
```diff
@@ -280,27 +280,45 @@ with [`add_data_node`](#add_data_node) first before attaching it.
...
-See [boostrapping description][multinode-bootstrap] for details.
+Note that, even if bootstrapping is not requested, the `add_data_node`
+command will try to connect to the remote data node in order to verify
+that it has a valid configuration and that the necessary database
+objects are installed. Distributed user roles are not automatically
+created on the new data node, however, and this might require
+additional setup on the new data node after it is added. The
+[`distributed_exec`](#distributed_exec) procedure can be used to
+create additional database objects on the data node after it is added,
+including distributed roles.
```
mkindahl, line 321: I think you need to show an example here.

api.md
```diff
@@ -317,6 +335,7 @@ See [boostrapping description][multinode-bootstrap] for details.
 | `port`               | Port to use on the remote data node. The default is the PostgreSQL port used by the access node on which the function is executed. |
 | `if_not_exists`      | Do not fail if the data node already exists. The default is `FALSE`. |
 | `bootstrap`          | Bootstrap the remote data node. The default is `TRUE`. |
+| `password`           | Password for authenticating with the remote data node during bootstrapping or validation. A password only needs to be provided if the data node requires password authentication and a password for the user does not exist in a local password file on the access node. |
```
mkindahl, line 338: What happens if a password is provided by it does not use password authentication?

getting-started/setup-multi-node-basic.md
```diff
@@ -23,30 +24,70 @@ node, and follow the instructions in the next section.  More information will be
...
 
 ```sql
 SELECT add_data_node('example_node_name', host => 'example_host_address');
 ```
 
-`example_node_name` should be a unique name for the node. `example_host_address` 
-is the host name or IP address of the data node.
+`example_node_name` should be a unique name for the
+node. `example_host_address` is the host name or IP address of the
+data node. It is also possible to specify a password to authenticate
```
mkindahl, line 43:
```suggestion
data node. You can specify a password to authenticate
```

getting-started/setup-multi-node-basic.md
```diff
@@ -23,30 +24,70 @@ node, and follow the instructions in the next section.  More information will be
...
+and the password is not provided through other means (e.g., a local
+password file). See the [`add_data_node`][add_data_node] API reference
+documentation for detailed information about this command.
+
+It should now be possible to create distributed hypertables. Note,
+however, that in order to create and use distributed hypertables as a
+non-superuser, the user role needs to be distributed as well (i.e.,
+exist on all data nodes) and requires privileges to use the data nodes
+on the access node. Please refer to the next section for instruction
+on how to create distributed roles and grant data node privileges.
```
mkindahl, line 55: Nit: the wording suggest that there is such a thing as a "distributed role" so I suggest to be more clear, also, suggest to remove the "should" and "possible" for a more straightforward wording.
```suggestion
You can now create distributed hypertables using `create_distributed_hypertable`, but note that in order to create and use distributed hypertables as a
non-superuser, the user role needs to exist on all data nodes and have the correct privileges to use the data nodes
on the access node. Please refer to the next section for instruction
on how to create roles on all data nodes and grant data node privileges.
```

getting-started/setup-multi-node-basic.md
```diff
@@ -23,30 +24,70 @@ node, and follow the instructions in the next section.  More information will be
...
 
-The system should be up and running at this point.
+## (Optional) Add distributed user roles
 
-## (Optional) Add user roles to the distributed database
+Like any other database object, roles are _not_ distributed by
+default. In other words, a newly created role on the access node will
```
mkindahl, line 60: Hypertables and chunks are distributed, so this is not entirely true. Suggest to reformulate. (Just discussing roles is fine.)

getting-started/setup-multi-node-basic.md
```diff
@@ -23,30 +24,70 @@ node, and follow the instructions in the next section.  More information will be
...
+
+It should now be possible to create distributed hypertables. Note,
+however, that in order to create and use distributed hypertables as a
+non-superuser, the user role needs to be distributed as well (i.e.,
+exist on all data nodes) and requires privileges to use the data nodes
+on the access node. Please refer to the next section for instruction
+on how to create distributed roles and grant data node privileges.
 
-The system should be up and running at this point.
+## (Optional) Add distributed user roles
```
mkindahl, line 57: The notion of a "distributed role" suggest that it is somehow maintained and is a novel concept. This section only talks about how to add roles on all nodes of a distributed database.
```suggestion
## (Optional) Adding roles for distributed hypertables
```

getting-started/setup-multi-node-basic.md
```diff
@@ -23,30 +24,70 @@ node, and follow the instructions in the next section.  More information will be
...
+Like any other database object, roles are _not_ distributed by
+default. In other words, a newly created role on the access node will
+not exist on the data nodes and it will not be possible to use the
+role to create and query distributed hypertables unless additional
+steps are taken.
 
-After adding your data nodes, it is possible to create and use distributed hypertables,
-but only if running as a superuser common to all nodes (e.g., by default,
-the `postgres` user).
+To create a distributed role from the access node, use these commands:
```
mkindahl, line 65:
```suggestion
To create a role on all data nodes from the access node, use these commands:
```

api.md
```diff
@@ -280,27 +280,45 @@ with [`add_data_node`](#add_data_node) first before attaching it.
...
+hypertables. This function needs to be executed on the access node by
+a superuser or a user with `USAGE` privileges on the `timescaledb_fdw`
+foreign data wrapper (superuser privileges might still be required on
+the data node). It is assumed that the user role adding a data node
+already has a way to authenticate with the data node being
+added. Alternatively, a password can be provided on the command line
+if the data node accepts password-based authentication and no password
+is provided in a local password file on the access node. Please refer
+to [Setting up Multi-Node TimescaleDB][multinode] for more information
+about node-to-node authentication.
```
mfreed, line 293: After reviewing this, this feels like maybe too much in the weeds for the API docs?  Maybe something simpler like:

```suggestion
Add a new data node to the database to be used by distributed
hypertables.   For security purposes, the data node validates the connection
attempt from the access node; TimescaleDB currently supports several 
different such authentication methods for flexibility (including trust, user
mappings, password, and certificate methods).  Please refer
to [Setting up Multi-Node TimescaleDB][multinode] for more information
about node-to-node authentication.
```

getting-started/setup-multi-node-basic.md
```diff
@@ -23,30 +24,70 @@ node, and follow the instructions in the next section.  More information will be
...
+with using the optional `password` parameter. But this is only
+necessary if password authentication is used to connect to data nodes
+and the password is not provided through other means (e.g., a local
+password file). See the [`add_data_node`][add_data_node] API reference
+documentation for detailed information about this command.
+
+It should now be possible to create distributed hypertables. Note,
+however, that in order to create and use distributed hypertables as a
+non-superuser, the user role needs to be distributed as well (i.e.,
+exist on all data nodes) and requires privileges to use the data nodes
```
mfreed, line 53: Otherwise, it sounds more like you always need ot take this step, which you often might not need to do for testing.
```suggestion
It should now be possible to create distributed hypertables, e.g., 
running as the default `postgres` superuser.  If you wish to
create and use distributed hypertables as a
non-superuser, the user role needs to be distributed as well (i.e.,
```

getting-started/setup-multi-node-basic.md
mfreed: Ha.  Mats already recommended.  Feel free to review both =)

mfreed (approved):

api.md
erimatnor: There's a link to section that provides examples. I don't think it is appropriate to have examples about one function in a section about another function.

api.md
erimatnor: I think permissions/superuser requirements should be part of reference docs. This will probably trip people up otherwise. Don't mind shortening in general though.

api.md
```diff
@@ -280,27 +280,45 @@ with [`add_data_node`](#add_data_node) first before attaching it.
...
+
+Newly created distributed hypertables will be able to make use of the
 data node for apportioning data across the distributed database. Note
 that existing distributed hypertables will not automatically use the
 newly added data node. For existing distributed hypertables to use
-added data nodes, use [`attach_data_node`](#attach_data_node).
+added data nodes, use [`attach_data_node`](#attach_data_node) after
+having added the data node.
```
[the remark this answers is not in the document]
erimatnor: Maybe I am missing something, but I see virtually no overlap with the previous paragraph. This is discussing need for attaching after adding, while the first paragraph talks about requirements for auth and privileges in order to execute add_data_node.

I will still try to streamline the text though.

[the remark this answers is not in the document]
mkindahl: You mention above that a new data node is added to be used by distributed tables, and then here again that newly added data nodes are used by the distribute hypertables. IMHO, the first sentence of the paragraph is redundant.

api.md
mkindahl: Then it might be better to only have a notice that the role needs to exists on the data nodes and then refer to the section about how to make that happen. Right now, you start to talk about something the user needs to do, which suggests that it should finish with an elaboration on how to do that.

api.md
erimatnor: Nothing, AFAIK. The method of authentication is determined by the data node.

mkindahl: There is a difference between giving an error if the authentication method does not require a password and just ignoring the password if the authentication method does not require the password. If it is just ignored in this case, it is good to mention it here.

getting-started/setup-multi-node-basic.md
mkindahl: I would probably make this even shorter and just say something like:
> the user role needs to exist on all data nodes and have the correct privileges on both the data and access node.

getting-started/setup-multi-node-basic.md
erimatnor: No, Hypertables and chunks are not distributed by default. You have to explicitly create a distributed hypertable, as opposed to a regular hypertable, for them to be distributed.

getting-started/setup-multi-node-basic.md
erimatnor: "Distributed" is used as adjective here, and I think users know what it means for something to be "distributed".  I don't think it is a bad idea to introduce the concept of distributed roles or, more generally, distributed objects. Otherwise, it becomes very cumbersome to always say "roles that exist on all data nodes". 

I will try to be more "general" for now, but I think we need to figure out the proper terminology to use throughout the documentation.

getting-started/setup-multi-node-basic.md
mkindahl: I think you should just remove the first sentence, it is redundant.

erimatnor:
Did some additional tweaks of the auth instructions.

api.md
erimatnor: I don't see why this is necessary and we generally don't list all error cases for optional parameters. The explanation here is already long.

api.md
```diff
@@ -280,27 +280,37 @@ with [`add_data_node`](#add_data_node) first before attaching it.
...
 
-Add a new data node to the database to be used for creating
-distributed hypertables.
-
-Newly created distributed hypertables will be able to make use of this
-data node for apportioning data across the distributed database. Note
-that existing distributed hypertables will not automatically use the
-newly added data node. For existing distributed hypertables to use
-added data nodes, use [`attach_data_node`](#attach_data_node).
+Add a new data node on the access node to be used by distributed
```
mkindahl, line 283:
```suggestion
Adds a new data node on the access node to be used by distributed
```

mkindahl (approved):

PR #583, opened 2020-11-12
Improve license API description

k-rus:


pmwkaa (commented):
What is OSS installation?

k-rus:
> What is OSS installation?

For example, PostgreSQL includes only OSS binary. Do you have good name? I will try to find a better name.

k-rus:
> > What is OSS installation?
> 
> For example, PostgreSQL includes only OSS binary. Do you have good name? I will try to find a better name.

@pmwkaa I improved the formulation.

pmwkaa (approved):

api.md
```diff
@@ -3663,8 +3663,8 @@ View or set currently used TimescaleDB license.
 It is possible to limit access to features by license by changing the `timescaledb.license`
 settings parameter in the server configuration file or on the server command line. For instance,
 by setting `timescaledb.license` to `apache`, it is only possible to use TimescaleDB features
-that are implemented under the Apache 2 license. The default value is `timescale`, however, which
-allows access to all features.
+that are implemented under the Apache 2 license. The default value is `timescale` and it
+allows access to all features. In the case of only apache binary the license value is `apache`.
```
mfreed, line 3667: I'm not sure I understand why this change is needed?  It seems unnecessary given the previous sentence:

``` For instance,
by setting `timescaledb.license` to `apache`, it is only possible to use TimescaleDB features
that are implemented under the Apache 2 license. The default value is `timescale`, however, which
allows access to all features.
```

This paragraph already there is `apache` and `timescale`.  `apache` only provides access to Apache-2 features, while `timescale` provides access to all remaining feature.  

Why do we need to again add  "In the case of only apache binary the license value is `apache`"?

mfreed (changes requested):

api.md
k-rus: It is not possible to switch from `timescale` license to `apache` license if the binary includes only OSS version.

k-rus:
I actually stumbled on the original formulation with `however` and thought to clarify about OSS binary. Now, I see that my suggestion doesn't make the text much more readable, so closed this version.

PR #592, opened 2020-11-17
Update Changes to link RC3 and clarify refresh

k-rus:
Undocumented refresh on drop functionality of dropping a chunk was
removed in RC3 and replaced with internal API function to refresh a
given chunk. This commit clarifies this change in the Change notes. It
also updates link to the latest RC.

### DIFF
--- a/release-notes/changes-in-timescaledb-2.md
+++ b/release-notes/changes-in-timescaledb-2.md
@@ -8,7 +8,7 @@ for time-series. In addition to multi-node capabilities, this release includes n
 features - and improvements to existing ones - focused on giving users more flexibility,
  control over their data and the ability to customize behavior to suit their needs.
 
-To facilitate many of the improvements in [TimescaleDB 2.0](https://github.com/timescale/timescaledb/releases/tag/2.0.0-rc2),
+To facilitate many of the improvements in [TimescaleDB 2.0](https://github.com/timescale/timescaledb/releases/tag/2.0.0-rc3),
  several existing APIs and function definitions have been modified which may require updates to your existing code. 
 
 Most notably, the following API and PostgreSQL compatibility changes may impact 
@@ -319,6 +319,48 @@ data wasn’t always materialized as expected.
 After upgrading to TimescaleDB 2.0, **retention policies will no longer fail due to incompatibilities with 
 continuous aggregates** and users have to ensure that retention and continuous aggregate policies have the 
 desired interplay. 
+
+Another change in 2.0 is that `drop_chunks` and the retention policy will no longer 
+automatically refresh continuous aggregates to account for changes in original hypertable 
+after the last refresh. Previously, the goal was to ensure that all updates were processed 
+prior to dropping chunks in the original hypertable. In practice, it often didn't work as intended.
+
+In TimescaleDB 2.0 users can ensure that all updates are processed before dropping data by 
+combining the following experimental function, which refreshes updates in the given chunk, 
+with `drop_chunks`.
+```sql
+_timescaledb_internal.refresh_continuous_aggregate(continuous_aggregate REGCLASS, chunk REGCLASS)
+```
+The example below demonstrates how to use the chunk-based refresh function to define a 
+retention policy, which ensures that all updates to data in the original hypertable are 
+refreshed in all continuous aggregates prior dropping chunks older than 2 weeks:
+
+```sql
+CREATE OR REPLACE PROCEDURE refresh_and_drop_policy(job_id int, config jsonb) LANGUAGE PLPGSQL AS
+$$
+DECLARE
+  drop_after interval;
+  hypertable text;
+BEGIN
+  SELECT jsonb_object_field_text(config, 'drop_after')::interval
+    INTO STRICT drop_after;
+  SELECT jsonb_object_field_text(config, 'hypertable')::regclass
+    INTO STRICT hypertable;
+  BEGIN
+    SELECT _timescaledb_internal.refresh_continuous_aggregate(cagg, 
+        show_chunks(older_than => drop_after))
+    FROM _timescaledb_catalog.continuous_agg cagg, _timescaledb_catalog.hypertable hyper
+    WHERE cagg.raw_hypertable_id = hyper.id AND 
+      format('%I.%I', hyper.schema_name, hyper.table_name)::regclass = hypertable;
+
+    SELECT drop_chunks(hypertable, older_than => drop_after);
+  END;
+END
+$$;
+
+SELECT add_job('refresh_and_drop_policy', '2 weeks', 
+   config => '{"hypertable":"public.conditions", "drop_after":"2 weeks"}');
+```
 
 ### Differences in the handling of backfill on continuous aggregates [](cagg-backfill)
 


release-notes/changes-in-timescaledb-2.md
```diff
@@ -320,6 +320,40 @@ After upgrading to TimescaleDB 2.0, **retention policies will no longer fail due
...
+aggregates on the dropped chunks. Previously dropping a chunk would call a refresh on each 
+continuous aggregate to recompute it according to existing invalidations. The refresh was 
+unlikely to include all updates due to `ignore_invalidations_older_than`, which prevents to 
+record invalidations for updates older than the threshold. If a user carefully crafted 
+parameters to the retention policy and the continuous aggregate that no update is missed, 
+the refresh could include previously dropped values leading to unexpected outliers in 
+aggregated values. TimescaleDB 2.0 introduces new function 
+`_timescaledb_internal.refresh_continuous_aggregate`, which refreshes 
+exactly a chunk, and can be used together with drop_chunks. For example, following action can 
+be defined:
```
erimatnor, line 333: I think this is a little bit too much "in the weeds". Suggested simplification:

```suggestion
Another change in 2.0 is that `drop_chunks` and the retention policy will no longer automatically refresh continuous aggregates to account for backfill in chunks when they are dropped. Previously, the goal was to ensure that all backfill in a chunk was processed prior to dropping it. However, this didn't always work as intended (e.g., if the materialization was far behind), and it also made retention policies slow.

If the old behavior of processing a chunk's backfill during drops is still desired, 2.0 offers the experimental function `_timescaledb_internal.refresh_continuous_aggregate`, which refreshes exactly one chunk. The function can be used together with `drop_chunks` in a single transaction, making it possible to define the following user-defined action:

```

release-notes/changes-in-timescaledb-2.md
```diff
@@ -320,6 +320,40 @@ After upgrading to TimescaleDB 2.0, **retention policies will no longer fail due
...
+    SELECT refresh_chunks(cagg, show_chunks(older_than => drop_after))
+    FROM _timescaledb_catalog.continuous_agg cagg, _timescaledb_catalog.hypertable hyper
+    WHERE hyper.id = cagg.raw_hypertable_id AND 
+      hyper.schema_name || '.' || hyper.table_name = hypertable;
+    SELECT drop_chunks(hypertable, older_than => drop_after);
+  END;
+END
+$$;
+```
+
```
erimatnor, line 356: The above user-defined action can be scheduled as a custom policy using the `add_job` function.

k-rus: @erimatnor Will it be good to add call to `add_job` in the script, like in the design doc?

release-notes/changes-in-timescaledb-2.md
k-rus: The suggestion changes the message I was out for. Also the backfill is not the only case. I will try to make my formulation shorter with keeping in mind your suggestion.

mfreed: I like this change, @erimatnor.  Because this is experimental/internal, I assume that's why it's not in the docs.  Should we suggest here what the args are?  Does it take in (chunk, start_offset, end_offset) ?

k-rus: @mfreed I will put the function signature in a separate line, so arguments fit also. I also added an example of a user policy using it. I hope the example will be reduced for the final release as it demonstrates a hole in the information view.

erimatnor: As always, this is a suggestion. The mention of only backfill can easily be modified. My main concern remains: that the original paragraph is too much in the weeds, which gets in the way of the important message that there's a new alternative.

release-notes/changes-in-timescaledb-2.md
erimatnor: I guess it is fine, if we expect users to run this always via policy, as opposed to manually.

release-notes/changes-in-timescaledb-2.md
```diff
@@ -320,6 +320,40 @@ After upgrading to TimescaleDB 2.0, **retention policies will no longer fail due
...
+DECLARE
+  drop_after interval;
+  hypertable text;
+BEGIN
+  SELECT jsonb_object_field_text(config, 'drop_after')::interval
+    INTO STRICT drop_after;
+  SELECT jsonb_object_field_text(config, 'hypertable')
+    INTO STRICT hypertable;
+  BEGIN
+    SELECT refresh_chunks(cagg, show_chunks(older_than => drop_after))
```
erimatnor, line 347: I think you mean this:

```suggestion
    SELECT _timescaledb_internal.refresh_continuous_aggregate(cagg, show_chunks(older_than => drop_after))
```

k-rus: 👍  Thank you for finding! I missed to fix this one after copying from the design doc.

release-notes/changes-in-timescaledb-2.md
k-rus: Thank you for the suggestion, I updated the formulation.

release-notes/changes-in-timescaledb-2.md
k-rus: > The above user-defined action can be scheduled as a custom policy using the `add_job` function.

I interpret it as it is good to show adding the policy with `add_job` in the example, so it is added.

release-notes/changes-in-timescaledb-2.md
k-rus: Thank you! Fixed

mfreed (approved):

release-notes/changes-in-timescaledb-2.md
```diff
@@ -320,6 +320,45 @@ After upgrading to TimescaleDB 2.0, **retention policies will no longer fail due
 continuous aggregates** and users have to ensure that retention and continuous aggregate policies have the 
 desired interplay. 
 
+Another change in 2.0 is that `drop_chunks` and the retention policy will no longer 
+automatically refresh continuous aggregates to account for changes in raw data 
+after the last refresh. Previously, the goal was to ensure that all updates were processed 
+prior dropping. However, it often didn't work as intended.
```
ryanbooz, line 326:
```suggestion
Another change in 2.0 is that invoking `drop_chunks`, either manually or through a retention policy, will no longer 
refresh continuous aggregates for the interval represented by the chunk. Previously, TimescaleDB 1.7 would attempt to materialize any modifications in raw chunk data into the continuous aggregate before the chunk was dropped. In practice, it often didn't work as intended.
```
Based on feedback from Erik, suggesting just a few tweaks. Mostly it wasn't clear as written (to me) that `drop_chunks` was used in both manual and policy updates here.  Take or leave, just suggesting.

release-notes/changes-in-timescaledb-2.md
```diff
@@ -320,6 +320,45 @@ After upgrading to TimescaleDB 2.0, **retention policies will no longer fail due
...
+    INTO STRICT drop_after;
+  SELECT jsonb_object_field_text(config, 'hypertable')::regclass
+    INTO STRICT hypertable;
+  BEGIN
+    SELECT _timescaledb_internal.refresh_continuous_aggregate(cagg, 
+        show_chunks(older_than => drop_after))
+    FROM _timescaledb_catalog.continuous_agg cagg, _timescaledb_catalog.hypertable hyper
+    WHERE cagg.raw_hypertable_id = hyper.id AND 
+      format('%I.%I', hyper.schema_name, hyper.table_name)::regclass = hypertable;
+    SELECT drop_chunks(hypertable, older_than => drop_after);
```
ryanbooz, line 353:
```suggestion
      format('%I.%I', hyper.schema_name, hyper.table_name)::regclass = hypertable;
      
    SELECT drop_chunks(hypertable, older_than => drop_after);
```
For visual clarity I'd add a blank line

ryanbooz (commented):
Overall great. Love the easy SQL example to help folks!

release-notes/changes-in-timescaledb-2.md
k-rus: @ryanbooz Retention policy doesn't call SQL function `drop_chunk`, but both calls the same internal function. I think you are missing that both are performing chunk drops.
I will see how incorporate it better into the text.

erimatnor, line 326:
```suggestion
prior to dropping. However, it often didn't work as intended.
```

release-notes/changes-in-timescaledb-2.md
```diff
@@ -320,6 +320,45 @@ After upgrading to TimescaleDB 2.0, **retention policies will no longer fail due
 continuous aggregates** and users have to ensure that retention and continuous aggregate policies have the 
 desired interplay. 
 
+Another change in 2.0 is that `drop_chunks` and the retention policy will no longer 
+automatically refresh continuous aggregates to account for changes in raw data 
+after the last refresh. Previously, the goal was to ensure that all updates were processed 
+prior dropping. However, it often didn't work as intended.
+
+In TimescaleDB 2.0 users can ensure that all updates are processed before dropping data by 
+combining a following experimental function, which refreshes a chunk, with drop_chunks.
```
erimatnor, line 329:
```suggestion
combining the following experimental function, which refreshes a chunk, with `drop_chunks`.
```

Also, I am always a bit hesitant when we say "refresh a chunk", since we are not really refreshing the chunk itself. We are refreshing the cagg based on data in the chunk.

Maybe this would work (just to avoid that issue):

```suggestion
combining the following experimental chunk-based refresh function with `drop_chunks`.
```

release-notes/changes-in-timescaledb-2.md
```diff
@@ -320,6 +320,45 @@ After upgrading to TimescaleDB 2.0, **retention policies will no longer fail due
...
+        show_chunks(older_than => drop_after))
+    FROM _timescaledb_catalog.continuous_agg cagg, _timescaledb_catalog.hypertable hyper
+    WHERE cagg.raw_hypertable_id = hyper.id AND 
+      format('%I.%I', hyper.schema_name, hyper.table_name)::regclass = hypertable;
+    SELECT drop_chunks(hypertable, older_than => drop_after);
+  END;
+END
+$$;
+
+SELECT add_job('refresh_and_drop_policy', '2 weeks',
```
erimatnor, line 358: I assume this has also been tested in practice? :)

release-notes/changes-in-timescaledb-2.md
```diff
@@ -320,6 +320,45 @@ After upgrading to TimescaleDB 2.0, **retention policies will no longer fail due
...
+automatically refresh continuous aggregates to account for changes in raw data 
+after the last refresh. Previously, the goal was to ensure that all updates were processed 
+prior dropping. However, it often didn't work as intended.
+
+In TimescaleDB 2.0 users can ensure that all updates are processed before dropping data by 
+combining a following experimental function, which refreshes a chunk, with drop_chunks.
+```sql
+_timescaledb_internal.refresh_continuous_aggregate(continuous_aggregate REGCLASS, chunk REGCLASS)
+```
+The following example demonstrates defining a retention policy, which ensures that all updates
```
erimatnor, line 333:
```suggestion
The example below demonstrates how to use the chunk-based refresh function to define a retention policy, which ensures that all updates 
```

erimatnor (approved):
Some nits and minor suggestions.

release-notes/changes-in-timescaledb-2.md
ryanbooz: Gotcha. Main thing I was trying to clarify in context is that this section is about retention policies, which in truth, we never specifically mention that it's dropping chunks. Obviously, it's implied... but I'm just now realizing that as I re-read.  :-)

Anyway, it feels like just a little more context/setup would be helpful.  Maybe something like...

```suggestion
One other change in 2.0 impacts how data in continuous aggregates is updated. When dropping older data from a hypertable, either manually when calling `drop_chunks` or with a data retention policy, the continuous aggregate will no longer be refreshed for the interval represented by the chunk....
```

k-rus: I made minor change to say dropping chunks, but haven't followed suggestion completely as it increases the text. I feel it is not the best place to go into much implementation details.

release-notes/changes-in-timescaledb-2.md
k-rus: > Also, I am always a bit hesitant when we say "refresh a chunk", since we are not really refreshing the chunk itself. We are refreshing the cagg based on data in the chunk.

I was also hesitant to say not fully correctly. On the other side it is not easy to be correct and short. I made an improvement there.

ryanbooz (approved):

PR #601, opened 2020-11-25
Add maintenance job to heal data nodes

erimatnor:
For self-managed multinode deployments, an administrator must configure
a maintenance job that "heals" data nodes that haven't fully completed
their part of distributed transactions.

The description of the job is added to the "advanced" setup section,
since, e.g., Forge and Cloud might automatically add this job.

### DIFF
--- a/getting-started/setup-multi-node-auth.md
+++ b/getting-started/setup-multi-node-auth.md
@@ -465,6 +465,36 @@ CALL distributed_exec($$ CREATE ROLE testrole LOGIN $$);
 ```
 
 ---
+## Data node maintenance job on the access node
+
+It is highly recommended that the access node is configured to run a
+maintenance job that regularly "heals" any non-completed distributed
+transactions. A distributed transaction can remain in a non-completed
+state in case a data node reboots or experiences temporary issues. The
+access node keeps a log of distributed transactions so that nodes that
+haven't yet completed their part of the distributed transaction can
+later complete it at the access node's request. The maintenance job
+can be run as a user-defined action (custom job):
+
+
+```sql
+CREATE OR REPLACE PROCEDURE data_node_maintenance(job_id int, config jsonb)
+LANGUAGE SQL AS
+$$
+    SELECT _timescaledb_internal.remote_txn_heal_data_node(fs.oid)
+    FROM pg_foreign_server fs, pg_foreign_data_wrapper fdw
+    WHERE fs.srvfdw = fdw.oid
+    AND fdw.fdwname = 'timescaledb_fdw';
+$$;
+
+SELECT add_job('data_node_maintenance', '5m');
+```
+
+It is also possible to schedule this job to run from outside the
+database, e.g, via a CRON job. Note that the job must be scheduled
+separately for each distributed database.
+
+---
 
 ## Next steps
 To start working with the system, you can look at documentation for [distributed hypertables][].


solugebefola (commented):
Is there a reason that this is in `setup-multi-node-auth.md` instead of `setup-multi-node-basic.md`?

mfreed (approved):

getting-started/setup-multi-node-auth.md
```diff
@@ -464,6 +464,36 @@ And finally add the role to all of the data nodes with [`distributed_exec`][dist
...
+## Data node maintenance job on the access node
+
+It is highly recommended that the access node is configured to run a
+maintenance job that regularly "heals" any non-completed distributed
+transactions. A distributed transaction can remain in a non-completed
+state in case a data node reboots or experiences temporary issues. The
+access node keeps a log of distributed transactions so that nodes that
+haven't yet completed their part of the distributed transaction can
+later complete it at the access node's request. The maintenance job
+can be run as a user-defined action (custom job):
```
davidkohn88, line 477: Should this link to the user-defined action section of the docs?

getting-started/setup-multi-node-auth.md
```diff
@@ -464,6 +464,36 @@ And finally add the role to all of the data nodes with [`distributed_exec`][dist
...
+
+
+```sql
+CREATE OR REPLACE PROCEDURE data_node_maintenance(job_id int, config jsonb)
+LANGUAGE SQL AS
+$$
+    SELECT _timescaledb_internal.remote_txn_heal_data_node(fs.oid)
```
davidkohn88, line 484: I haven't tested this, so I just want to be sure, this function can't be called inside a transaction block, and I know there's an implicit transaction in single sql statement, but I'm not sure if that would cause issues here or if we'd need to write this as a separate function call per fs.oid with a PLPGSQL loop or something? Anyway, just wanted to make sure we'd tested this with multiple data nodes so that we know there's no txn issues in that case.

davidkohn88 (commented):
Looks good, one question about potential transaction management issues with the heal function, and just want to make sure that we've tested that inside the implicit txn of a single sql statement.

getting-started/setup-multi-node-auth.md
erimatnor: So, if you do `SELECT myfunction();` or `CALL myproc()`, those functions/procedures will execute as part of a single transaction (unless the procedure itself does something funky with transactions). In any case, if there is no explicit `BEGIN;` prior to calling this procedure, there is no transaction block and the procedure will be executed. I did test this with two data nodes, and both the procedure and job executes successfully covering two data nodes (although no actual healing was done).

Obviously, manually executing this query:

```
SELECT _timescaledb_internal.remote_txn_heal_data_node(fs.oid)
    FROM pg_foreign_server fs, pg_foreign_data_wrapper fdw
    WHERE fs.srvfdw = fdw.oid
    AND fdw.fdwname = 'timescaledb_fdw';
```
is a single transaction, but still involves multiple data nodes.

That said, this is @cevian's code and design, so I am not really sure myself what is correct here. A comment in the `remote_txn_heal_data_node` function gives me pause: 

```
	/*
	 * This function cannot be called inside a transaction block since effects
	 * cannot be rolled back
	 */
	PreventInTransactionBlock(true, "remote_txn_heal_data_node");

```
I assume that the intention here is to only execute this on one node per transaction as to ensure that there is not a situation where a failure on the success node means that changes on the first one cannot be rolled back (since this isn't executed under 2PC). I suspect that preventing a transaction block cannot protect against executing this on two data nodes in a single transaction.

erimatnor: FWIW, should verify with @cevian, but I think `remote_txn_heal_data_node` should probably be turned into a procedure that starts a new transaction every time it is called. Otherwise, I think we need to keep some state in the function during the transaction that prevents calling this function multiple times in the same transaction.

davidkohn88: I think something like: 
```
CREATE OR REPLACE PROCEDURE data_node_maintenance(job_id int, config jsonb)
AS $$
DECLARE
    fs_oid oid;
BEGIN
    
    FOR fs_oid IN 
        SELECT fs.oid
        FROM pg_foreign_server fs, pg_foreign_data_wrapper fdw
        WHERE fs.srvfdw = fdw.oid
        AND fdw.fdwname = 'timescaledb_fdw';
    LOOP 
        RAISE NOTICE 'Healing foreign server with OID: %s', fs_oid;
        PERFORM _timescaledb_internal.remote_txn_heal_data_node(fs_oid);
        COMMIT;
    END LOOP;
END;
$$ LANGUAGE PLPGSQL;
```
would solve the problem.

erimatnor: @davidkohn88 Thanks for the suggestion. I think the issue is that you want to commit after healing each data node, to ensure the state of the node is recorded locally on the AN and not later rolled back due to a failure on another data node. 

Still, not 100% sure it is an issue to begin with.

davidkohn88: You can stick a commit in that loop, but I believe in this case, it is going to commit after each perform, there's no implicit txn because we're no longer in a single statement.

cevian: the way this is should be fine. I prevented the txn block was because 
```
BEGIN;
SELECT remote_txn_heal_data_node();
ROLLBACK;
```
is nonesense since the `commit transaction` 2pc command that's sometimes executed on the DN as part of this cannot be rolled back.

erimatnor: Ok, so it sounds like the maintenance job in the PR is OK.

erimatnor:
> Is there a reason that this is in `setup-multi-node-auth.md` instead of `setup-multi-node-basic.md`?

@solugebefola Yes. Basically, I think those file names are misleading. I think `setup-multi-node-auth.md` is about self-managed deployments (rather than authentication specifically), and I imagined that Forge would configure the cleanup job automatically for you.

solugebefola:
If you think the files would be better named as something else, can you make the modification to the file names and the `page-index.js` file with the new names?  It would be good if you included a redirect for the old name to the new name (to catch links in the wild)

erimatnor:
I will merge this as is and then I or someone else can make further changes (including file renames) in as separate PR.

PR #624, opened 2020-12-22
Tweak instructions for Forge multi-node setup

erimatnor:
* Simplify `add_data_node` command to user bare minimum parameters. In
  particular, remove the bootstrap parameter since one shouldn't have
  to care about it unless under special circumstances that do not
  apply to Forge. The port defaults to 5432, so not needed for
  internal connections.
* Add section about maintenance tasks.

### DIFF
--- a/getting-started/forge-multi-node.md
+++ b/getting-started/forge-multi-node.md
@@ -78,7 +78,7 @@ Once you are connected to each service, immediately run the following SQL as the
 **first** command to update TimescaleDB version to 2.0:
 
 ```SQL
-ALTER EXTENSION timescaledb UPDATE TO "2.0.0"; 
+ALTER EXTENSION timescaledb UPDATE TO '2.0.0';
 ```
 >:WARNING:If you created your service before 12/21/2020, note that your version of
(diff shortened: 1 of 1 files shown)


getting-started/forge-multi-node.md
```diff
@@ -137,13 +137,18 @@ connections. Copy this hostname to use in the `add_data_node` command below.
...
+```bash
 psql -h {AN hostname} -p {port} tsdb tsdbadmin
 
-SELECT add_data_node('dn1', host => 'your_DN1_hostname', port => 5432 , password => 'tsdbadmin_user_password_for_DN1', bootstrap => false);
+```
+
+Then add a data node as follows.
+
+```sql
+SELECT add_data_node('dn1', host => 'your_DN1_hostname',
```
erimatnor, line 150: FWIW, @niksajakovljevic brought up the point that eliding `bootstrap=>false` prints some notices about database and extension not being created. If we prefer to not print those notices, we can keep `bootstrap=>false`.

Generally, though, the intention is that you normally shouldn't have to care about bootstrapping. Disabling bootstrapping is only necessary for special cases where the default settings do not work.

solugebefola (approved):

mfreed (approved):

PR #633, opened 2020-12-30
Clarify versions which require database restart

k-rus:
Updates the warning, which notes the need to restart database during
the extension upgrade, with all versions, which require restart
according to the release notes.

### DIFF
--- a/using-timescaledb/update-db.md
+++ b/using-timescaledb/update-db.md
@@ -34,8 +34,11 @@ triggering the load of a previous TimescaleDB version on session startup.
 It must also be the first command you execute in the session. 
 <!-- -->
 
->:WARNING: If upgrading from a TimescaleDB version older than 0.12.0,
+>:WARNING: When upgrading from an old version of TimescaleDB before upgrading 
+to version 0.12.0 or version 1.5.0,
 you will need to restart your database before calling `ALTER EXTENSION`.
+After upgrading to 1.6.1 you will need to restart the database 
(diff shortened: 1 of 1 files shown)


using-timescaledb/update-db.md
```diff
@@ -34,8 +34,11 @@ triggering the load of a previous TimescaleDB version on session startup.
 It must also be the first command you execute in the session. 
 <!-- -->
 
->:WARNING: If upgrading from a TimescaleDB version older than 0.12.0,
+>:WARNING: When upgrading from an old version of TimescaleDB before upgrading 
+to version 0.12.0 or version 1.5.0,
 you will need to restart your database before calling `ALTER EXTENSION`.
+After you upgraded to 1.6.1 you will need to restart the database
```
solugebefola, line 40:
```suggestion
After upgrading to 1.6.1 you will need to restart the database 
```

solugebefola: Dunno if this should be in a separate warning.  Also, maybe these should just be bolded notes for folks updating from particular versions, rather than warnings?

mfreed (approved):

using-timescaledb/update-db.md
k-rus: It is already in the release notes for corresponding release. However, the general note was only for 0.12. I guess some users update through several versions at once, so they might not check notes for each version.

solugebefola (approved):

PR #638, opened 2021-01-12
Clarify supported ALTER MATERIALIZE VIEW options

k-rus:
Four PG clauses of ALTER MATERIALIZE VIEW are supported on continuous
aggregate views.

### DIFF
--- a/api.md
+++ b/api.md
@@ -1756,7 +1756,16 @@ WITH (timescaledb.continuous) AS
 ---
 
 ## ALTER MATERIALIZED VIEW (Continuous Aggregate) :community_function: [](continuous_aggregate-alter_view)
-`ALTER MATERIALIZED VIEW` statement can be used to modify some of the `WITH` clause [options](#continuous_aggregate-create_view-with-optional) for the continuous aggregate view.
+`ALTER MATERIALIZED VIEW` statement can be used to modify some of the `WITH` clause [options](#continuous_aggregate-create_view-with-optional) for the continuous aggregate view. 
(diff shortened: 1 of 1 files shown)


erimatnor (approved):

api.md
```diff
@@ -1756,7 +1756,16 @@ WITH (timescaledb.continuous) AS
...
+continuous aggregate view: 
+
+- `RENAME TO` clause to rename the continuous aggregate view;
+- `SET SCHEMA` clause to set the new schema for the continuous aggregate view;
+- `SET TABLESPACE` clause to move the materialization of the continuous 
+  aggregate view to the new tablespace;
+- `OWNER TO` clause to set new owner for the continuous aggregate view.
 
 ``` sql
 ALTER MATERIALIZED VIEW <view_name> SET ( timescaledb.<option> =  <value> [, ... ] )
```
mfreed, line 1771: Minor suggestion:  Maybe give one example with the clauses?

ALTER MATERIALIZED VIEW foo SET (blah) WITH OWNER TO X, SET TABLESPACE Y;

mfreed (approved):

api.md
k-rus: Then it will be necessary to explain all the parameters. 

I made first version of this proposal, where I was verbose, but it was too much, since they cannot be chained as you suggest. Each needs to be a separate statement. Thus I provided link to PG doc, which contains the syntax, instead.

WDYT? Can I keep as it is?

PR #653, opened 2021-02-05
add promscale getting started docs

VineethReddy02:
Add Promscale getting started docs


ryanbooz:
It looks like somehow nothing got committed with this file @VineethReddy02??

That said, I've been working on the MD file locally with some formatting suggestions, on top of proposing that we **do** break this down into multiple files with child links in the navigation like the following (titles can change to whatever makes sense). Otherwise, this just ends up being a really, really long document.

Thoughts?

Screen Shot 2021-02-05 at 2 21 07 PM

solugebefola:
Closed in favor of another PR

PR #657, opened 2021-02-14
Update continuous-aggregates.md

SiddiqueAhmad:
hourly view does not have column device_id so running this query will fail, i feel it is typo error


solugebefola (approved):

solugebefola:
Thanks for the contribution!

PR #660, opened 2021-02-18
Add example with getting hypertable size across all hypertables

mfreed:


davidkohn88:
Thanks for the report @akamensky!

PR #661, opened 2021-02-18
Add example with getting hypertable size across all hypertables

mfreed:


davidkohn88 (commented):
I might add a similar note to the `timescaledb_information.hypertables` view section and link to the hypertable size functions, as folks might look there as well.

davidkohn88 (approved):
(but looks fine otherwise)

PR #666, opened 2021-02-23
Clarify cagg refresh window requirements

erimatnor:
When refreshing a continuous aggregate or creating a policy, we have
specific requirements on the given refresh windows. Clarify the
requirements and tweak the related documentation.

### DIFF
--- a/api.md
+++ b/api.md
@@ -1838,29 +1838,78 @@ DROP MATERIALIZED VIEW contagg_view;
 
 ## refresh_continuous_aggregate() :community_function: [](refresh_continuous_aggregate)
 
-Refresh all buckets of a continuous aggregate between two points of
-time. 
+Refresh all buckets of a continuous aggregate in the _refresh window_
+given by `window_start` and `window_end`.
 
-The function expects the parameter values to have the same time type
-as used in the continuous aggregate's time bucket expression (e.g., if
-the time bucket specifies in `timestamptz`, then the start and end time
-supplied should also be `timestamptz`).
+A continuous aggregate materializes aggregates in time buckets (e.g.,
+min, max, average over 1 day worth of data), as determined by the
+`time_bucket` interval specified when the continuous aggregate was
+created. Therefore, when refreshing the continuous aggregate, only
+buckets that completely fit within the refresh window will be
+refreshed. In other words, it is not possible to compute the aggregate
+over, for example, half a bucket. Therefore, any buckets that do no
+fit within the given refresh window will be excluded.
+
+The function expects the window parameter values to have a time type
+that is compatible with the continuous aggregate's time bucket
+expression&mdash;for example, if the time bucket is specified in
+`TIMESTAMP WITH TIME ZONE`, then the start and end time should be a
+date or timestamp type. Note that a continuous aggregate using the
+`TIMESTAMP WITH TIME ZONE` type aligns with the UTC time zone, so, if
+`window_start` and `window_end` is specified in the local time zone,
+any time zone shift relative UTC needs to be accounted for when
+refreshing in order to align with bucket boundaries (for examples, see
+[Sample Usage](#refresh_continuous_aggregate-examples)).
+
 
 #### Required Arguments [](refresh_continuous_aggregate-required-arguments)
 
 |Name|Description|
 |---|---|
 | `continuous_aggregate` | (REGCLASS) The continuous aggregate to refresh. |
-| `window_start` | Start of the window to refresh, has to be before `window_end`. `NULL` is eqivalent to `MIN(timestamp)` of the hypertable. |
-| `window_end` | End of the window to refresh, has to be after `window_start`. `NULL` is eqivalent to `MAX(timestamp)` of the hypertable. |
+| `window_start` | Start of the window to refresh (inclusive), has to be before `window_end`. `NULL` is equivalent to an open-ended start of the window. |
+| `window_end` | End of the window to refresh (exclusive), has to be after `window_start`. `NULL` is equivalent to open-ended end of the window. |
+
+#### Errors
+
+An error is given if:
+
+- The refresh window does not cover at least one full bucket. Note
+  that, even though the window is one bucket in size, it might not
+  align with the start and the end of a bucket in the continuous
+  aggregate.
+- The function is called in a transaction block.
+  
 
 #### Sample Usage [](refresh_continuous_aggregate-examples)
 
-Refresh the continuous aggregate `conditions` between `2020-01-01` and
-`2020-02-01` exclusive.
+A continuous aggregate `conditions_summary` uses the `TIMESTAMP WITH TIME
+ZONE` type and has a bucket size of 1 day. To refresh only 1 day
+(bucket) between `2020-01-01` (inclusive) and `2020-01-02`
+(exclusive):
 
 ```sql
-CALL refresh_continuous_aggregate('conditions', '2020-01-01', '2020-02-01');
+CALL refresh_continuous_aggregate('conditions_summary', '2020-01-01 00:00 UTC', '2020-01-02 00:00 UTC');
+```
+
+If the local time zone is already UTC, the time zone information can be elided:
+
+```sql
+CALL refresh_continuous_aggregate('conditions_summary', '2020-01-01', '2020-01-02');
+```
+
+Alternatively, refresh at least two buckets to ensure at least one
+bucket is refreshed, irrespective of local time zone:
+
+```sql
+CALL refresh_continuous_aggregate('conditions_summary', '2020-01-01', '2020-01-03');
+```
+
+To refresh a longer interval, e.g., one month between `2020-01-01`
+(inclusive) and `2020-02-01` (exclusive):
+
+```sql
+CALL refresh_continuous_aggregate('conditions_summary', '2020-01-01 00:00 UTC', '2020-02-01 00:00 UTC');
 ```
 
 ---
@@ -1883,7 +1932,18 @@ Information about jobs created by policies can be viewed by querying
 
 ## add_continuous_aggregate_policy() :community_function: [](add_continuous_aggregate_policy)
 
-Create a policy that automatically refreshes a continuous aggregate.
+Create a policy that automatically refreshes a continuous
+aggregate. When the policy runs it refreshes the continuous aggregate
+in the refresh window range given by the `start_offset` and
+`end_offset` relative the current time when the policy executes. Since
+a refresh only materializes full buckets, and the policy execution
+time likely won't align with the end of a bucket, the policy's refresh
+window must be at least two buckets in size to ensure that at least
+one bucket is refreshed when the policy runs. Note that the manual
+refresh function
+[`refresh_continuous_aggregate`](#refresh_continuous_aggregate) allows
+specifying a window of only one bucket, since it is possible to
+manually align its refresh window with bucket boundaries.
 
 #### Required Arguments [](add_continuous_aggregate_policy-required-arguments)
 
@@ -1912,15 +1972,28 @@ The `start_offset` and `end_offset` parameters should be specified differently d
 |`job_id`| (INTEGER)  TimescaleDB background job id created to implement this policy|
 
 
+#### Errors
+
+An error will be given if:
+
+- The `start_offset` or `end_offset` specify a range less than two buckets in size.
+- The type of `start_offset` or `end_offset` is not compatible with
+  the time type of the continuous aggregate.
+
 #### Sample Usage [](add_continuous_aggregate_policy-examples)
 
-Add a policy that refreshes the last month once an hour, excluding the latest hour from the aggregate (for performance reasons, it is recommended to exclude buckets that still see lots of writes):
+Add a policy that refreshes the last month once an hour, excluding the
+latest hour from the aggregate:
+
 ```sql
 SELECT add_continuous_aggregate_policy('conditions_summary',
 	start_offset => INTERVAL '1 month',
 	end_offset => INTERVAL '1 hour',
 	schedule_interval => INTERVAL '1 hour');
 ```
+
+Note that, for performance reasons, it is recommended to exclude
+buckets that still see lots of writes.
 
 ---
 ## add_job() :community_function: [](add_job)


api.md
```diff
@@ -1832,31 +1832,68 @@ DROP MATERIALIZED VIEW contagg_view;
...
+determined by the continuous aggregate view query's `time_bucket`
+expression. Therefore, only buckets that are completely covered by the
+_refresh window_ will be updated, and the window should account for
+any time zone shifts when applicable (note that a continuous aggregate
+using the `DATE` or `TIMESTAMP WITH TIME ZONE` type aligns with the
+UTC time zone).
+
+The function expects the window parameter values to have the same time
+type as used in the continuous aggregate's time bucket expression
+(e.g., if the time bucket specifies in `TIMESTAMP WITH TIME ZONE`,
```
solugebefola, line 1848: > if the time bucket specifies in

"is specified in"?

solugebefola (approved):

api.md
```diff
@@ -1832,31 +1832,68 @@ DROP MATERIALIZED VIEW contagg_view;
...
+To align the start and end of the refresh window with the bucket time
+zone, specify UTC. For instance, to refresh only 1 day/bucket:
+
+
+```sql
+CALL refresh_continuous_aggregate('conditions', '2020-01-01 00:00 UTC', '2020-01-02 00:00 UTC');
+```
+
+To ensure that at least one bucket is refreshed, without accounting
+for time zone, specify a window of at least two buckets:
```
ryanbooz, line 1890: I'm honestly not sure how to do it, but it feels like this "at least two buckets" comment needs to be highlighted in some way. Assuming you're putting something together for the policy... maybe just link there??

Not a show stopper at all, just about to record examples of this and the wider your bucket is the more this will probably impact you.

ryanbooz: Oops... just keep scrolling Ryan... my apologies.

k-rus (approved):
LGTM
I am not sure if all examples work as expected, see my specific comment. Few comments for possible improvements.

api.md
```diff
@@ -1832,31 +1832,68 @@ DROP MATERIALIZED VIEW contagg_view;
...
-as used in the continuous aggregate's time bucket expression (e.g., if
-the time bucket specifies in `timestamptz`, then the start and end time
-supplied should also be `timestamptz`).
+The refresh will only materialize at the granularity of a bucket, as
+determined by the continuous aggregate view query's `time_bucket`
+expression. Therefore, only buckets that are completely covered by the
+_refresh window_ will be updated, and the window should account for
+any time zone shifts when applicable (note that a continuous aggregate
+using the `DATE` or `TIMESTAMP WITH TIME ZONE` type aligns with the
+UTC time zone).
```
mkindahl, line 1844: "Granularity" is a little to esoteric and suggest to use plural form for "continuous aggregates" (since we use that elsewhere). Also, parentheses have a tendency to clutter the text and are usually avoided or at least limited to very small phrases. An attempt at a rewrite below, but not entirely happy about it so feel free to tweak it.

```suggestion
Refresh process entire buckets and not partial buckets. This means that only buckets that are completely covered by the refresh window will be updated, and the window should account for
time zone shifts (when applicable). Note that continuous aggregates
using the `DATE` or `TIMESTAMP WITH TIME ZONE` type align with the
UTC time zone.
```
It is also not very clear what "aligns with the [...] time zone" means, so maybe just mention that or say that `DATE` and `TIMESTAMP WITH TIME ZONE` works/do not works/something else.

api.md
```diff
@@ -1832,31 +1832,68 @@ DROP MATERIALIZED VIEW contagg_view;
...
-| `window_start` | Start of the window to refresh, has to be before `window_end`. `NULL` is eqivalent to `MIN(timestamp)` of the hypertable. |
-| `window_end` | End of the window to refresh, has to be after `window_start`. `NULL` is eqivalent to `MAX(timestamp)` of the hypertable. |
+| `window_start` | Start of the window to refresh (inclusive), has to be before `window_end`. `NULL` is equivalent to refreshing from the beginning of time. |
+| `window_end` | End of the window to refresh (exclusive), has to be after `window_start`. `NULL` is equivalent to refreshing to the end of time. |
+
+#### Errors
+
+An error is given if:
+
+- The refresh window does not cover a full bucket. Note that, even
```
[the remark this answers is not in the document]
mkindahl: "at least one full bucket"?

api.md
```diff
@@ -1832,31 +1832,68 @@ DROP MATERIALIZED VIEW contagg_view;
...
+- The function is called in a transaction block.
+  
 
 #### Sample Usage [](refresh_continuous_aggregate-examples)
 
-Refresh the continuous aggregate `conditions` between `2020-01-01` and
-`2020-02-01` exclusive.
+A continuous aggregate `conditions` has a bucket size of 1 day. To
+refresh `conditions` between `2020-01-01` and `2020-02-01` exclusive,
+do:
```
mkindahl, line 1875: Not clear which one of them are inclusive and exclusive.
```suggestion
A continuous aggregate `conditions` has a bucket size of 1 day. To
refresh `conditions` between `2020-01-01` (inclusive) and `2020-02-01` (exclusive):
```

api.md
```diff
@@ -1906,9 +1949,20 @@ The `start_offset` and `end_offset` parameters should be specified differently d
...
+- The `start_offset` or `end_offset` specify a range less than two buckets in size.
+- The type of `start_offset` or `end_offset` is not compatible with
+  the time type of the continuous aggregate.
+
 #### Sample Usage [](add_continuous_aggregate_policy-examples)
 
-Add a policy that refreshes the last month once an hour, excluding the latest hour from the aggregate (for performance reasons, it is recommended to exclude buckets that still see lots of writes):
+Add a policy that refreshes the last month once an hour, excluding the
+latest hour from the aggregate (for performance reasons, it is
+recommended to exclude buckets that still see lots of writes):
```
mkindahl, line 1964: IMHO, it flows better if the comment in parentheses is after the sample code and without parentheses.

api.md
```diff
@@ -1832,31 +1832,68 @@ DROP MATERIALIZED VIEW contagg_view;
...
 
 #### Required Arguments [](refresh_continuous_aggregate-required-arguments)
 
 |Name|Description|
 |---|---|
 | `continuous_aggregate` | (REGCLASS) The continuous aggregate to refresh. |
-| `window_start` | Start of the window to refresh, has to be before `window_end`. `NULL` is eqivalent to `MIN(timestamp)` of the hypertable. |
-| `window_end` | End of the window to refresh, has to be after `window_start`. `NULL` is eqivalent to `MAX(timestamp)` of the hypertable. |
+| `window_start` | Start of the window to refresh (inclusive), has to be before `window_end`. `NULL` is equivalent to refreshing from the beginning of time. |
+| `window_end` | End of the window to refresh (exclusive), has to be after `window_start`. `NULL` is equivalent to refreshing to the end of time. |
```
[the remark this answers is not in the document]
mkindahl: Maybe just call it open-ended, people are familiar with that term even in non-technical writing.

api.md
```diff
@@ -1832,31 +1832,68 @@ DROP MATERIALIZED VIEW contagg_view;
 
 ## refresh_continuous_aggregate() :community_function: [](refresh_continuous_aggregate)
 
-Refresh all buckets of a continuous aggregate between two points of
+Refresh all buckets of a continuous aggregate between two points in
```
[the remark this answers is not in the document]
k-rus: @mkindahl Your suggestion gives impression that the both points need to be provided, while they can be omitted. So the word `given` needs to be replaces with something more correct, I believe.

api.md
```diff
@@ -1906,9 +1949,20 @@ The `start_offset` and `end_offset` parameters should be specified differently d
 |`job_id`| (INTEGER)  TimescaleDB background job id created to implement this policy|
 
 
+#### Errors
+
+An error will be given if:
+
+- The `start_offset` or `end_offset` specify a range less than two buckets in size.
```
k-rus, line 1956: Do I follow correctly that the policy requires 2 buckets while the refresh function only 1 bucket? (just checking)

api.md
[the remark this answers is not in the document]
mkindahl: @k-rus Well, since each point can be NULL (which means that it's open) they *are* both provided.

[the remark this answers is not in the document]
k-rus: You are right, I forgot that we decided to always require both points.

api.md
```diff
@@ -1832,31 +1832,68 @@ DROP MATERIALIZED VIEW contagg_view;
...
+
 
 #### Required Arguments [](refresh_continuous_aggregate-required-arguments)
 
 |Name|Description|
 |---|---|
 | `continuous_aggregate` | (REGCLASS) The continuous aggregate to refresh. |
-| `window_start` | Start of the window to refresh, has to be before `window_end`. `NULL` is eqivalent to `MIN(timestamp)` of the hypertable. |
-| `window_end` | End of the window to refresh, has to be after `window_start`. `NULL` is eqivalent to `MAX(timestamp)` of the hypertable. |
+| `window_start` | Start of the window to refresh (inclusive), has to be before `window_end`. `NULL` is equivalent to refreshing from the beginning of time. |
```
ryanbooz, line 1858:
```suggestion
| `window_start` | Start of the window to refresh (inclusive), has to be before `window_end`. `NULL` is equivalent to refreshing from the earliest point of data in the underlying hypertable |
```

api.md
ryanbooz, line 1859:
```suggestion
| `window_end` | End of the window to refresh (exclusive), has to be after `window_start`. `NULL` is equivalent to refreshing to the most recent point of data in the underlying hypertable |
```

api.md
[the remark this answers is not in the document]
ryanbooz: Agreed

api.md
ryanbooz: Yes, the reality is that you _can_ supply the correct values to update one full bucket to the manual refresh, but a policy is always going to execute at a time that almost never would get you an exact bucket width if you only updated one bucket of time.

I think this is a little confusing overall, but it does provide a little bit more flexibility to users in the end.

k-rus: Thank you for clarifying! It will be definitely confusing to users. However, changing the bevaviour is outside this PR.

@erimatnor May be provide a place where the acceptable bucket sizes are explicitly mentioned between the policy and the function. Otherwise, some users might remember only one of the limitations or think that one of them a typo.

erimatnor: @k-rus I am not sure I understand what you are suggesting, in particular: "May be provide a place where the acceptable bucket sizes are explicitly mentioned between the policy and the function."

erimatnor: FWIW, it is correct that the policy requires two buckets and refresh function only one, for the very reason @ryanbooz explains.

I think the refresh function is doing things correctly, and the policy is the awkward interface. This is, IMO, a _general_ problem with the background scheduling framework and policies in general. For instance, there is no way to control when the policy runs; one would likely prefer CRON-like scheduling over what is implemented now (there are also multiple issues for this). Perhaps allowing people to say "always refresh the latest bucket or something" is a better interface, but that also has issues since it isn't obvious what the bucket size is. 

I do think this might be a smaller problem that we think though. Policies that refresh less than two buckets don't make a whole lot of sense anyway, and I think it is much better to raise an error if people try to create them than silently accepting something that is semi-broken. So hopefully this is not the common case and a big problem in the end.

k-rus: It is common that users try to specify policy to refresh only one bucket, I've seen this number of times.
It is confusing/bad UX that the policy and refresh functions accept different refresh windows. So users might miss that the requirements are different between the policy and the function. Thus I suggest to mention explicitly that the refresh function and the cagg policy have different minimal requirement on the refresh window, e.g.,
> Note that the refresh window in the policy needs to be at least two buckets, while in the refresh function at least one bucket.

api.md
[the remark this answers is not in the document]
erimatnor: I think you missed that one example refreshes a month and one example one day. But I will try to clarify nevertheless.

api.md
erimatnor: @mkindahl  I don't understand what you are suggesting in your last paragraph.

erimatnor: I am not sure "granularity" is a particularly difficult word to understand, especially for a developer audience. Still, I will try to avoid this term.

api.md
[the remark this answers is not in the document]
k-rus: I copied from the wrong place. The case I was looking into is:
> CALL refresh_continuous_aggregate('conditions', '2020-01-01 00:00 UTC', '2020-01-02 00:00 UTC');
As it was explained it works with the refresh function.

api.md
erimatnor: Technically, this isn't correct. Let's say you removed/dropped data in the hypertable. You are still able to refresh those regions that have no data and actually remove the data in the continuous aggregate.

api.md
erimatnor: Same here as above.

api.md
```diff
@@ -1832,29 +1832,75 @@ DROP MATERIALIZED VIEW contagg_view;
...
+  align with the start and the end of a bucket in the continuous
+  aggregate.
+- The function is called in a transaction block.
+  
 
 #### Sample Usage [](refresh_continuous_aggregate-examples)
 
-Refresh the continuous aggregate `conditions` between `2020-01-01` and
-`2020-02-01` exclusive.
+A continuous aggregate `conditions` uses the `TIMESTAMP WITH TIME
```
k-rus, line 1877: `conditions` is quite common name for hypertables in our examples, thus it is strange to see as the name of a continuous aggregate.  Examples in the create cagg uses `continuous_aggregate_view`.
It is quite common to use the bucket size in the name of a cagg. May be `conditions_cagg_1d`? In such case it would be good to fix the create cagg example to use the same name.

api.md
```diff
@@ -1832,29 +1832,75 @@ DROP MATERIALIZED VIEW contagg_view;
...
 
 #### Required Arguments [](refresh_continuous_aggregate-required-arguments)
 
 |Name|Description|
 |---|---|
 | `continuous_aggregate` | (REGCLASS) The continuous aggregate to refresh. |
-| `window_start` | Start of the window to refresh, has to be before `window_end`. `NULL` is eqivalent to `MIN(timestamp)` of the hypertable. |
-| `window_end` | End of the window to refresh, has to be after `window_start`. `NULL` is eqivalent to `MAX(timestamp)` of the hypertable. |
+| `window_start` | Start of the window to refresh (inclusive), has to be before `window_end`. `NULL` is equivalent to an open-ended start of the window. |
+| `window_end` | End of the window to refresh (exclusive), has to be after `window_start`. `NULL` is equivalent to open-ended end of the window. |
```
k-rus, line 1862: open-ended start and end of the window is very mathematical/scientific and unlikely be understood by users. What does NULL translate too? Will it be MIN and MAX values of the time type? If so, can it be such specific?

k-rus (approved):
I have few suggestions for improvements.

I still think it is important to clarify to users that refresh window has different requirements in the refresh function and in the policy.

api.md
mkindahl: Not really limited to science or mathematics, it's in general use for anything that does not have a predetermined limit. See, for example, definition in Cambridge Dictionary (https://dictionary.cambridge.org/dictionary/english/open-ended).

api.md
mkindahl: I have used `conditions` for hypertables and `conditions_summary` (`conditions_weekly_summary`,  `conditions_monthly_summary`, etc.) for various continuous aggregates.

api.md
mkindahl: Still suggest to move the comment in the parentheses to after the code sample.

api.md
```diff
@@ -1832,29 +1832,75 @@ DROP MATERIALIZED VIEW contagg_view;
...
+any buckets that do no fit within the given refresh window will be
+excluded.
+
+The function expects the window parameter values to have a time type
+that is compatible with the continuous aggregate's time bucket
+expression (e.g., if the time bucket is specified in `TIMESTAMP WITH
+TIME ZONE`, then the start and end time supplied should be a date or
+timestamp type). Note that a continuous aggregate using the `TIMESTAMP
+WITH TIME ZONE` type aligns with the UTC time zone, so any potential
+time zone shifts need to be accounted for when refreshing.
```
mkindahl, line 1853: I suggest to recast the sentence so that the comment in parentheses is not needed and can be, for example, a separate sentence. Remarks inside sentences, either using parentheses or em-dashes, have a tendency to make the sentence harder to read since the reader is detracted from following the main topic of the sentence.

Also, It is still not very clear if the fact that a type aligns with the UTC time zone means I can use it or not use it directly or what measures I need to take if I want to use it correctly.

api.md
```diff
@@ -1832,29 +1832,75 @@ DROP MATERIALIZED VIEW contagg_view;
...
-time. 
+Refresh all buckets of a continuous aggregate in the _refresh window_
+given by `window_start` and `window_end`.
+
+A continuous aggregate materializes aggregates in time buckets (e.g.,
+min, max, average over 10 minutes of data), as determined by the
+continuous aggregate view query's `time_bucket` expression. Therefore,
+when refreshing the continuous aggregate, only buckets that completely
+fit within the refresh window will be refreshed. In other words, it is
+not possible to compute the aggregate over, e.g., half a bucket, and
```
mkindahl, line 1843: I usually avoid "e.g." and "i.e." in more formal writing, such as manuals and books, and instead use "for example" and "that is". It is easier on the eye.
```suggestion
not possible to compute the aggregate over, for example, half a bucket, and
```

api.md
k-rus: I still doubt that many users will understand it.

api.md
```diff
@@ -1832,29 +1832,75 @@ DROP MATERIALIZED VIEW contagg_view;
...
 ## refresh_continuous_aggregate() :community_function: [](refresh_continuous_aggregate)
 
-Refresh all buckets of a continuous aggregate between two points of
-time. 
+Refresh all buckets of a continuous aggregate in the _refresh window_
+given by `window_start` and `window_end`.
+
+A continuous aggregate materializes aggregates in time buckets (e.g.,
+min, max, average over 10 minutes of data), as determined by the
+continuous aggregate view query's `time_bucket` expression. Therefore,
```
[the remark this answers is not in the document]
erimatnor: Maybe? The only concern I have is that it is confused with a query to read the continuous aggregate.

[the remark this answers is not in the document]
k-rus: Then it might be better to reformulate this considerably. I don't think it is easy to follow so long phrase.

api.md
erimatnor: Strictly, saying MAX (exclusive) means the max value is not refreshed. Also, because of bucketing, we might not strictly refresh to the min or max since that might not cover a full bucket. We could still say that we refresh to min or max if that is preferred, but I also fear that this can be misinterpreted as refreshing to the min and max value in the hypertable (which is the reason I changed this formulation).

k-rus: My point is that I don't think it is easy to follow the current formulation.

erimatnor: Right now I don't know how to address this since there is only one individual that thinks this is unclear and using min/max as suggested has other drawbacks. No other suggestions have been given either.

api.md
```diff
@@ -1838,29 +1838,77 @@ DROP MATERIALIZED VIEW contagg_view;
...
+`time_bucket` interval specified when the continuous aggregate was
+created. Therefore, when refreshing the continuous aggregate, only
+buckets that completely fit within the refresh window will be
+refreshed. In other words, it is not possible to compute the aggregate
+over, for example, half a bucket. Therefore, any buckets that do no
+fit within the given refresh window will be excluded.
+
+The function expects the window parameter values to have a time type
+that is compatible with the continuous aggregate's time bucket
+expression---for example, if the time bucket is specified in
```
mkindahl, line 1855: If you really want an m-dash, I think you should use `&mdash;`. An simpler alternative is to split it into two separate sentences.
```suggestion
expression. For example, if the time bucket is specified in
```

api.md
```diff
@@ -1838,29 +1838,77 @@ DROP MATERIALIZED VIEW contagg_view;
...
+`TIMESTAMP WITH TIME ZONE`, then the start and end time should be a
+date or timestamp type. Note that a continuous aggregate using the
```
mkindahl, line 1857:
```suggestion
date or timestamp type. Note that continuous aggregates using the
```

api.md
```diff
@@ -1838,29 +1838,77 @@ DROP MATERIALIZED VIEW contagg_view;
...
+`TIMESTAMP WITH TIME ZONE` type aligns with the UTC time zone, so, if
+`window_start` and `window_end` is specified in the local time zone,
+any time zone shift relative UTC needs to be accounted for when
+refreshing in order to align with bucket boundaries.
```
mkindahl, line 1861: Still hard for the user to understand what she needs to do. Not sure how to rewrite this in a good way. Maybe provide an example?

mkindahl (approved):
Some minor things that you can handle if you like, but I think it is good to go.

api.md
k-rus: I guess only me and Mats seen this new formulation. May be @ryanbooz can look to this new version of the formulation.

k-rus: Just based on what you wrote:
```suggestion
| `window_start` | Start of the window to refresh (inclusive), has to be before `window_end`. `NULL` is equivalent to the inclusive start of the first possible bucket for the time dimension. |
| `window_end` | End of the window to refresh (exclusive), has to be after `window_start`. `NULL` is equivalent to the exclusive end of the last possible bucket for the time dimension.. |
```
Will it be correct formulation?

k-rus:
@erimatnor I don't think it is possible to publish documentation per patch version. I.e., I don't think `2.0.2` can be used and it should be either `2.0` or `2.1`.

api.md
erimatnor: Examples are provided in the example section.

PR #669, opened 2021-02-25
Minor change for new 4TB size update in Forge

ryanbooz:
The upcoming Forge release will increase the available storage size to 4TB. There was a small text change to match that. Images for the 4TB update are uploaded separately.


lukas-bernert (approved):
lgtm

PR #673, opened 2021-03-02
Add create_distributed_restore_point() description

pmwkaa:
New API function for creating distributed restore points: https://github.com/timescale/timescaledb/pull/2843

### DIFF
--- a/api.md
+++ b/api.md
@@ -29,6 +29,7 @@
 > - [detach_tablespace](#detach_tablespace)
 > - [detach_tablespaces](#detach_tablespaces)
 > - [distributed_exec](#distributed_exec)
+> - [create_distributed_restore_point](#create_distributed_restore_point)
 > - [drop_chunks](#drop_chunks)
 > - [drop materialized view (continuous aggregate)](#continuous_aggregate-drop_view)
 > - [first](#first)
@@ -1032,6 +1033,61 @@ Create new databases `dist_database` on data nodes, which requires to set `trans
 
 ```sql
 CALL distributed_exec('CREATE DATABASE dist_database', transactional => FALSE);
+```
+
+---
+
+## create_distributed_restore_point() [](create_distributed_restore_point)
+
+Create a same-named marker record, i.e., "restore point", in the write-ahead logs of all nodes in a
+multi-node TimescaleDB cluster.
+
+The restore point can later be used as a recovery target on each node, ensuring the entire multi-node cluster
+can be restored to a consistent state. The function returns the write-ahead log locations for all
+nodes where the marker record was written.
+
+This function is an analogue of [`pg_create_restore_point`](https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADMIN-BACKUP-TABLE) designed
+to work with a distributed database.
+
+This function can only be run on the access node and requires superuser privileges.
+
+#### Required Arguments [](create_distributed_restore_point-required-arguments)
+
+|Name|Description|
+|---|---|
+| `name` | (TEXT) The restore point name. |
+
+#### Errors
+
+An error will be given if:
+- Restore point `name` is more than 64 chars.
+- Recovery is in progress.
+- Current WAL level is not set to `replica` or `logical`.
+- Current user is not a superuser.
+- Current server is not the access node.
+- TimescaleDB's 2PC transactions are disabled.
+
+#### Returns [](create_distributed_restore_point-returns)
+
+| Column          | Description                               |
+|-----------------|-------------------------------------------|
+| `node_name`     | (NAME) Node name (or `NULL` for access node).    |
+| `node_type`     | (TEXT) Node type name (`access_node` or `data_node`). |
+| `restore_point` | ([PG_LSN](https://www.postgresql.org/docs/current/datatype-pg-lsn.html)) Restore point log sequence number. |
+
+#### Sample Usage [](create_distributed_restore_point-examples)
+
+Create the restore point `pitr` across three data nodes and access node:
+
+```sql
+SELECT * FROM create_distributed_restore_point('pitr');
+ node_name |  node_type  | restore_point 
+-----------+-------------+---------------
+           | access_node | 0/3694A30
+ dn1       | data_node   | 0/3694A98
+ dn2       | data_node   | 0/3694B00
+ dn3       | data_node   | 0/3694B68
+(4 rows)
 ```
 
 ---


api.md
```diff
@@ -1034,6 +1035,43 @@ Create new databases `dist_database` on data nodes, which requires to set `trans
 CALL distributed_exec('CREATE DATABASE dist_database', transactional => FALSE);
 ```
 
+---
+## create_distributed_restore_point() [](create_distributed_restore_point)
+
+Create a named point for performing restore across a distributed database.
+
+This function is an analogue of `pg_create_restore_point()` designed to work with a distributed database
```
k-rus, line 1043: Can you add a link to PG documentation, which describes the function?

api.md
```diff
@@ -1034,6 +1035,43 @@ Create new databases `dist_database` on data nodes, which requires to set `trans
...
+and taking into account synchronization of the remote transactions.
+
```
k-rus, line 1045: I think it is important to describe what it does:
- It will obtain locks and lock all distributed hypertables for queries to have a consistent state at the restore point.
- It will create restore point on every data node and the access node with the given name.

api.md
```diff
@@ -1034,6 +1035,43 @@ Create new databases `dist_database` on data nodes, which requires to set `trans
...
+|---|---|
+| `name` | The restore point name. |
+
+#### Returns
+
+| Column          | Description                               |
+|-----------------|-------------------------------------------|
+| `node_name`     | Node name (or `NULL` for access node).    |
+| `node_type`     | Node type (`access_node` or `data_node`). |
+| `restore_point` | Restore point LSN.                        |
```
k-rus, line 1058: Is it possible to add link to a PG page describing LSN?

api.md
```diff
@@ -1034,6 +1035,43 @@ Create new databases `dist_database` on data nodes, which requires to set `trans
...
+SELECT * FROM create_distributed_restore_point('pitr');
+ node_name |  node_type  | restore_point 
+-----------+-------------+---------------
+           | access_node | 0/3694A30
+ dn1       | data_node   | 0/3694A98
+ dn2       | data_node   | 0/3694B00
+ dn3       | data_node   | 0/3694B68
+(4 rows)
+```
+
```
k-rus, line 1074: It is good to describe all cases when an error will happen.

k-rus (changes requested):
Good starting point. I suggest to give more details about the function and describe errors.

Also links to PG documentation might be useful.

api.md
pmwkaa: I guess I can add more info about it, but just to note it does not lock any hypertables at all. It is more about remote transaction synchronization and postponing `COMMIT PREPARE` command in some cases to ensure consistency. Not really sure that this implementation details should be exposed here.

k-rus: Sorry, I believe I wasn't correct in the formulation. I feel it will be good for users to know that the distributed restore point is blocking, so uncommitted distributed transactions will wait until the function completes. Do I follow correctly?

pmwkaa: It does not block remote transactions completely, only when they decide it is time to commit. This commit operation might be postponed until the function completes, in order to achieve consistent WAL history across the cluster. I don't really think this should be mentioned in more details then necessary in the API description, unless people disagree.

api.md
pmwkaa: I've updated the description with requirements, not really sure what other errors are expected to be described here

api.md
k-rus: I see the improvement. I think it is good.

api.md
```diff
@@ -1036,6 +1037,50 @@ CALL distributed_exec('CREATE DATABASE dist_database', transactional => FALSE);
 
 ---
 
+## create_distributed_restore_point() [](create_distributed_restore_point)
+
+Create a named point for performing restore across a distributed database.
+
+This function is an analogue of [pg_create_restore_point()](https://www.postgresql.org/docs/current/functions-admin.html) designed
```
k-rus, line 1044: The rest of the API doc doesn't use function parentheses after the function name during referring.
You can also use code inside reference. I suggest:
```suggestion
This function is an analogue of [`pg_create_restore_point`](https://www.postgresql.org/docs/current/functions-admin.html) designed
```

api.md
```diff
@@ -1036,6 +1037,50 @@ CALL distributed_exec('CREATE DATABASE dist_database', transactional => FALSE);
...
+|---|---|
+| `name` | The restore point name. |
+
+#### Returns [](create_distributed_restore_point-returns)
+
+| Column          | Description                               |
+|-----------------|-------------------------------------------|
+| `node_name`     | Node name (or `NULL` for access node)    |
+| `node_type`     | Node type (`access_node` or `data_node`) |
+| `restore_point` | Restore point [LSN](https://www.postgresql.org/docs/9.4/datatype-pg-lsn.html) |
```
k-rus, line 1065: You don't want to refer version 9.4 of docs.
```suggestion
| `restore_point` | Restore point [LSN](https://www.postgresql.org/docs/current/datatype-pg-lsn.html) |
```

k-rus (commented):
Two comments on referencing. One is critical to not link PG 9.4 doc.

api.md
```diff
@@ -1036,6 +1037,50 @@ CALL distributed_exec('CREATE DATABASE dist_database', transactional => FALSE);
 
 ---
 
+## create_distributed_restore_point() [](create_distributed_restore_point)
+
+Create a named point for performing restore across a distributed database.
+
+This function is an analogue of [pg_create_restore_point](https://www.postgresql.org/docs/current/functions-admin.html) designed
```
k-rus, line 1044: I still think it is good to put the function name into code as in the rest of the doc:
```suggestion
This function is an analogue of [`pg_create_restore_point`](https://www.postgresql.org/docs/current/functions-admin.html) designed
```

api.md
k-rus: According to the implementation there are many cases when error will be returned. See https://github.com/timescale/timescaledb/pull/2843/files#diff-cf55e2617d97e45d62909cc52c6382e2ae31e0cb57e40e246638c73036f6d057R77-R114

k-rus (changes requested):
Error cases need to be described.
One previous comment was neither fully addressed nor explanation was given why to not address.
See the comments.

api.md
pmwkaa: You sure we need that change? it doesn't make much sense from markdown perspective, since link will be highlighted anyway. Also couldn't find any similar way of doing as you suggested in the API doc, do you have any reference?

k-rus: It does. See an example in the first sentence under https://docs.timescale.com/latest/api#attach_data_node

api.md
```diff
@@ -1036,6 +1037,60 @@ CALL distributed_exec('CREATE DATABASE dist_database', transactional => FALSE);
...
+
+#### Errors
+
+An error will be given if:
+- Restore point `name` is more than 64 chars.
+- Recovery is in progress.
+- Current WAL level is not set to `replica` or `logical`.
+- Current user is not a superuser.
+- Current server is not the access node.
+- Timescaledb 2PC transactions are disabled.
```
k-rus, line 1067: The correct name is `TimescaleDB`:
```suggestion
- TimescaleDB's 2PC transactions are disabled.
```

k-rus (approved):
It is good to properly capitalize the product name.

api.md
```diff
@@ -1036,6 +1037,60 @@ CALL distributed_exec('CREATE DATABASE dist_database', transactional => FALSE);
...
+Create a named point for performing restore across a distributed database.
+
+This function is an analogue of [`pg_create_restore_point`](https://www.postgresql.org/docs/current/functions-admin.html) designed
+to work with a distributed database.
+
+On successful execution the same named restore point will be created on the access node and each
+associated data node. It synchronizes execution of ongoing remote transactions to ensure that
+each node will have a consistent restore point view across the cluster.
+
+This function can only be run on the access node and requires superuser preveleges.
```
mfreed, line 1051:
```suggestion
This function can only be run on the access node and requires superuser privileges.
```

api.md
```diff
@@ -1036,6 +1037,60 @@ CALL distributed_exec('CREATE DATABASE dist_database', transactional => FALSE);
...
+- Current server is not the access node.
+- TimescaleDB's 2PC transactions are disabled.
+
+#### Returns [](create_distributed_restore_point-returns)
+
+| Column          | Description                               |
+|-----------------|-------------------------------------------|
+| `node_name`     | Node name (or `NULL` for access node)    |
+| `node_type`     | Node type (`access_node` or `data_node`) |
+| `restore_point` | Restore point [LSN](https://www.postgresql.org/docs/current/datatype-pg-lsn.html) |
```
mfreed, line 1075: Describe the type of returned values?

api.md
```diff
@@ -1036,6 +1037,60 @@ CALL distributed_exec('CREATE DATABASE dist_database', transactional => FALSE);
...
+## create_distributed_restore_point() [](create_distributed_restore_point)
+
+Create a named point for performing restore across a distributed database.
+
+This function is an analogue of [`pg_create_restore_point`](https://www.postgresql.org/docs/current/functions-admin.html) designed
+to work with a distributed database.
+
+On successful execution the same named restore point will be created on the access node and each
+associated data node. It synchronizes execution of ongoing remote transactions to ensure that
+each node will have a consistent restore point view across the cluster.
```
mfreed, line 1049: Add something why you want this?  e.g.,:

"With such a restore point, and backup/restore point should return to the same restore point across nodes in a multi-node TimescaleDB cluster, to ensure a consistent point in time for recovery."

api.md
```diff
@@ -1036,6 +1037,60 @@ CALL distributed_exec('CREATE DATABASE dist_database', transactional => FALSE);
 
 ---
 
+## create_distributed_restore_point() [](create_distributed_restore_point)
+
+Create a named point for performing restore across a distributed database.
```
mfreed, line 1042: Consistency, do we say "distributed database" or "multi-node TimescaleDB cluster"?

api.md
```diff
@@ -1036,6 +1037,60 @@ CALL distributed_exec('CREATE DATABASE dist_database', transactional => FALSE);
...
+
+This function is an analogue of [`pg_create_restore_point`](https://www.postgresql.org/docs/current/functions-admin.html) designed
+to work with a distributed database.
+
+On successful execution the same named restore point will be created on the access node and each
```
mfreed, line 1047:
```suggestion
On successful execution, the same named restore point will be created on the access node and on each
```

mfreed (approved):
Looks good other than small comments.  Are these other parts in the multi-node or backup/restore sections that we want to reference this?  Or wait until more operational information to share (separate PR)?

api.md
```diff
@@ -1036,6 +1037,63 @@ CALL distributed_exec('CREATE DATABASE dist_database', transactional => FALSE);
 
 ---
 
+## create_distributed_restore_point() [](create_distributed_restore_point)
+
+Create a named point for performing restore across a multi-node TimescaleDB cluster.
+
+This function is an analogue of [`pg_create_restore_point`](https://www.postgresql.org/docs/current/functions-admin.html) designed
```
erimatnor, line 1044: I suggest using the link to the table where the function is specified instead of the top-level page: 

```suggestion
This function is an analogue of [`pg_create_restore_point`](https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADMIN-BACKUP-TABLE) designed
```

api.md
```diff
@@ -1036,6 +1037,63 @@ CALL distributed_exec('CREATE DATABASE dist_database', transactional => FALSE);
 
 ---
 
+## create_distributed_restore_point() [](create_distributed_restore_point)
+
+Create a named point for performing restore across a multi-node TimescaleDB cluster.
```
erimatnor, line 1042: Make it similar to the PG documentation, but adapt to distributed case:

```suggestion
Create a named marker record in the write-ahead logs of all nodes in a multi-node TimescaleDB cluster. This named marker, i.e., "restore point", can later be used as a recovery target, ensuring the entire multi-node cluster can be restored to a consistent state. The function returns the write-ahead log locations for all nodes where the marker record was written.
```

api.md
```diff
@@ -1036,6 +1037,63 @@ CALL distributed_exec('CREATE DATABASE dist_database', transactional => FALSE);
...
+
+This function is an analogue of [`pg_create_restore_point`](https://www.postgresql.org/docs/current/functions-admin.html) designed
+to work with a distributed database.
+
+On successful execution, the same named restore point will be created on the access node and on each
+associated data node. It synchronizes execution of ongoing remote transactions to ensure that
+each node will have a consistent restore point view across the cluster.
+
+With such a restore point, and backup/restore point should return to the same restore point
+across nodes in a multi-node TimescaleDB cluster, to ensure a consistent point in time for recovery.
```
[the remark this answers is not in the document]
pmwkaa: I've replaced it with the paragraph that Erik suggested below

erimatnor (approved):

PR #700, opened 2021-04-03
Include restore points in backup/restore section

pmwkaa:
Since restore points are now supported for the multi-node environment, remove previous warning about it and update the section

### DIFF
--- a/page-index/page-index.js
+++ b/page-index/page-index.js
@@ -658,6 +658,11 @@ const pageIndex = [
             type: ANCHOR,
             href: "#docker-wale",
           },
+          {
+            Title: "Restore points",
+            type: ANCHOR,
+            href: "#restore_points",
+          },
         ],
       },
       {
--- a/using-timescaledb/backup.md
+++ b/using-timescaledb/backup.md
@@ -6,8 +6,6 @@ physical backups with `pg_basebackup` or another tool, or logical backups with
 `pg_dump` and `pg_restore`. Physical backups may also be used with Write-Ahead Log
 (WAL) archiving to achieve an ongoing backup.
 
->:WARNING:TimescaleDB currently does not natively support a consistent restore point for multi-node environments. Care should be taken to ensure third-party solutions properly quiesce the environment before backing up, so that the backup point used across nodes does not have outstanding transactions.
-
 ## Performing Physical Backups [](physical-backups)
 
 For full instance physical backups (which are especially useful for starting up
@@ -17,7 +15,6 @@ external backup and restore managers such as [`pg_backrest`][pg-backrest],
 [`barman`][pg-barman], or [`wal-e`][wale official]. These allow you to take
 online, hot physical backups of your entire instance, and many offer incremental
 backups and other automation options.
-
 
 
 ## Using `pg_dump` and `pg_restore` [](pg_dump-pg_restore)
@@ -331,6 +328,26 @@ end as the recovery will be complete when no further files can be
 found in the archive. See the PostgreSQL documentation on
 [continuous archiving][pg archiving] for more information.
 
+### Restore points [](restore_points)
+
+TimescaleDB natively supports a restore point both for single-node and for
+multi-node environments. The restore point can later be used as a recovery target on
+single-node or on each node of the entire multi-node cluster, ensuring that it can be
+restored to a consistent state.
+
+To create a restore point on a single-node, the [`pg_create_restore_point`][pg_create_restore_point]
+function can be used.
+
+On a multi-node cluster, the [`create_distributed_restore_point`][create_distributed_restore_point] function must be used
+instead and executed on the access node. In order to achieve synchronization across the multi-node cluster, this function ensures
+that the restore point created on the access node is synchronized with each data node. Care should be taken to ensure
+that the data nodes have no locally-running user transactions executed without the access node's involvement when
+creating distributed restore point (i.e., by an external user connecting directly to a data node and executing a local
+transaction on that data node).
+
+Third-party backup and restore managers such as [`pg_backrest`][pg-backrest],
+[`barman`][pg-barman], or [`wal-e`][wale official] can be used and configured to use the restore points.
+
 [replication-tutorial]: /tutorials/replication
 [postgres-pg_basebackup]: https://www.postgresql.org/docs/current/app-pgbasebackup.html
 [pg-backrest]: https://pgbackrest.org/
@@ -342,3 +359,5 @@ found in the archive. See the PostgreSQL documentation on
 [parallel importer]: https://github.com/timescale/timescaledb-parallel-copy
 [pg archiving]: https://www.postgresql.org/docs/current/continuous-archiving.html#BACKUP-PITR-RECOVERY
 [wale image]: https://hub.docker.com/r/timescale/timescaledb-wale
+[pg_create_restore_point]: https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADMIN-BACKUP-TABLE
+[create_distributed_restore_point]: /api#create_distributed_restore_point


k-rus:
@pmwkaa Will the short description how to do multinode backup be in a separate PR?

pmwkaa:
@k-rus It is not clear to me how we should document it other then already there, since this functionality is PostgreSQL compatible and follows `Performing Physical Backups` section. What user should do exactly, other then already described? This is much more related on the internal infrastructure of user, rather then specific set of commands to execute.

If you take a look at the backup section using WAL-E https://docs.timescale.com/latest/using-timescaledb/backup#docker-wale, everything is described in terms of docker containers rather then a way to configure PostgreSQL recovery, which makes sense.

I don't think we have the same for multi-node. Feels like this should be a part of a way to configure and maintain/run multi-node cluster using Kubernetes, etc.

I guess it is up for discussion, what people think we should include there? @erimatnor @mfreed

k-rus:
@pmwkaa The description can be simple, like: create the distributed restore point from the access node, and then backup the access node and each data node. During the restore of each node use the restore point. Since the distributed restore point was created on each node to preserve consistency between nodes, the restoring to it will result in the consistent state.

pmwkaa:
> @pmwkaa The description can be simple, like: create the distributed restore point from the access node, and then backup the access node and each data node. During the restore of each node use the restore point. Since the distributed restore point was created on each node to preserve consistency between nodes, the restoring to it will result in the consistent state.

Feels like the same is already said here and in the api description

page-index/page-index.js
```diff
@@ -658,6 +658,11 @@ const pageIndex = [
             type: ANCHOR,
             href: "#docker-wale",
           },
+          {
+            Title: "Named restore point",
```
mfreed, line 662: We probably want to consistently use a common term for this.  Restore points?  Consistent restore points?  Named restore points?  something else?  Below we use multiple ones.

(Is one widely used already from DB / postgres literature here?)

using-timescaledb/backup.md
```diff
@@ -331,6 +328,26 @@ end as the recovery will be complete when no further files can be
...
 [continuous archiving][pg archiving] for more information.
 
+### Named restore point [](named_restore_point)
+
+TimescaleDB natively support a consistent restore point both for single-node and for
+multi-node environments. The restore point can later be used as a recovery target on
+single-node or on each node of the entire multi-node cluster, ensuring that it can be
+restored to a consistent state.
+
+To create a named restore point on a single-node [`pg_create_restore_point`][pg_create_restore_point]
```
mfreed, line 338: What's the advantage of this in single-node?  

I think the above paragraph introduction should be a bit clearer about the user advantage of single vs. multi-node clusters.  In single this (I think) is more about ease-of-use --> just having a specific name to restore to.  In multi-node this is also (mainly?) about correctness/consistency.

page-index/page-index.js
NunoFilipeSantos: @pmwkaa Mike has a point we should be consistent.

NunoFilipeSantos (changes requested):

using-timescaledb/backup.md
pmwkaa: The advantage of using restore points in single node is the same as in multi-node, the idea here is that it expected to work in the same way. Maybe we should just concentrate on `distributed restore points` in such case and remove mentioning of the single-node?

NunoFilipeSantos (approved):

using-timescaledb/backup.md
```diff
@@ -331,6 +328,26 @@ end as the recovery will be complete when no further files can be
 found in the archive. See the PostgreSQL documentation on
 [continuous archiving][pg archiving] for more information.
 
+### Restore points [](restore_points)
+
+TimescaleDB natively support a restore point both for single-node and for
```
mfreed, line 333:
```suggestion
TimescaleDB natively supports a restore point both for single-node and for
```

using-timescaledb/backup.md
```diff
@@ -331,6 +328,26 @@ end as the recovery will be complete when no further files can be
...
+multi-node environments. The restore point can later be used as a recovery target on
+single-node or on each node of the entire multi-node cluster, ensuring that it can be
+restored to a consistent state.
+
+To create a restore point on a single-node [`pg_create_restore_point`][pg_create_restore_point]
+function can be used.
```
mfreed, line 339:
```suggestion
To create a restore point on a single-node, the [`pg_create_restore_point`][pg_create_restore_point]
function can be used.
```

using-timescaledb/backup.md
```diff
@@ -331,6 +328,26 @@ end as the recovery will be complete when no further files can be
...
+
+On a multi-node node [`create_distributed_restore_point`][create_distributed_restore_point] function must be used
```
mfreed, line 341:
```suggestion
On a multi-node cluster, the [`create_distributed_restore_point`][create_distributed_restore_point] function must be used
```

using-timescaledb/backup.md
```diff
@@ -331,6 +328,26 @@ end as the recovery will be complete when no further files can be
...
+instead and executed on the access-node. In order to achieve synchronization across the multi-node cluster, this function ensures
+that the restore point created on the access node is synchronized with each data node. Care should be taken to ensure
+third-party solutions properly quiesce the environment before backing up, so that the
+restore point used across nodes does not have outstanding transactions running on the data nodes.
```
mfreed, line 345: This should be clarified about what is needed here.  I thought the restore point would take a lock.  This makes it sound like something is needed for _correctness_, or are you just worried about lock delay / deadlock?  Can this be clarified?

using-timescaledb/backup.md
```diff
@@ -331,6 +328,26 @@ end as the recovery will be complete when no further files can be
...
+
+Before using restore point make sure PostgreSQL is configured for [WAL Archiving][wal archiving]. To recover to the desired
```
mfreed, line 347: I think this paragraph is confusing.  It points to a specific section of postgres doc, where instead I think it should also apply to higher-level tools like pgBackrest or WAL-G that use WAL-archiving, no?  I probably would clarify this.

Also, the language could be clearer:

```In order to use restore points, you must configure PostgreSQL to use [WAL archiving][wal archiving], such
that the database stores a copy of all WAL segment, where the restore point will point to a specific
segment (location) in the WAL in which to restore to.  To use a restore point, ...
```
maybe you can finish the paragraph.  It could refer to the same language you currently use, but I think there are probably higher-level pointers to tooks like pgbackrest or WAL-G?

using-timescaledb/backup.md
pmwkaa: The basic ideas here is that the function that creates distributed restore point can only guarantee consistency while using TimescaleDB remote transactions. If user have some custom logic and running transactions on a data node directly at the same time (basically without access node knowledge), there are no consistency guarantees for such case.

using-timescaledb/backup.md
pmwkaa: I guess part of this section was related to the data nodes synchronization, mentioned above. I've tried to reformulate the paragraph and reference the third-party tools instead of pointing in to the PG doc

using-timescaledb/backup.md
```diff
@@ -331,6 +328,25 @@ end as the recovery will be complete when no further files can be
...
+restored to a consistent state.
+
+To create a restore point on a single-node, the [`pg_create_restore_point`][pg_create_restore_point]
+function can be used.
+
+On a multi-node cluster, the [`create_distributed_restore_point`][create_distributed_restore_point] function must be used
+instead and executed on the access node. In order to achieve synchronization across the multi-node cluster, this function ensures
+that the restore point created on the access node is synchronized with each data node. Care should be taken to ensure
+that the data nodes has no locally running user transactions executed without the access node knowledge at the time
+of creating distributed restore point.
```
mfreed, line 345:
```suggestion
that the data nodes have no locally-running user transactions executed without the access node's involvement when
creating distributed restore point (i.e., by an external user connecting directly to a data node and executing a local
transaction on that data node).
```

mfreed (approved):

helrabelo:
Please add the labels for the versions these changes apply? @pmwkaa

PR #701, opened 2021-04-04
fix(quickstart-node): Fix broken link in "Generate hypertable" tip, add link for Sequelize, fix code

janpio:
Fixes a small error that make the tip look like this right now:

image

And adds an actual link for Sequelize so this misrender is also gone:

image

### DIFF
--- a/tutorials/quickstart-node.md
+++ b/tutorials/quickstart-node.md
@@ -433,7 +433,7 @@ app.get('/', async (req, res) => {
         });
 
         // send response 
-        res.send('Inserted!);
+        res.send('Inserted!');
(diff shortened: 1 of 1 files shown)


jonatas (approved):
Thanks for the fixes @janpio!

Loquacity:
This repo is deprecated, please use https://github.com/timescale/docs

janpio:
Nice to be reminded of an old PR this way...

But seems the changes suggested here are already applied to https://github.com/timescale/docs/blob/latest/timescaledb/quick-start/node.md besides the missing Sequelize link.

